PR #4, opened 2022-08-03
Socket work, UI list of participants, better state wrapping

filbabic:


stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -212,21 +191,55 @@ public class CallViewModel(
...
+
             init(room, call!!, url, token)
         }
     }
+
+    override fun onCleared() {
+        super.onCleared()
+        val room = _roomState.value ?: return
+
+        room.disconnect()
```
skydoves, line 227: It would be nice if we could prevent side effects from the `return` by using the null safe expression.
```kotlin
_roomState.value?.disconnect()
```

stream-video-android/build.gradle
```diff
@@ -84,7 +84,7 @@ dependencies {
     implementation Dependencies.okhttpLoggingInterceptor
     implementation Dependencies.kotlinSerializationJson
 
-    api Dependencies.liveKit
+    implementation Dependencies.liveKit
```
skydoves, line 89: This is breaking the `org.webrtc.SurfaceViewRenderer` in the `activity_main.xml` layout. 
We can resolve this by including `api 'com.github.webrtc-sdk:android:104.5112.01'` in our video module.

References: https://github.com/webrtc-sdk/android

It seems the LiveKit developer built WebRTC pre-compiled library.

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
filbabic: Good point! :]

PR #29, opened 2022-10-03
[14] Add remaining events to WS and events parser/mapper

kanat:


stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/IncomingCallActivity.kt
```diff
@@ -41,16 +42,18 @@ public class IncomingCallActivity : AppCompatActivity() {
...
+                    callType = CallType.VIDEO,
+                    participants = emptyList(),
+                    onDeclineCall = { finish() },
+                    onAcceptCall = { callId, isVideoEnabled ->
+                        joinCall(callId, isVideoEnabled)
+                    },
+                    onVideoToggleChanged = { }
+                )
+            }
         }
```
kanat, line 57: it was crashing down to the flow because of missing theme

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/participants/ParticipantInformation.kt
```diff
@@ -86,15 +86,19 @@ public fun ParticipantInformation(
...
     }
 }
 
 private fun buildLargeCallText(participants: List<CallParticipant>): String {
+    if (participants.isEmpty()) return "No participants"
     val initial = buildSmallCallText(participants)
+    if (participants.size == 1) return initial
 
     return "$initial and +${participants.size - 2} more"
 }
```
kanat, line 104: fixes for `IndexOutOfBoundsException`

stream-video-android/src/main/kotlin/io/getstream/video/android/events/VideoEvent.kt
```diff
@@ -43,7 +46,60 @@ public data class HealthCheckEvent(
...
+) : VideoEvent()
+
+/**
+ * Sent when call members get updated.
+ */
+public data class CallMembersDeletedEvent(
+    val callId: String,
+    val users: Map<String, CallUser>,
+    val info: CallInfo?,
+    val details: CallDetails?
```
kanat, line 102: new events can be found here

stream-video-android/src/main/kotlin/io/getstream/video/android/events/model/VideoModel.kt
```diff
@@ -0,0 +1,89 @@
...
+    val type: String,
+    val createdByUserId: String,
+    val createdAt: Date?,
+    val updatedAt: Date?
+)
+
+public data class CallDetails(
+    val memberUserIds: List<String>,
+    val members: Map<String, CallMember>
+)
```
kanat, line 53: Events content

stream-video-android/src/main/kotlin/io/getstream/video/android/socket/EventMapper.kt
```diff
@@ -36,7 +44,57 @@ internal object EventMapper {
...
+            )
+        }
+
+        socketEvent.call_members_deleted != null -> with(socketEvent.call_members_deleted) {
+            CallMembersDeletedEvent(
+                callId = call_cid,
+                users = socketEvent.users.toCallUsers(),
+                info = socketEvent.calls[call_cid]?.toCallInfo(),
+                details = socketEvent.call_details[call_cid]?.toCallDetails(),
+            )
```
kanat, line 97: WS events mapping

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/connection/StreamPeerConnection.kt
```diff
@@ -181,38 +208,81 @@ public class StreamPeerConnection(
...
+    }
+
+    private fun observeStats() = coroutineScope.launch {
+        while (isActive) {
+            delay(10_000L)
+            connection.getStats {
+                logger.v { "[observeStats] #sfu; #$typeTag; stats: $it" }
+            }
+        }
+    }
```
kanat, line 276: Added `stats` observation. Was wondering if it contains any insight about audio issue

buildSrc/src/main/kotlin/io/getstream/video/android/Dependencies.kt
```diff
@@ -25,6 +25,7 @@ object Versions {
     internal const val OKHTTP = "4.10.0"
 
     internal const val WEBRTC = "104.5112.01"
+    internal const val STREAM = "5.8.0"
```
filbabic, line 28: can probably bump now

stream-video-android/src/main/kotlin/io/getstream/video/android/events/VideoEvent.kt
```diff
@@ -43,7 +46,60 @@ public data class HealthCheckEvent(
  * Sent when someone creates a call and invites another person to participate.
  */
 public data class CallCreatedEvent(
-    val callId: String
+    val callId: String,
+    val users: Map<String, CallUser>,
+    val info: CallInfo?,
```
filbabic, line 51: None of these events should have nullable values. 

If there's something defined in proto,. unfortunately it's nullable in kotlin generated code, but the data should be there.

We should either expect an object to be 100% there, or in case of collections to be empty in the wrost case.

Otherwise these events are nothing different from proto definitions and users will have a hard time using them

We could also try to flatten the structure here not ot have so many objects nested, but rather have it as flat as possible

stream-video-android/src/main/kotlin/io/getstream/video/android/events/model/VideoModel.kt
filbabic: Maybe we don't need all of these in the events - we could flatten out the events and have these directly as properties

stream-video-android/src/main/kotlin/io/getstream/video/android/socket/EventMapper.kt
```diff
@@ -36,7 +44,57 @@ internal object EventMapper {
         }
 
         socketEvent.call_created != null -> with(socketEvent.call_created) {
-            CallCreatedEvent(call_cid)
+            CallCreatedEvent(
+                callId = call_cid,
+                users = socketEvent.users.toCallUsers(),
```
filbabic, line 49: we should check with BE but all of these should be non-null, it's just that wire generates them as nullable

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/connection/StreamPeerConnection.kt
filbabic: Good job!

We'll also need to push this to the BE once it's ready!

stream-video-android/src/main/kotlin/io/getstream/video/android/events/VideoEvent.kt
kanat: if we will flatten the structure, then we def won't avoid having lots of nullable props, since in proto is wrapped into colletions

PR #31, opened 2022-10-07
Add audio level handling, primary speaker, expose more state in Call

filbabic:


app/src/main/kotlin/io/getstream/video/android/app/ui/DeeplinkingActivity.kt
```diff
@@ -84,7 +86,7 @@ class DeeplinkingActivity : AppCompatActivity() {
         VideoApp.initializeStream(
             credentialsProvider = FakeCredentialsProvider(
                 userCredentials = selectedUser,
-                apiKey = "key1"
+                apiKey = "key10"
```
kanat, line 89: Hey @filbabic,

Is this gonna work with local running servers, right?

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/signal/RemoteSignalService.kt
```diff
@@ -33,24 +33,24 @@ import stream.video.sfu.UpdateSubscriptionsResponse
...
     public override suspend fun join(@Body joinRequest: JoinRequest): JoinResponse
 
     @Headers("Content-Type: application/protobuf")
-    @POST("/rpc/twirp/stream.video.sfu.SignalServer/SetPublisher")
+    @POST("/twirp/stream.video.sfu.SignalServer/SetPublisher")
     public override suspend fun setPublisher(@Body request: SetPublisherRequest): SetPublisherResponse
 
     @Headers("Content-Type: application/protobuf")
-    @POST("/rpc/twirp/stream.video.sfu.SignalServer/UpdateSubscriptions")
+    @POST("/twirp/stream.video.sfu.SignalServer/UpdateSubscriptions")
```
kanat: Did they change the path? 🤔

filbabic: Yeah so it seems for this staging environment they don't require the /rpc stuff 🤷 We'll see how things go and what will change in the future

app/src/main/kotlin/io/getstream/video/android/app/ui/DeeplinkingActivity.kt
filbabic: They should!

We have an app10 in hte default apps for local environments, so if things don't work we can check with BE

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/signal/RemoteSignalService.kt
filbabic: Removed for local yeah! Might change in the future

PR #32, opened 2022-10-14
Task/incoming call, state management

filbabic:


app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
```diff
@@ -16,63 +16,104 @@
...
-        val userPreferences: UserPreferences by lazy {
-            UserPreferencesImpl(
-                instance.getSharedPreferences(KEY_PREFERENCES, MODE_PRIVATE)
-            )
+        return StreamCallsBuilder(
+            context = this,
+            credentialsProvider = credentialsProvider,
+            loggingLevel = loggingLevel
+        ).build().also {
+            // TODO can we initialize streamCalls directly in VideoApp?
```
filbabic: We shouldn't initialize anything directly - the customer should be responsible for initializing the StreamCalls construct, with a given user, API key and similar. This is different from chat and clearly communicates that the user will have to be present in order to connect to the rest

app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
```diff
@@ -16,63 +16,104 @@
...
-                credentialsProvider = credentialsProvider,
-                loggingLevel = loggingLevel
-            ).build()
-
-            return streamCalls
+    private fun observeState() {
+        appScope.launch {
+            streamCalls.callState.collect { state ->
+                when (state) {
+                    is StreamCallState.Incoming -> {
```
filbabic: While I do like that we have an internal state wrapper that handles events and exposes them to the user, we should consider if a user is going to want to do things this way.

For example - we need to start the activity in NEW_TASK just because we're not starting from an Activity context. It migth just be easier to instead create a router like this:

```kotlin
public class StreamRouterImpl(private val context: Context): StreamRouter {

    override fun onIncomingCall(incomingCallState) {
        startActivity(IncomingCallActivity.getIntent(context, incomingCallState))
    }

.... // other navigation functions
}
```

And then build them with each ViewModel and pass them in to, so that we can navigate to the respective place.

I also don't like the idea of collecting events as state (as can be seen later in the PR), because Flows are supposed to represent states that change, not one-off events.

filbabic: This also breaks the SRP because now the Application class also takes care of navigation, which is a bit odd

app/src/main/kotlin/io/getstream/video/android/app/lifecycle/StreamActivityLifecycleCallbacks.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.getstream.video.android.app.lifecycle
+
+import android.app.Activity
+import android.os.Bundle
+
+internal class StreamActivityLifecycleCallbacks(
+    private val onActivityCreated: (Activity) -> Unit = {},
+    private val onActivityStarted: (Activity) -> Unit = {},
```
filbabic: Can we inline all of these?

app/src/main/kotlin/io/getstream/video/android/app/ui/DeeplinkingActivity.kt
```diff
@@ -81,12 +91,12 @@ class DeeplinkingActivity : AppCompatActivity() {
...
-        val selectedUser = VideoApp.userPreferences.getCachedCredentials()
-
-        VideoApp.initializeStream(
+        val selectedUser = videoApp.userPreferences.getCachedCredentials()
+        logger.d { "[logIn] selectedUser: $selectedUser" }
+        videoApp.initializeStreamCalls(
             credentialsProvider = FakeCredentialsProvider(
                 userCredentials = selectedUser,
-                apiKey = "key10"
+                apiKey = "key1"
```
filbabic: I think the current API keys are set to `key10` so we should probably migrate to that

app/src/main/kotlin/io/getstream/video/android/app/ui/call/CallActivity.kt
```diff
@@ -148,23 +157,7 @@ class CallActivity : AppCompatActivity() {
...
-            callId = callId,
-            sfuUrl = sfuUrl,
-            userToken = userToken,
-            iceServers = iceServers?.iceServers ?: emptyList(),
-            callSettings = CallSettings(
-                audioOn = false,
-                videoOn = false
-            )
-        )
+        callViewModel.createCall()
```
filbabic: Rename this to something like `startCall` or `connectToCall` because the call has already been created

app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -191,44 +207,68 @@ class HomeActivity : AppCompatActivity() {
...
+                joinCall(callId = callIdState.value)
             },
             content = {
                 Text(text = "Join call")
             }
         )
     }
 
-    private fun joinCall(callId: String, participants: List<String> = emptyList()) {
+    private fun dial(callId: String, participants: List<String> = emptyList()) {
```
filbabic: As mentioned - I don't think we should force the dial option in the sample - we should be able to create a call without any participants invited directly, but they can join later on or we can invite them.

app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -191,44 +207,68 @@ class HomeActivity : AppCompatActivity() {
...
+            }
+            result.onError {
+                logger.e { "[dial] failed: $it" }
+                Toast.makeText(this@HomeActivity, it.message, Toast.LENGTH_SHORT).show()
+            }
+        }
+    }
 
-            val createCallResult = controller.createAndJoinCall(
+    private fun joinCall(callId: String) {
```
filbabic, line 337: Joining call should be the same regardless of participants, the only thing we should see if is the user wants to "ring" people.

We should consult the design/other teams how they make this difference

app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -300,6 +340,31 @@ class HomeActivity : AppCompatActivity() {
...
 
+    @Composable
+    fun BoxScope.UserIcon() {
+        val user = videoApp.credentialsProvider.getUserCredentials()
+        Box(
+            modifier = Modifier
+                .padding(top = 8.dp, start = 8.dp)
+                .align(TopStart)
+        ) {
+            Box(
```
filbabic: We can flatten this Box to just be a modifier that goes into the avatar instead

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingCallActivity.kt
```diff
@@ -0,0 +1,130 @@
...
+class IncomingCallActivity : AppCompatActivity() {
+
+    private val logger by lazy { StreamLog.getLogger("Call:Incoming-View") }
+    private val viewModel by viewModels<IncomingViewModel> { IncomingViewModelFactory(videoApp.streamCalls) }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        showWhenLockedAndTurnScreenOn()
+        super.onCreate(savedInstanceState)
+
+        observeIncomingCall()
```
filbabic: So I don't like these events, because we're listening to observable constructs which can have multiple events, but we're only consuming one.

Instead, I'd go the "router" way and just have our users provide it to the ViewModels. Worst case - we won't expose the UI or flow as much and we might just have users implement the whole thing

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+            _dropEvent.emit(Unit)
+        }
+    }
+
+    private fun observeCallState() {
+        viewModelScope.launch {
+            streamCalls.callState.collect { state ->
+                logger.i { "[observeCallState] state: $state" }
+                when (state) {
+                    is State.Incoming -> {
```
filbabic: This doesn't make much sense to me - we already know we're in the Incoming State, we've started the IncomingCallActivity iwth the state and then in the ViewModel that's created, we listen to what the state is and if it's incoming (which it has to be initially), we show some UI

I think it's much better to just have these as imperative rather than declarative, as we're overcomplicating what should be simple: send data through an intent, show that data.

Rather than - store data in a state machine, react to it, start an activity with no state, observe the state again in the activity, show the UI

It's not going to be the best UX for our customers either

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+                        _dropEvent.emit(Unit)
+                    }
+                }
+            }
+        }
+    }
+
+    fun hangUp() {
+        val data = _incomingData.value ?: return
+        viewModelScope.launch {
```
filbabic: similar here - I think we can just use an imperative style approach with a router

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+                UserEventType.USER_EVENT_TYPE_REJECTED_CALL
+            )
+            streamCalls.leaveCall()
+
+            logger.d { "[hangUp] result: $result" }
+            _dropEvent.emit(Unit)
+        }
+    }
+
+    fun pickUp() {
```
filbabic: .

app/src/main/kotlin/io/getstream/video/android/app/ui/outgoing/OutgoingCallActivity.kt
```diff
@@ -0,0 +1,171 @@
...
+import io.getstream.video.android.model.CallType
+import io.getstream.video.android.model.IceServer
+import io.getstream.video.android.socket.SocketListener
+import io.getstream.video.android.utils.onError
+import io.getstream.video.android.utils.onSuccessSuspend
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import stream.video.coordinator.client_v1_rpc.UserEventType
+
+class OutgoingCallActivity : AppCompatActivity(), SocketListener {
```
filbabic: This example is much more imperative, but it makes sense, as there's no state to observe here other than WS events -> and we react to them by triggering navigational items.

app/src/main/kotlin/io/getstream/video/android/app/utils/UserUtils.kt
```diff
@@ -24,21 +24,21 @@ fun getUsers(): List<UserCredentials> {
             id = "filip",
             name = "Filip",
             image = "https://avatars.githubusercontent.com/u/17215808?v=4",
-            token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdHJlYW0tdmlkZW8tZ29AdjAuMS4wIiwic3ViIjoidXNlci9maWxpcCIsImlhdCI6MTY2NTE0MjQ0OCwidXNlcl9pZCI6ImZpbGlwIn0.E5qKFJAA8bE39Y3AmZECcpk_ayYQKLTMLJ90QPrI3p0",
+            token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdHJlYW0tdmlkZW8tZ29AdjAuMS4wIiwic3ViIjoidXNlci9maWxpcCIsImlhdCI6MTY2NTY0NzMzMCwidXNlcl9pZCI6ImZpbGlwIn0.0JjbNmaL2F0oHulx9w6QLWb4HoxhTADa8n3SjHwvi9A",
```
filbabic, line 27: Again - we could probably just use one set of keys, I think key10 is better since it's also covered in the hosted (staging) env

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCalls.kt
```diff
@@ -64,9 +70,21 @@ public interface StreamCalls {
...
+    /**
+     * Authenticates the user to join a given [CallMetadata].
+     *
+     * @param type The call type.
+     * @param id The call ID.
+     *
+     * @return [Result] which contains the [JoinedCall] with the auth information required to fully
+     * connect.
+     */
+    public suspend fun acceptCall(type: String, id: String): Result<JoinedCall>
```
filbabic: Again - this might not just be "acceptCall". We can also join a call that's already active - like a zoom/hangouts meeting

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -79,45 +91,108 @@ public class StreamCallsImpl(
...
     ): Result<JoinedCall> {
         logger.d { "[createAndJoinCall] type: $type, id: $id, participantIds: $participantIds" }
-        return callClient.createAndJoinCall(type, id, participantIds)
+        return callClient.createAndJoinCall(type, id, participantIds, ringing)
+    }
+
+    override suspend fun acceptCall(type: String, id: String): Result<JoinedCall> {
+        logger.d { "[acceptCall] type: $type, id: $id" }
+        // TODO engine.onCallAccepting()
+        val eventResult = sendEvent(id, type, UserEventType.USER_EVENT_TYPE_ACCEPTED_CALL)
```
filbabic: I don't think we should send this event here - we're not necessarily accepting someone's call, it might clog the BE/WS with faulty events

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -79,45 +91,108 @@ public class StreamCallsImpl(
...
+        logger.v { "[acceptCall] eventResult: $eventResult" }
+        return when (eventResult) {
+            is Failure -> eventResult.also {
```
filbabic: Especially because of this - if we're trying to accept a call that no one invited us to, what's going to be the result?

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -79,45 +91,108 @@ public class StreamCallsImpl(
...
+            is Success -> {
+                val getOrCreateResult = callClient.getOrCreateCall(type, id, emptyList(), false)
+                logger.v { "[acceptCall] getOrCreateResult: $getOrCreateResult" }
+                when (getOrCreateResult) {
+                    is Success -> {
+                        val metadata = getOrCreateResult.data.call?.toCall()
+                            ?: error("[acceptCall] no CallEnvelope found")
+                        callClient.joinCall(metadata).also { joinResult ->
+                            when (joinResult) {
+                                is Success -> engine.onCallJoined(joinResult.data)
```
filbabic: we don't need this because it's also being called in `joinCall`

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -79,45 +91,108 @@ public class StreamCallsImpl(
...
+    ): Result<Boolean> {
+        logger.d { "[sendEvent] event type: $userEventType" }
+        return callClient.sendUserEvent(
+            userEventType = userEventType,
+            callId = callId,
+            callType = callType,
+        )
+    }
+
+    override suspend fun sendEvent(userEventType: UserEventType): Result<Boolean> {
```
filbabic: We should remove this most likely

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -79,45 +91,108 @@ public class StreamCallsImpl(
...
+        logger.d { "[sendEvent] event type: $userEventType" }
+        return callClient.sendUserEvent(userEventType)
     }
 
     override fun leaveCall() {
         credentialsProvider.setSfuToken(null)
         socket.updateCallState(null)
+        engine.onCallLeaved()
```
filbabic: left* or onLeaveCall

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -140,11 +215,6 @@ public class StreamCallsImpl(
     override fun removeSocketListener(socketListener: SocketListener): Unit =
         socket.removeListener(socketListener)
 
-    /**
-     * Domain - WebRTC.
-     */
-    private var webRTCClient: WebRTCClient? = null
```
filbabic: So I purposefully left this in the middle, to signal that this is only important when we reach the functions that create it. We can keep it at the top but to me it made more sense to be at the bottom

stream-video-android/src/main/kotlin/io/getstream/video/android/audio/AudioDeviceManager.kt
```diff
@@ -61,15 +68,19 @@ internal class AudioDeviceManager(
     @SuppressLint("NewApi")
     fun setAudioFocus() {
         // Request audio focus before making any device switch.
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        if (/*Build.VERSION.SDK_INT >= Build.VERSION_CODES.O*/false) {
```
filbabic: Any reason why this is commented out?

stream-video-android/src/main/kotlin/io/getstream/video/android/client/CallClient.kt
```diff
@@ -196,21 +201,37 @@ public class CallClient(
...
+                call_type = callType,
+                event_type = userEventType
+            )
+        )
+    }
+
+    /**
+     * @see CallCoordinatorClient.sendUserEvent for details.
+     */
+    public suspend fun sendUserEvent(userEventType: UserEventType): Result<Boolean> {
```
filbabic: probs remove

stream-video-android/src/main/kotlin/io/getstream/video/android/model/Call.kt
```diff
@@ -157,7 +157,7 @@ public class Call(
             mediaStream.audioTracks.forEach { track ->
                 logger.v { "[addStream] #sfu; audioTrack: ${track.stringify()}" }
                 track.setEnabled(true)
-                track.setVolume(15.0)
+                track.setVolume(16.5)
```
filbabic: not sure what this is about - probably can remove

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.getstream.video.android.model.CallDetails
+import io.getstream.video.android.model.CallInfo
+import io.getstream.video.android.model.CallUser
+import io.getstream.video.android.model.JoinedCall
+
+public sealed interface StreamCallState {
+
+    public object Idle : StreamCallState
+
+    public data class Drop(
```
filbabic: I really think we should expose the states that make sense here not to convolute this too much:

- Drop state doesn't have to be here - if we drop the call, we don't care anymore, it's dropped and we can go to the Idle state

- Creating is a bit weird, but I guess we don't want to intercept any incoming calls if we're in the process of creating one, so it can stay

- Created also doesn't make sense - we created a call, but we're not in it, so we're in the Idle state. If we think about the Meeting room examples (Zoom, Google Meet...), you can create any number of meeting rooms, drop out, but people can still join them, and so can you.

The rest seem okay - if we're in the outgoing or incoming call state, we don't want to have more incoming calls. 

If we're in a call, the same, we don't want to start more calls or receive any

stream-video-android/src/main/kotlin/io/getstream/video/android/module/HttpModule.kt
```diff
@@ -78,7 +78,7 @@ internal class HttpModule(
             credentialsProvider.getCachedToken()
         }
         val updated = original.newBuilder()
-            .addHeader(HEADER_AUTHORIZATION, token)
+            .addHeader(HEADER_AUTHORIZATION, "Bearer $token")
```
filbabic, line 81: Do we really need this?

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -74,31 +78,32 @@ public class CallViewModel(
...
     public val isShowingSettings: StateFlow<Boolean> = _isShowingSettings
 
-    public fun init(
-        callId: String,
-        sfuUrl: String,
-        userToken: String,
-        iceServers: List<IceServer>,
-        callSettings: CallSettings
-    ) {
+    public fun createCall() {
```
filbabic: connectToCall or init makes more sense I'd say, because we're not creating a call here

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/WebRTCClientImpl.kt
```diff
@@ -485,6 +490,7 @@ internal class WebRTCClientImpl(
 
         val audioTrack = makeAudioTrack()
         audioTrack.setEnabled(true)
+        audioTrack.setVolume(17.5)
```
filbabic: any reason for this?

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/connection/StreamPeerConnection.kt
```diff
@@ -227,6 +229,8 @@ public class StreamPeerConnection(
             logger.v { "[onAddTrack] #sfu; #$typeTag; mediaStream: $mediaStream" }
             mediaStream.audioTracks?.forEach { remoteAudioTrack ->
                 logger.v { "[onAddTrack] #sfu; #$typeTag; remoteAudioTrack: ${remoteAudioTrack.stringify()}" }
+                remoteAudioTrack.setVolume(15.6)
```
filbabic: or this

app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
kanat: We can also have a case with Push Notifications where we might not have any UI running, so that's why I made an App layer as a launcher for calls

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingViewModel.kt
kanat: We may consider merging all call screens into one where we have to distinguish between states

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
kanat: `Drop` might be useful for cases when you wanna for instance play a different sounds for different drop reasons

stream-video-android/src/main/kotlin/io/getstream/video/android/module/HttpModule.kt
kanat: At least iOS & RN has this in their header

stream-video-android/src/main/kotlin/io/getstream/video/android/model/Call.kt
kanat: Oh, I've been experimenting with WebRTC to see places in internal Logs where this is being set.
Good catch

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/WebRTCClientImpl.kt
kanat: Same here, I've been experimenting with WebRTC to see places in internal Logs where this is being set.
Good catch

stream-video-android/src/main/kotlin/io/getstream/video/android/webrtc/connection/StreamPeerConnection.kt
kanat: Same here =]
My bad.

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
kanat: I just moved it up for the sake of readably, especially when you're unfamiliar with the codebase =]

app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
kanat: I'm mostly concerned about the way we gonna describe this in tutorials.

I mean if we tell customers that the SDK can be initialized and held anywhere in the app, maybe even in a single activity. 
However, it's not entirely true because the call might live much longer than any activity since users should be able to continue call flow even in BG or with activity being killed.

Taking into account what has been said above, we can state that `StreamCalls` in most cases should live within the `AppScope`.
Initialization is delayed, cause it requires a user and other data, but in the end, it lives in `AppScope` even in our sample app.

app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
kanat: We didn't finish the engine at all, it is just the beginning =]
The state will change as soon as corresponding events are coming

app/src/main/kotlin/io/getstream/video/android/app/lifecycle/StreamActivityLifecycleCallbacks.kt
kanat: sure

app/src/main/kotlin/io/getstream/video/android/app/ui/DeeplinkingActivity.kt
kanat: dunno why, but I've seen from the `video` project they still use `key1`.

I had to change it back to `key1` to keep everything working on my end.
It seems like I accidentally committed these changes

PR #34, opened 2022-10-14
[19] implement CallEngine as a single source of truth

kanat:


app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -111,13 +113,15 @@ class HomeActivity : AppCompatActivity() {
         override fun onEvent(event: VideoEvent) {
             if (event is CallCreatedEvent) {
                 // TODO - viewModel ?
-                router.onIncomingCall(
-                    IncomingCallData(
-                        callInfo = event.info,
-                        callType = CallType.fromType(event.info.type),
-                        participants = event.users.values.toList()
+                if (event.ringing) {
```
filbabic: I think we can chain these calls in the `if` statement ->  `if event is CallCreatedEvent && event.ringing`

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -93,55 +98,124 @@ internal class StreamCallEngineImpl(
...
                     info = info,
                     details = details
                 )
             }
         )
     }
+
+    private fun MutableStateFlow<State>.post(state: State) {
+        if (state == value) {
+            logger.w { "[post] rejected (duplicate state): $state" }
```
filbabic, line 331: We shouldn't have to do this ourselves, stateflow should use distinct/equals by default internally:

```
Strong equality-based conflation
Values in state flow are conflated using [Any.equals](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) comparison in a similar way to [distinctUntilChanged](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/distinct-until-changed.html) operator. It is used to conflate incoming updates to [value](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-state-flow/value.html) in [MutableStateFlow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-state-flow/index.html) and to suppress emission of the values to collectors when new value is equal to the previously emitted one. State flow behavior with classes that violate the contract for [Any.equals](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) is unspecified.
```

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -142,43 +144,31 @@ public class StreamCallsImpl(
         participantIds: List<String>,
         ringing: Boolean
     ): Result<JoinedCall> {
-        logger.d { "[createAndJoinCall] type: $type, id: $id, participantIds: $participantIds" }
-        return callClient.createAndJoinCall(type, id, participantIds, ringing)
+        logger.d { "[getOrCreateAndJoinCall] type: $type, id: $id, participantIds: $participantIds" }
+        engine.onCallStarting(type, id, participantIds, ringing, forcedNewCall = false)
+        return callClient.getOrCreateCall(type, id, participantIds, ringing)
+            .onSuccess { engine.onCallStarted(it.call) }
```
filbabic: why do we have two on success calls?

Can't we just sequentially call them one by one in one block?

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCallsImpl.kt
```diff
@@ -142,43 +144,31 @@ public class StreamCallsImpl(
...
+            .onSuccess { engine.onCallJoining(it.call) }
+            .flatMap { callClient.joinCall(it.call) }
+            .onSuccess { engine.onCallJoined(it) }
```
filbabic, line 157: and I guess another one here, but this makes sense as we're mapping it from one to another

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -98,60 +102,171 @@ internal class StreamCallEngineImpl(
...
 
-    private fun onCallAccepted() = scope.launchWithLock(mutex) {
-        logger.d { "[onCallAccepted] no args" }
+    private fun onCallAccepted(event: CallAcceptedEvent) = scope.launchWithLock(mutex) {
+        val state = _callState.value
+        if (state !is State.Outgoing) {
+            logger.w { "[onCallAccepted] rejected (state is not Outgoing): $state" }
+            return@launchWithLock
+        }
+        logger.d { "[onCallRejected] state: $state" }
```
filbabic: should be onCallAccepted

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -98,60 +102,171 @@ internal class StreamCallEngineImpl(
...
-        logger.d { "[onCallRejected] no args" }
+    private fun onCallRejected(event: CallRejectedEvent) = scope.launchWithLock(mutex) {
         val state = _callState.value
-        if (state is State.Idle) {
+        if (state !is State.Active) {
             return@launchWithLock
         }
-        _callState.post(State.Drop(reason = DropReason.Rejected))
+        logger.d { "[onCallRejected] no args" }
+        _callState.post(State.Drop(state.callGuid, DropReason.Rejected))
```
filbabic: maybe we should expose a function on the call state, called: `dropCall` that does two things:
- posts the state of the Drop
- Immediately posts the Idle state

That way we won't forget to post the idle state for any of our dropped calls

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -98,60 +102,171 @@ internal class StreamCallEngineImpl(
...
-        logger.d { "[onCallSendEvent] type: $type, id: $id, event: $UserEventType" }
+    override fun onCallEventSent(type: String, id: String, eventType: UserEventType) = scope.launchWithLock(mutex) {
+        logger.d { "[onCallEventSent] type: $type, id: $id, eventType: $UserEventType" }
     }
 
     override fun onCallJoining(call: CallMetadata) = scope.launchWithLock(mutex) {
-        logger.d { "[onCallJoining] call: $call" }
+        val state = _callState.value
+        if (state !is State.Starting && state !is State.Outgoing && state !is State.Incoming) {
+            logger.w { "[onCallStarting] rejected (state is not Starting/Outgoing/Accepting): $state" }
```
filbabic: wrong tag

stream-video-android/src/main/kotlin/io/getstream/video/android/model/CallCid.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.model
+
+public typealias StreamCallCid = String
+public typealias StreamCallType = String
+public typealias StreamCallId = String
```
filbabic, line 21: I'm honestly kind of against these, because we are hiding strings away behind a certain typealias, then we're improting that typealias as a different name

It's certainly going to cause confusion for us and customers if these are not aligned. Worst case I'd go with typealiases but only have one variant, rather than having both "StreamCallCid" and "CallCid" and so on

stream-video-android/src/main/kotlin/io/getstream/video/android/socket/EventsParser.kt
```diff
@@ -48,9 +48,10 @@ internal class EventsParser(
         super.onMessage(webSocket, bytes)
         try {
             val rawEvent = WebsocketEvent.ADAPTER.decode(bytes)
-            logger.v { "[onMessage] rawEvent: $rawEvent" }
+
+            if (rawEvent.healthcheck == null) logger.v { "[onMessage] rawEvent: $rawEvent" }
             val processedEvent = EventMapper.mapEvent(rawEvent)
-            logger.v { "[onMessage] processedEvent: $processedEvent" }
+            if (rawEvent.healthcheck == null) logger.v { "[onMessage] processedEvent: $processedEvent" }
```
filbabic: this should be processedEvent, right?

PR #35, opened 2022-10-18
Task/dogfooding

filbabic:


dogfooding/src/main/java/io/getstream/chat/android/dogfooding/DogfoodingApp.kt
```diff
@@ -35,16 +48,25 @@ class DogfoodingApp : Application() {
...
             loggingLevel = loggingLevel
         ).build().also {
-            streamCalls = it
+            calls = it
         }
     }
+
+    fun logOut() {
+        FirebaseAuth.getInstance().signOut()
+        credentialsProvider
```
skydoves: Seems to added by mistake?

dogfooding/src/main/java/io/getstream/chat/android/dogfooding/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,43 @@
...
+        val editor = sharedPreferences.edit()
+
+        editor.putString(KEY_ID, userCredentials.id)
+        editor.putString(KEY_TOKEN, userCredentials.token)
+        editor.putString(KEY_IMAGE, userCredentials.image)
+        editor.putString(KEY_NAME, userCredentials.name)
+        editor.putString(KEY_ROLE, userCredentials.role)
+
+        editor.apply()
+    }
```
skydoves: We can refactor like this with an `edit` extension.

```kotlin
sharedPreferences.edit {
            putString(KEY_ID, userCredentials.id)
            putString(KEY_TOKEN, userCredentials.token)
            putString(KEY_IMAGE, userCredentials.image)
            putString(KEY_NAME, userCredentials.name)
            putString(KEY_ROLE, userCredentials.role)   
}
```

filbabic: yeah that's true - good call! :D

dogfooding/src/main/java/io/getstream/chat/android/dogfooding/DogfoodingApp.kt
filbabic: Yeah it should be clearing instead! thanks!

PR #39, opened 2022-10-20
[19] support call accept / reject / cancel + call drop

kanat:


app/src/main/kotlin/io/getstream/video/android/app/ui/call/content/VideoCallContent.kt
```diff
@@ -97,8 +99,11 @@ internal fun VideoCallContent(
             val roomState = room
 
             Column(modifier = Modifier.fillMaxSize()) {
-
-                CallActionBar(callViewModel, "TODO - Call ID") // TODO -
+                val callId = when (val state = callState) {
+                    is StreamCallState.Active -> state.callGuid.cid
```
filbabic: Should we maybe use just the ID here, rather than CID? because we'll show something like: <call-type>:<call-id> so it might be a bit ugly

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingCallViewModel.kt
```diff
@@ -60,37 +60,23 @@ class IncomingCallViewModel(
...
-                declineCall()
-            }
+            streamCalls.joinCall(callType, callId)
+                .flatMap { joined ->
+                    logger.v { "[acceptCall] joined: $joined" }
+                    streamCalls.sendEvent(
+                        callCid = joined.call.cid,
+                        eventType = CallEventType.ACCEPTED
+                    ).also {
+                        streamRouter.navigateToCall(callInput = joined.toCallInput(), finishCurrent = true)
```
filbabic: Should we maybe move this out as a sequential thing, instead of using also, because it's not doing anything with the sendEvent response/result

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCalls.kt
```diff
@@ -121,12 +121,14 @@ public interface StreamCalls {
...
         callCid: String,
-        eventType: UserEventType
+        eventType: CallEventType
     ): Result<Boolean>
 
     /**
      * Leaves the currently active call and clears up all connections to it.
      */
+    // TODO
+    //  can be called internally when [StreamCallState.Idle] comes into the place.
```
filbabic, line 131: I think we can do it, but at the same time, it might be valuable to not automatically clean up based on the call state, as users might want to respond more to it

kanat: what if we give an option for customers to do it manually or let SDK handle everything automatically 🤔

app/src/main/kotlin/io/getstream/video/android/app/ui/call/content/VideoCallContent.kt
kanat: yeah, whatever works best =]

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamCalls.kt
filbabic: yeah that sounds good!

PR #49, opened 2022-10-27
[39] Implement a call service to keep calls running with no UI

kanat:
mp3 files were taken from https://pixabay.com/sound-effects/


stream-video-android/src/main/kotlin/io/getstream/video/android/service/StreamCallService.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    private val streamCalls: StreamCalls by lazy { getStreamCalls(this) }
+
+    private val notificationBuilder: StreamNotificationBuilder by lazy {
+        createNotificationBuilder(application)
+    }
+
+    private val vibroManager: StreamVibroManager by lazy { createVibroManager(application) }
+
+    private var curState: State = State.Idle
```
TGazica, line 57: `curState` can be renamed to `currentState`. Don't see the need to shorten the name.

stream-video-android/src/main/kotlin/io/getstream/video/android/service/notification/StreamNotificationBuilder.kt
```diff
@@ -0,0 +1,33 @@
...
+
+package io.getstream.video.android.service.notification
+
+import android.app.Notification
+import io.getstream.video.android.model.state.StreamCallState
+
+/**
+ * Handler responsible for showing and dismissing notification.
+ */
+public interface StreamNotificationBuilder {
```
TGazica, line 25: It says that it is a handler, but name is `StreamNotificationBuilder` and its logic builds a `IdentifiedNotification` instance. Are the kdocs correct?

stream-video-android/src/main/kotlin/io/getstream/video/android/service/vibro/StreamVibroManager.kt
```diff
@@ -0,0 +1,81 @@
...
+import android.os.VibrationEffect
+import android.os.Vibrator
+import androidx.annotation.RequiresApi
+import io.getstream.logging.StreamLog
+import io.getstream.video.android.utils.vibrator
+import io.getstream.video.android.utils.vibratorManager
+
+public interface StreamVibroManager {
+
+    public fun vibrateIncoming()
```
TGazica, line 30: maybe `vibrationStart` insted of `vibrationIncoming`?

PR #51, opened 2022-10-28
[50] Fix race condition during join flow

kanat:


stream-video-android/src/main/kotlin/io/getstream/video/android/service/StreamCallService.kt
```diff
@@ -108,8 +108,9 @@ public abstract class StreamCallService : Service() {
             is State.Active -> {
                 updateNotification(state)
                 when (state) {
-                    is State.Incoming -> {
-                        vibroManager.vibrateIncoming()
+                    is State.Incoming -> when (!state.acceptedByMe) {
+                        true -> vibroManager.vibrateIncoming()
+                        else -> vibroManager.cancel()
```
TGazica, line 113: i dont think when is needed here, simple if would suffice.

kanat: We should stop vibrating when the callee accepts the call.

TGazica: yes, but `when (!state.acceptedByMe)` is a simple if/else

PR #65, opened 2022-11-03
Task/api improvements

filbabic:
Various improvements around the API, added mute events


stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideo.kt
```diff
@@ -153,38 +152,26 @@ public interface StreamCalls {
...
-
-    public fun flipCamera()
-
-    public fun getAudioDevices(): List<AudioDevice>
+    ): CallClient
 
-    public fun selectAudioDevice(device: AudioDevice)
+    public fun getActiveCallClient(): CallClient?
 
     public suspend fun acceptCall(type: String, id: String): Result<JoinedCall>
```
filbabic, line 176: @kanat I'm curious if we need these functions exposed? I think we discussed these before - but is it maybe just easier to have the sendEvent function and use that, because this way we have four functions for effectively the same thing and it might be confusing to customers when to use sendEvent(ACCEPT_CALL) vs acceptCall()

PR #67, opened 2022-11-07
[38] refactor Outgoing & Incoming flows to represent call-engine state

kanat:
Relates to: https://github.com/GetStream/stream-video-android/issues/38

I've created this PR to avoid a chonker one in the end.
A final merge of 3 screens will be completed in the next PR.

Main changes:
* Outgoing & Incoming flows refactored to represent the call-engine state
* Modified `CallEngine` to handle the following cases:
  * do not drop a `meeting` if someone leaves it (but regular calls are dropped normally)
  * do not drop a `regular call` if not all participants reject incoming call
- Added fallback for latency measures (aligned with iOS)

### DIFF
--- a/stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
+++ b/stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
@@ -246,8 +246,8 @@ internal class StreamCallEngineImpl(
                         id = id,
                         cid = CallCid(type, id),
                     ),
-                    memberUserIds = participantIds,
-                    callKind = if (ringing) StreamCallKind.REGULAR else StreamCallKind.MEETING
+                    callKind = if (ringing) StreamCallKind.RINGING else StreamCallKind.MEETING,
+                    memberUserIds = participantIds
                 )
             )
         } else if (state is State.Incoming) {
@@ -274,11 +274,8 @@ internal class StreamCallEngineImpl(
         logger.d { "[onCallStarted] call: $call, state: $state" }
         _callState.post(
             State.Outgoing(
-                callGuid = StreamCallGuid(
-                    type = call.type,
-                    id = call.id,
-                    cid = call.cid,
-                ),
+                callGuid = state.callGuid,
+                callKind = state.callKind,
                 createdByUserId = call.createdByUserId,
                 broadcastingEnabled = call.broadcastingEnabled,
                 recordingEnabled = call.recordingEnabled,
@@ -286,11 +283,10 @@ internal class StreamCallEngineImpl(
                 updatedAt = StreamDate.from(call.updatedAt),
                 users = call.users,
                 members = call.members,
-                callKind = state.callKind,
                 acceptedByCallee = false
             )
         )
-        if (state.callKind == StreamCallKind.REGULAR) {
+        if (state.callKind == StreamCallKind.RINGING) {
             waitForCallToBeAccepted()
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingCallActivity.kt
```diff
@@ -27,16 +27,17 @@ import androidx.appcompat.app.AppCompatActivity
...
+import io.getstream.video.android.viewmodel.IncomingCallViewModelFactory
 
 class IncomingCallActivity : AppCompatActivity() {
 
     private val viewModel by viewModels<IncomingCallViewModel> {
         IncomingCallViewModelFactory(
             videoApp.streamVideo,
-            StreamRouterImpl(this),
-            requireNotNull(intent.getSerializableExtra(KEY_CALL_DATA) as? IncomingCallData)
+            // TODO passing [this] may lead to memory leak, cause VM may live longer than Activity
```
filbabic, line 39: Could you elaborate on this a bit? Any suggestions on how to avoid it?

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -218,7 +247,7 @@ internal class StreamCallEngineImpl(
                         cid = CallCid(type, id),
                     ),
                     memberUserIds = participantIds,
-                    ringing = ringing
+                    callKind = if (ringing) StreamCallKind.REGULAR else StreamCallKind.MEETING
```
filbabic: This might be defined as Ringing based on https://getstream.slack.com/archives/C022N8JNQGZ/p1667831316951249

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -129,3 +150,82 @@ public data class StreamCallGuid(
     val id: CallId,
     val cid: CallCid
 ) : Serializable
+
+public enum class StreamCallKind : Serializable {
```
filbabic, line 154: If we don't send this data into an intent, then we might not need serializable anywhere.

But it could be good for our users to be able to send it. WDYT?

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -103,6 +103,8 @@ public class CallViewModel(
         logger.d { "[createCall] input: $input" }
         // this._callState.value = videoClient.getCall(callId) TODO - load details
 
+        // TODO CallClient is supposed to live longer than VM
```
filbabic, line 106: Good point - maybe we can create the client before the ViewModel and hten pass it in as a parameter

app/src/main/kotlin/io/getstream/video/android/app/ui/incoming/IncomingCallActivity.kt
kanat: So `IncomingCallViewModel` keeps ref to `StreamRouterImpl`, which keeps ref to `IncomingCallActivity`, and in case of Activity recreation (like screen rotation etc), `IncomingCallViewModel ` will keep the ref to the dead obj, through `StreamRouterImpl `, because VM instance will still gonna be the same.

We could resolve this by using `ActivityLifecycleCallbacks` inside `StreamRouterImpl` to let it use any running activity

filbabic: This sounds good yeah - given that these are built for hte purposes of our apps, maybe it's not as important to expose all of this on the get go

Although, based on what we expose to users (e.g. Outgoing/Incoming screens), it might be good to just ship it too and make it clean!

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
kanat: Yeah, I've marked it as `Serializable`, cause states are already `Serializable`

PR #69, opened 2022-11-08
Feat/41 proto retrofit service generator

### LINKED ISSUE #41: Write a script or a Gradle task that generates Retrofit HTTP service files from proto RPC definitions
We have several examples of [RPC definitions](https://github.com/GetStream/stream-video-android/blob/b131b54bfbfdb6875eb5df0df9341fdb10fdb3ed/stream-video-android/src/main/proto/video/coordinator/client_v1_rpc/client_rpc.proto#L13) in our proto files which represent HTTP services. Other SDKs are using Twirp for generating such files and code, but Android doesn't have any viable [generators](https://github.com/twitchtv/twirp).

We should investigate and research if we can do the following:
- Reuse the code and setup of any of the Java generators.
- Re-write the code to generate Kotlin files that represent Retrofit services
- Either write a script that generates code before the build process, that we add to the project manually, or a Gradle task that generates them before the Retrofit code generation consumes them to generate HTTP communication code.

TGazica:
Closes #41 

Added `GenerateRPCServiceExtension`, `GenerateRPCServicePlugin`, `GenerateRPCServiceTask` that generate services to be used with Retrofit.

Task can be ran standalone: `./gradlew generateServices`. The task will also automatticaly run on builds.


buildSrc/src/main/kotlin/io/getstream/video/android/commands/rpc/task/GenerateRPCServiceTask.kt
```diff
@@ -0,0 +1,196 @@
...
+        for (protoFile in protoFiles) {
+            val proto = getCleanProtoFile(protoFile.readLines())
+            if (!proto.contains("service")) continue
+
+            val packageName = getPackageName(proto, protoFile.name)
+            val imports = getImports(proto)
+            val services = getRPCServices(proto, protoFile.name)
+
+            services.forEach { service ->
+                val isRPC = service.name.contains("RPC")
```
TGazica, line 79: This is an assumption on our side that if the Service has RPC in name the url for the call will have `/rpc/` prefix, otherwise `/twirp/` will be added.

buildSrc/src/main/kotlin/io/getstream/video/android/commands/rpc/plugin/GenerateRPCServicePlugin.kt
```diff
@@ -0,0 +1,31 @@
...
+            project.extensions.create(CONFIG_CLOJURE_NAME, GenerateRPCServiceExtension::class.java)
+
+        project.tasks.create(COMMAND_NAME, GenerateRPCServiceTask::class.java) {
+            // We take the projects build dir and set the output folder inside it.
+            extension.outputDir = "${project.buildDir}/generated/source/services/io/getstream/video/android/api"
+            this.config = extension
+        }
+
+        project.afterEvaluate {
+            this.tasks.getByName("preBuild").finalizedBy(project.tasks.getByName(COMMAND_NAME))
```
TGazica, line 28: When added to run after wire models generation, the `compileKotlin` fails. This will just run the services generation earlier in the build process.

buildSrc/src/main/kotlin/io/getstream/video/android/commands/rpc/task/GenerateRPCServiceTask.kt
filbabic: This should change and be aligned in the future, so we'll keep an eye out (as per your TODO)

PR #70, opened 2022-11-08
Add guest login to dogfooding, minor UI cleanup

filbabic:


app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -445,18 +444,31 @@ class HomeActivity : AppCompatActivity() {
     }
 
     @Composable
-    fun BoxScope.UserIcon() {
+    fun UserIcon() {
```
TGazica, line 447: Should we maybe move the `UserIcon` composable to `stream-video-android-compose` and expose it?

filbabic: We already have a UserAvatar and initials etc, but it requires a user avatar so I didn't want complicate things as we don't have the API for auth that fills in all the data

once we clear those things up should be easier as we'll have full users ready

TGazica: makes sense.

PR #72, opened 2022-11-11
Merge Incoming, Outgoing, Call activities into a single one

kanat:
Closes: https://github.com/GetStream/stream-video-android/issues/38

### DIFF
--- a/stream-video-android-compose/api/stream-video-android-compose.api
+++ b/stream-video-android-compose/api/stream-video-android-compose.api
@@ -236,9 +236,11 @@ public final class io/getstream/video/android/compose/theme/VideoThemeKt {
 	public static final fun VideoTheme (ZLio/getstream/video/android/compose/theme/StreamColors;Lio/getstream/video/android/compose/theme/StreamDimens;Lio/getstream/video/android/compose/theme/StreamTypography;Lio/getstream/video/android/compose/theme/StreamShapes;Lio/getstream/video/android/compose/theme/StreamIcons;Landroidx/compose/material/ripple/RippleTheme;Lio/getstream/video/android/compose/imageloading/StreamCoilImageLoaderFactory;Lkotlin/jvm/functions/Function2;Landroidx/compose/runtime/Composer;II)V
 }
 
-public abstract class io/getstream/video/android/compose/ui/AbstractComposeCallActivity : androidx/appcompat/app/AppCompatActivity, io/getstream/video/android/activity/StreamCallActivity {
+public abstract class io/getstream/video/android/compose/ui/AbstractComposeCallActivity : androidx/appcompat/app/AppCompatActivity, io/getstream/video/android/StreamVideoProvider {
 	public static final field $stable I
 	public fun <init> ()V
+	protected final fun buildContent ()Lkotlin/jvm/functions/Function2;
+	protected final fun getDefaultCallSettings ()Lio/getstream/video/android/model/CallSettings;
 	protected fun onCreate (Landroid/os/Bundle;)V
 	protected fun onResume ()V
 }
@@ -353,8 +355,8 @@ public final class io/getstream/video/android/compose/ui/components/outgoingcall
 }
 
 public final class io/getstream/video/android/compose/ui/components/outgoingcall/OutgoingCallContentKt {
-	public static final fun OutgoingCall (Lio/getstream/video/android/model/CallType;Ljava/lang/String;Ljava/util/List;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Landroidx/compose/runtime/Composer;II)V
-	public static final fun OutgoingCallScreen (Lio/getstream/video/android/viewmodel/CallViewModel;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Landroidx/compose/runtime/Composer;II)V
+	public static final fun OutgoingCall (Lio/getstream/video/android/model/CallType;Ljava/lang/String;Ljava/util/List;Lkotlin/jvm/functions/Function0;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Landroidx/compose/runtime/Composer;II)V
+	public static final fun OutgoingCallScreen (Lio/getstream/video/android/viewmodel/CallViewModel;Lkotlin/jvm/functions/Function0;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Landroidx/compose/runtime/Composer;II)V
 }
 
 public final class io/getstream/video/android/compose/ui/components/participants/CallParticipantKt {
--- a/stream-video-android/api/stream-video-android.api
+++ b/stream-video-android/api/stream-video-android.api
@@ -654,7 +654,6 @@ public abstract interface class io/getstream/video/android/StreamVideo {
 	public abstract fun getUser ()Lio/getstream/video/android/model/User;
 	public abstract fun joinCall (Lio/getstream/video/android/model/CallMetadata;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public abstract fun joinCall (Ljava/lang/String;Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public abstract fun launch (Lkotlin/jvm/functions/Function2;)V
 	public abstract fun rejectCall (Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public abstract fun removeSocketListener (Lio/getstream/video/android/socket/SocketListener;)V
 	public abstract fun sendCustomEvent (Ljava/lang/String;Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
@@ -696,14 +695,13 @@ public final class io/getstream/video/android/StreamVideoImpl : io/getstream/vid
 	public fun getUser ()Lio/getstream/video/android/model/User;
 	public fun joinCall (Lio/getstream/video/android/model/CallMetadata;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public fun joinCall (Ljava/lang/String;Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public fun launch (Lkotlin/jvm/functions/Function2;)V
 	public fun rejectCall (Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public fun removeSocketListener (Lio/getstream/video/android/socket/SocketListener;)V
 	public fun sendCustomEvent (Ljava/lang/String;Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public fun sendEvent (Ljava/lang/String;Lio/getstream/video/android/model/CallEventType;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
-public abstract interface class io/getstream/video/android/activity/StreamCallActivity {
+public abstract interface class io/getstream/video/android/StreamVideoProvider {
 	public abstract fun getStreamVideo (Landroid/content/Context;)Lio/getstream/video/android/StreamVideo;
 }
 
@@ -1593,7 +1591,7 @@ public final class io/getstream/video/android/logging/LoggingLevel : java/lang/E
 }
 
 public final class io/getstream/video/android/model/Call {
-	public fun <init> (Landroid/content/Context;Lio/getstream/video/android/token/CredentialsProvider;Lorg/webrtc/EglBase;)V
+	public fun <init> (Landroid/content/Context;Lkotlin/jvm/functions/Function0;Lorg/webrtc/EglBase;)V
 	public final fun getActiveSpeakers ()Lkotlinx/coroutines/flow/Flow;
 	public final fun getCallParticipants ()Lkotlinx/coroutines/flow/StateFlow;
 	public final fun getLocalParticipant ()Lio/getstream/video/android/model/CallParticipantState;
@@ -1667,15 +1665,6 @@ public final class io/getstream/video/android/model/CallInfo : java/io/Serializa
 	public final fun getRecordingEnabled ()Z
 	public final fun getType ()Ljava/lang/String;
 	public final fun getUpdatedAt ()Ljava/util/Date;
-	public fun hashCode ()I
-	public fun toString ()Ljava/lang/String;
-}
-
-public final class io/getstream/video/android/model/CallInput : java/io/Serializable {
-	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;ZZ)V
-	public final fun copy (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;ZZ)Lio/getstream/video/android/model/CallInput;
-	public static synthetic fun copy$default (Lio/getstream/video/android/model/CallInput;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;ZZILjava/lang/Object;)Lio/getstream/video/android/model/CallInput;
-	public fun equals (Ljava/lang/Object;)Z
 	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
 }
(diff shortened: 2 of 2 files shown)


app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -96,38 +89,33 @@ class HomeActivity : AppCompatActivity() {
...
     }
 
+    override fun finish() {
+        logger.d { "[finish] no args" }
+        super.finish()
+    }
+
     override fun onDestroy() {
-        streamVideo.removeSocketListener(socketListener)
+        logger.d { "[onDestroy] no args" }
```
filbabic: Do we need this then if we're not doing anything in onDestroy?

app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
```diff
@@ -96,38 +89,33 @@ class HomeActivity : AppCompatActivity() {
...
-        streamVideo.addSocketListener(socketListener)
         setContent {
             VideoTheme {
                 HomeScreen()
             }
         }
     }
 
+    override fun finish() {
+        logger.d { "[finish] no args" }
```
filbabic: Same here

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+import io.getstream.video.android.activity.StreamCallActivity
+import io.getstream.video.android.compose.theme.VideoTheme
+import io.getstream.video.android.compose.ui.components.call.CallScreen
+import io.getstream.video.android.model.CallSettings
+import io.getstream.video.android.model.state.StreamCallState
+import io.getstream.video.android.viewmodel.CallViewModel
+import io.getstream.video.android.viewmodel.CallViewModelFactory
+import io.getstream.video.android.viewmodel.PermissionManagerImpl
+
+public abstract class AbstractComposeCallActivity : AppCompatActivity(), StreamCallActivity {
```
filbabic: Will this be the recommended way of building our Call screen for customers?

We'll have to outline this in our tutorial if so

filbabic: Also - it mgith be better to rename `StreamCallActivity` to something else, as this seems like it implements two different set of activities (not possible in Android).

Maybe: `StreamVideoProvider`

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+            isGranted -> startVideoFlow()
+            deniedCamera || deniedMicrophone -> showPermissionsDialog()
+            else -> {
+                checkPermissions()
+            }
+        }
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        showWhenLockedAndTurnScreenOn()
```
filbabic, line 75: The reason why I don't like inheritence as much as composition is that in this case, users can override onCreate, not call super, or call setContent again and we have potential overhead, memory leaks or broken UI.

I do understand the drive to make it as simple as possible for our customers, but this feels like it'll cause bugs.

For example, if they override it and want to customize the UI, they either have to not call super (which is not intuitive and usually an antipattern) or call super, set the UI, then set new UI.

filbabic: But we can ship this as a standalone thing and if our customers want to build something themselves, we write a guide on how to connect things properly (as in our base example)

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+            }
+        )
+    }
+
+    private fun startVideoFlow() {
+        val isInitialized = callViewModel.isVideoInitialized.value
+        if (isInitialized) return
+        callViewModel.connectToCall(
+            CallSettings(
+                audioOn = false,
```
filbabic: Maybe this should be a function that returns the call settings and users can override it to provide a different default:

```kotlin
protected open fun getDefaultCallSettings(): CallSettings {
    return CallSettings(false, true, false)
}
```

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/CallContent.kt
```diff
@@ -312,3 +316,14 @@ private fun CallActionBar(callViewModel: CallViewModel, callId: String) {
         )
     }
 }
+
+private fun StreamCallState.asTitle() = when (this) {
+    is StreamCallState.Drop -> "Drop"
```
filbabic, line 323: Is it possible to use the class name here maybe? Or just expose a function in `StreamCallState` called:  `formatAsTitle` that each sealed class member implements

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/CallScreen.kt
```diff
@@ -0,0 +1,57 @@
...
+import io.getstream.video.android.model.state.StreamCallState as State
+
+@Composable
+public fun CallScreen(
+    viewModel: CallViewModel,
+    onRejectCall: () -> Unit,
+    onAcceptCall: () -> Unit,
+    onCancelCall: () -> Unit = {},
+    onMicToggleChanged: (Boolean) -> Unit = {},
+    onVideoToggleChanged: (Boolean) -> Unit,
```
filbabic, line 33: I would say that either all have default or none. Having a mix is confusing

Prefer all having default empty here

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideo.kt
```diff
@@ -37,6 +37,8 @@ public interface StreamVideo {
      */
     public val callState: StateFlow<StreamCallState>
 
+    public fun launch(block: suspend CoroutineScope.() -> Unit)
```
filbabic: Why do we need this here?

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
```diff
@@ -126,19 +111,40 @@ public class StreamVideoImpl(
...
+            }
+        }
+    }
+
     /**
      * Represents the state of the currently active call.
      */
     override val callState: StateFlow<StreamCallState> = engine.callState
 
+    override fun launch(block: suspend CoroutineScope.() -> Unit) {
```
filbabic: Yeah this feels a bit weird to me - if we're going for Java interop or something, not sure if they can pass in a suspending lambda

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
```diff
@@ -449,15 +454,17 @@ public class StreamVideoImpl(
...
-            .flatMap { joined ->
-                sendEvent(
-                    callCid = joined.call.cid,
-                    eventType = CallEventType.ACCEPTED
-                ).map { joined }
-            }.also { logger.v { "[acceptCall] result: $it" } }
+    override suspend fun acceptCall(cid: String): Result<JoinedCall> {
+        logger.d { "[acceptCall] cid: $cid" }
+        val (type, id) = cid.split(":").apply {
+            if (size != 2) return Failure(VideoError(message = "invalid cid: $cid"))
```
filbabic: Maybe we can pull these two functions out as utilities:

```kotlin
public fun isValidCid(cid: String): Boolean { // can also be an extension
    return cid.split(":").size == 2
}

public fun toIdAndType(cid: String): Pair<String, String> {
    return (maybe valid check here) cid.split(":").take(2)
}
```

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -114,6 +140,97 @@ internal class StreamCallEngineImpl(
         }
     }
 
+    override fun onSfuEvent(event: SfuDataEvent) {
+        if (event !is HealthCheckResponseEvent) {
+            logger.i { "[onSfuEvent] event: $event" }
+        }
+        when (event) {
+            is AudioLevelChangedEvent -> { }
```
filbabic, line 144: Does it make sense to only handle the ones we care about and then just use:
```kotlin
else -> {}
```

As a default? I know that if we add more events it might be missed in the else branch, but this way it's just a bit ugly because we're effectively doing that but in a much longer way

stream-video-android/src/main/kotlin/io/getstream/video/android/input/CallAndroidInputLauncher.kt
```diff
@@ -0,0 +1,67 @@
...
+            is CallActivityInput -> start(input)
+            is CallServiceInput -> start(input)
+        }
+    }
+
+    fun Context.start(input: CallActivityInput) {
+        StreamLog.d(TAG) { "/start/ activity: ${input.className}" }
+        startActivity(
+            newIntent(input.className).apply {
+                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TOP)
```
filbabic, line 51: Do we always need these flags? I imagine so because we're coming here from services

stream-video-android/src/main/kotlin/io/getstream/video/android/router/StreamRouter.kt
```diff
@@ -16,22 +16,11 @@
...
-    public fun navigateToCall(
-        joinedCall: JoinedCall,
-        finishCurrent: Boolean
-    )
-
-    public fun onIncomingCall()
+    public fun onUserLoggedIn()
 
-    public fun onOutgoingCall()
+    public fun onUserLoggedOut()
```
filbabic: Why does the router only know about login and logout now?

If that's the only thing it does, should be just remove the router altogether?

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -90,66 +95,109 @@ public class CallViewModel(
...
-    public val streamCallState: StateFlow<StreamCallState> get() = streamVideo.callState
+    public val streamCallState: StateFlow<State> get() = streamVideo.callState
+
+    private val _callType: MutableStateFlow<CallType> = MutableStateFlow(CallType.VIDEO)
+    public val callType: StateFlow<CallType> = _callType
+
+    private val _callId: MutableStateFlow<String> = MutableStateFlow(value = "")
+    public val callId: StateFlow<String> = _callId
+
+    private val _participants: MutableStateFlow<List<CallUser>> = MutableStateFlow(emptyList())
```
filbabic, line 106: We already have participantList, and now we also have CallUser - what is the difference/use case for these?

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -90,66 +95,109 @@ public class CallViewModel(
...
+    public val participants: StateFlow<List<CallUser>> = _participants
+
+    private val _isMicrophoneEnabled = MutableStateFlow(false)
```
filbabic: These should be moved up to permissions and videoEnabled and such

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -90,66 +95,109 @@ public class CallViewModel(
...
+                        _participants.value = state.users.values.toList()
+                    }
+                    is State.Starting -> {
+                        _callType.value = CallType.fromType(state.callGuid.type)
+                        _callId.value = state.callGuid.id
+                    }
+                    is State.Outgoing -> {
+                        _callType.value = CallType.fromType(state.callGuid.type)
+                        _callId.value = state.callGuid.id
+                        _participants.value = state.users.values
```
filbabic, line 132: We should probably call this something else then, not participants - as if we're in Incoming/Outgoing calls, we're not participanting yet, maybe members or invitees or idk

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -90,66 +95,109 @@ public class CallViewModel(
...
-            input.userToken,
-            input.iceServers,
-            credentialsProvider
-        )
-        _isVideoInitialized.value = true
-
-        initializeCall(callSettings = callSettings)
+        viewModelScope.launch {
+            logger.d { "[connectToCall] state: ${streamCallState.value}" }
+            withTimeout(30_000) {
```
filbabic: Pull out as a constant

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -90,66 +95,109 @@ public class CallViewModel(
...
+                val state = streamCallState.first { it is State.InCall } as State.InCall
+                logger.v { "[connectToCall] received: ${streamCallState.value}" }
+                client = streamVideo.createCallClient(
+                    state.callUrl.removeSuffix("/twirp"),
```
filbabic, line 161: ideally we'd not hardcode this but the BE should provide us with a sanitized URL, however we can maybe pull out as a util

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -259,4 +311,107 @@ public class CallViewModel(
...
+        logger.d { "[onMicrophoneChanged] microphoneEnabled: $microphoneEnabled" }
+        this._isMicrophoneEnabled.value = microphoneEnabled
+    }
+
+    public fun onVideoChanged(videoEnabled: Boolean) {
+        logger.d { "[onVideoChanged] videoEnabled: $videoEnabled" }
+        this._isCameraEnabled.value = videoEnabled
+    }
+
+    private fun State.Outgoing.toMetadata(): CallMetadata =
```
filbabic: Can be pulled out to a utility function rather than be placed here

app/src/main/kotlin/io/getstream/video/android/app/ui/home/HomeActivity.kt
kanat: Yeah, left after debugging, thanks for catching this

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
kanat: Yeah, kinda lazy way for lazy people 😁

`StreamVideoProvider` - looks good 👌

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
kanat: So basically they won't be able to override `onCreate` without calling `super`, otherwise, the app will crash with `android.util.SuperNotCalledException`, I bet they should be aware of this default OS behavior =]


As an option, they can always make their own Activity to make any UI they want 🤷‍♂️
For meanwhile lemme just create an overridable content building function.

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideo.kt
kanat: I was debugging the call flow and left it by mistake 🙈

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
kanat: will be removed

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
kanat: Since Engine is still something as WIP and we're supposed to support other events, so I just left the whole `when` options to see how many are left to support.
But we can replace it with `else`, I have no issues with that

stream-video-android/src/main/kotlin/io/getstream/video/android/input/CallAndroidInputLauncher.kt
kanat: It's the default behavior defined in `DefaultCallAndroidInputLauncher`, cause `Activity` can be launched from the appContext when PN arrives, for instance.

This can be easily overridden by providing a custom implementation of `CallAndroidInputLauncher` interface

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
kanat: I left it as it is for now, just merged what we had in all 3 flows (outgoing, incoming, in-call).

That's what I was thinking about as well, cause we have duplicate data in the following models:
- CallUser
- CallMember
- CallParticipant

Maybe we could leave just `CallUser` & `CallParticipant`, what do you think?

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
kanat: Not sure about this URL massaging, cause it was there before, so I just left it as it is =]

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
filbabic: That sounds good, thanks - we'll need an alternative for XML too, but should be doable in the same way.

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
filbabic: I'm fine with leaving it as-is until we reach a more "final" state

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
filbabic: Yeah that sounds good, the CallParticipant is when a user is in a call and has possible tracks published, whereas CallMember can be outside-of-the-call, but still a member

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
filbabic: Yeah I added this bit since the BE returns a URL with an extra path suffix, and retrofit hates that

We can pull this out to a utils function maybe

Issue #79, opened 2022-11-16
Track PeerConnection state changes within CallEngine


kanat:
`CallEngine` is required to track peer connection state changes to properly notify UI layer about connection changes

Issue #80, opened 2022-11-16
Merge CallMember and CallParticipant into CallUser


kanat:
We got repetitive data across the following structures: `CallMember`, `CallParticipant`, `CallUser`.

As agreed with other Android devs we gonna merge all of this into `CallUser`

PR #81, opened 2022-11-16
[79-80] add new states & merge models

kanat:
Closes: #79 and #80


stream-video-android/src/main/kotlin/io/getstream/video/android/call/CallClientImpl.kt
```diff
@@ -139,7 +142,7 @@ internal class CallClientImpl(
         }
     }
 
-    private val connectionConfiguration: PeerConnection.RTCConfiguration by lazy {
+    private val connectionConfiguration: RtcPeerConnection.RTCConfiguration by lazy {
```
filbabic: Any special reason why we import this differently? It doesn't seem to clash with anything

stream-video-android/src/main/kotlin/io/getstream/video/android/call/connection/StreamPeerConnection.kt
```diff
@@ -81,7 +85,7 @@ public class StreamPeerConnection(
     /**
      * The wrapped connection for all the WebRTC communication.
      */
-    public lateinit var connection: PeerConnection
+    public lateinit var connection: RtcPeerConnection
```
filbabic: Same here, it shouldn't be clashing

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -208,28 +216,79 @@ internal class StreamCallEngineImpl(
...
+        val eventUsers = event.callState.participants.toCallUserMap()
+        _callState.post(
+            state.copy(
+                users = state.users merge eventUsers,
+                sfuSessionId = StreamSfuSessionId.Confirmed(state.sfuSessionId.value)
+            )
+        )
+    }
+
+    private fun onSfuParticipantJoined(event: SfuParticipantJoinedEvent) = scope.launchWithLock(mutex) {
```
filbabic, line 245: Would love KDocs on all of these functions (even though they're private - for our sanity)

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -208,28 +216,79 @@ internal class StreamCallEngineImpl(
...
+        logger.d { "[onSfuParticipantJoined] event: $event" }
+        val state = _callState.value
+        if (state !is State.InCall) {
+            logger.w { "[onSfuParticipantJoined] rejected (state is not Connecting): $state" }
+            return@launchWithLock
+        }
+        // TODO event.call.id contains cid; should be fixed on BE
```
filbabic, line 252: can we report this to BE?

stream-video-android/src/main/kotlin/io/getstream/video/android/model/IceConnection.kt
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.model
+
+import org.webrtc.PeerConnection.IceConnectionState
+
+public enum class IceConnection {
```
filbabic: Docs on all of this would be good

stream-video-android/src/main/kotlin/io/getstream/video/android/model/PeerConnection.kt
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.model
+
+import org.webrtc.PeerConnection.PeerConnectionState
+
+public enum class PeerConnection {
```
filbabic: I don't think we should call this PeerConnection. It's not a connection on its own, we're typealiasing several things just to make it work and it's confusing out of the box.

We could've easily resolved this if we put this as an inner/static class for the StreamPeerConnection class:

```
class StreamPeerConnection {

    enum class ConnectionState {
    }
}
```

The import would've been longer, but it would easily say: `StreamPeerConnection.ConnectionState` or something. 

It could've also been just called: `StreamPeerConnectionState`, so that we can mirror what `org.webrtc` is doing:
`PeerConnectionState.NEW` from `org.webrtc.PeerConnection.PeerConnectionState`.

Otherwise it's confusing, duplicate names, using a name for something that it isn't (it's not a connection, it's the connection state) and it complicates oru codebase with typealiases

stream-video-android/src/main/kotlin/io/getstream/video/android/model/VideoModel.kt
```diff
@@ -18,21 +18,33 @@ package io.getstream.video.android.model
...
     val role: String,
+    val state: CallUserState,
     val imageUrl: String,
     val createdAt: Date?,
     val updatedAt: Date?,
     val teams: List<String>
 ) : Serializable
 
+public sealed class CallUserState {
+    public data class Specified(
```
filbabic: Maybe this can be called: `Active` or `Connected` and then the other value can be `Inactive`.

`Undefined` makes me think of undefined from JS, whereas the state here doesn't actually exist or rather it exists as inactive state (no state).

Alternatively this could be just a nullable property, so that we don't have to do type checks every time we want to access something:

```
val state = callUser.state

if (state is Specified) {
  // smart cast, do stuff
}

It could be just easier to say:

callUser.state?.run {
  // do something with the state
}

File: stream-video-android/src/main/kotlin/io/getstream/video/android/model/VideoModel.kt [hunk 8]
```diff
@@ -109,3 +122,64 @@ public fun ProtoCall?.toCallInfo(): CallInfo {
...
+
+public infix fun Map<String, CallUser>.merge(that: Map<String, CallUser>): Map<String, CallUser> {
+    val new = this - that.keys
+    val merged = this.map { (userId, user) ->
+        userId to user.merge(that[userId])
+    }.toMap()
+    return merged + new
+}
+
+public infix fun Map<String, CallUser>.merge(that: CallUser): Map<String, CallUser> = when (contains(that.id)) {
```
filbabic, line 176: I'd love kdocs over all of these to describe what and why we're merging the data and transforming it in such way.

Also - these are public, do we need to put them as public? And do we need to have them as infix since they're only supposed to be used here?

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -110,25 +106,59 @@ public sealed interface StreamCallState : Serializable {
...
-        public val iceServers: List<IceServer>,
-        public val sfuSessionId: StreamSfuSessionId,
-        public val sfuSessionJoined: Boolean
-    ) : Started()
+        override val callUrl: String,
+        override val sfuToken: SfuToken,
+        override val iceServers: List<IceServer>,
+    ) : InCall()
+
+    public data class Connecting(
```
filbabic, line 143: Kdocs please, this is getting more and more complex and to help everyone work on this part of the codebase, we need to document it thoroughly!

stream-video-android/src/main/kotlin/io/getstream/video/android/call/connection/StreamPeerConnection.kt
```diff
@@ -117,7 +121,7 @@ public class StreamPeerConnection(
     }
 
     /**
-     * Initialize a [StreamPeerConnection] using a WebRTC [PeerConnection].
+     * Initialize a [StreamPeerConnection] using a WebRTC [StreamPeerConnectionState].
```
filbabic:
```suggestion
     * Initialize a [StreamPeerConnection] using a WebRTC [PeerConnection].
```

stream-video-android/src/main/kotlin/io/getstream/video/android/call/connection/StreamPeerConnection.kt
```diff
@@ -336,7 +340,7 @@ public class StreamPeerConnection(
     }
 
     /**
-     * Triggered whenever there's a new negotiation needed for the active [PeerConnection].
+     * Triggered whenever there's a new negotiation needed for the active [StreamPeerConnectionState].
```
filbabic:
```suggestion
     * Triggered whenever there's a new negotiation needed for the active [PeerConnection].
```

stream-video-android/src/main/kotlin/io/getstream/video/android/call/connection/StreamPeerConnection.kt
```diff
@@ -358,7 +362,7 @@ public class StreamPeerConnection(
     /**
      * Triggered when the connection state changes.  Used to start and stop the stats observing.
      *
-     * @param newState The new state of the [PeerConnection].
+     * @param newState The new state of the [StreamPeerConnectionState].
```
filbabic:
```suggestion
     * @param newState The new state of the [PeerConnection].
```

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngine.kt
```diff
@@ -100,4 +102,6 @@ internal interface StreamCallEngine {
      * Called when [CallEventType] message is sent to Coordinator.
      */
     fun onCallEventSent(callCid: String, eventType: CallEventType)
+
+    fun onCallConnectionChange(sfuSessionId: String, peerType: StreamPeerType, connection: StreamPeerConnectionState)
```
filbabic: Maybe the last param should be: `connectionState`

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -208,28 +216,85 @@ internal class StreamCallEngineImpl(
...
+        _callState.post(
+            state.copy(
+                users = state.users merge eventUsers,
+                sfuSessionId = StreamSfuSessionId.Confirmed(state.sfuSessionId.value)
+            )
+        )
+    }
+
+    /**
+     * Called when participant joins to the existing call.
```
filbabic:
```suggestion
     * Called when participant joins an existing call.
```

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -208,28 +216,85 @@ internal class StreamCallEngineImpl(
...
+     */
```
filbabic:
```suggestion
     * @param event Contains information about the participant who joined. 
     */
```

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -208,28 +216,85 @@ internal class StreamCallEngineImpl(
...
+        val eventUser = event.participant.toCallUser()
+        _callState.post(
+            state.copy(
+                users = state.users merge eventUser
+            )
+        )
+    }
+
+    /**
+     * Called when participant leaves a call.
```
filbabic, line 270: Similar here, can we add the param description, even though it's not as hard to figure out what it is

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -485,25 +540,49 @@ internal class StreamCallEngineImpl(
...
         val state = _callState.value
         if (state !is State.Active) {
             logger.w { "[onCallFailed] rejected (state is not Active): $state" }
             return@launchWithLock
         }
-        logger.e { "[onCallFailed] error: $error, state: $state" }
         dropCall(State.Drop(state.callGuid, state.callKind, DropReason.Failure(error)))
     }
 
+    override fun onCallConnectionChange(
```
filbabic, line 567: Kdocs here (preferably with params)

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -110,33 +109,83 @@ public sealed interface StreamCallState : Serializable {
...
+    public sealed class InCall : Started() {
+        public abstract val callUrl: String
+        public abstract val sfuToken: SfuToken
+        public abstract val iceServers: List<IceServer>
+        public abstract val sfuSessionId: StreamSfuSessionId
+    }
+
+    /**
+     * Set when joined to Coordinator.
+     * [StreamCallState] stays [Joined] until [sfuSessionId] reaches [StreamSfuSessionId.Confirmed] inclusively.
```
filbabic:
```suggestion
     * [StreamCallState] stays [Joined] until [sfuSessionId] reaches [StreamSfuSessionId.Confirmed] explicitly.
```

filbabic: Not sure how inclusively would fit here, but maybe I misunderstood, I think explicitly might make more sense - it HAS to reach that state.

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -110,33 +109,83 @@ public sealed interface StreamCallState : Serializable {
...
+        override val createdAt: StreamDate,
+        override val updatedAt: StreamDate,
+        override val users: Map<String, CallUser>,
+        override val callUrl: String,
+        override val sfuToken: SfuToken,
+        override val iceServers: List<IceServer>,
+    ) : InCall()
 
+    /**
+     * Signifies that the call was drop because of the following [reason].
```
filbabic:
```suggestion
     * Signifies that the call was dropped because of the following [reason].
```

stream-video-android/src/main/kotlin/io/getstream/video/android/utils/stringify.kt
```diff
@@ -16,7 +16,14 @@
 
 package io.getstream.video.android.utils
 
+import io.getstream.video.android.model.StreamPeerType
+
 public fun <T : Any> Result<T>.stringify(toString: (T) -> String): String = when (this) {
```
filbabic: Oh - can we push this into a StringUtils file maybe? `stringify` is kind of web/BE naming for files, being lowercase, and it might make more sense to just have these bundled in a StringUtils

PR #93, opened 2022-11-17
Maven/gradle setup

filbabic:
Re-opened on dev

### DIFF
--- a/docusaurus/.env
+++ b/docusaurus/.env
@@ -1,0 +1,1 @@
+PRODUCT=video
--- a/docusaurus/docs/Android/01-basics/02-introduction.mdx
+++ b/docusaurus/docs/Android/01-basics/02-introduction.mdx
@@ -1,3 +1,7 @@
+---
+slug: /
+---
+
 ## Introduction
 
 The [Stream Video Android SDK](https://github.com/GetStream/stream-video-android) enables you to easily build any type of Video and Audio call experiences in your Android apps, either in Kotlin or Java.


.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,34 @@
+name: Publish Snapshot builds
+
+on:
+  push:
+    branches:
+      - main
```
JcMinarro: Snapshot shouldn't be started from `develop` branch? :thinking:

skydoves: Nice catch! We should. Actually, the `develop` branch was created after making this PR 😄

scripts/publish-module.gradle
```diff
@@ -0,0 +1,99 @@
...
+                            name = 'Jaewoong Eum'
+                            email = "jaewoong.eum@getstrea.io"
+                            url = "https://github.com/skydoves"
+                        }
+                        developer {
+                            id = 'kanat'
+                            name = 'Kanat Kiialbaev'
+                            email = "kanat.kiialbaev@getstream.io"
+                            url = "https://github.com/kanat"
+                        }
```
filbabic, line 79:
```suggestion
                        }
                        developer {
                            id = 'TGazica'
                            name = 'Tomislav Gazica'
                            email = "tomislav.gazica@getstream.io"
                            url = "https://github.com/TGazica"
                        }
```

Issue #60, opened 2022-11-03
Enable a way for users to join calls without any permissions


filbabic:
Currently we require permissions for our users to join the call, but it should be possible to join it based without audio or video.

We should then add handlers for A/V permissions when the user attempts to enable such features through the UI.

Note: This also depends on if we'll ship the UI controls for A/V in the SDK.

PR #103, opened 2022-11-18
[102] Detach CallClient lifecycle from the VM lifecycle

### LINKED ISSUE #102: Detach CallClient lifecycle from the VM lifecycle
`CallClient` is currently getting killed when VM gets killed. 
We should make it live as long as call lives, since call can be continued in background.

kanat:
Closes #102

### DIFF
--- a/stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
+++ b/stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
@@ -444,10 +444,7 @@ public class StreamVideoImpl(
         iceServers: List<IceServer>
     ): CallClient {
         logger.i { "[createCallClient] signalUrl: $signalUrl, sfuToken: $sfuToken, iceServers: $iceServers" }
-        /**
-         * TODO [sfuToken] can be set as soon as we get [JoinedCall] in [joinCallInternal] function.
-         */
-        credentialsProvider.setSfuToken(sfuToken)
+
         return CallClientBuilder(
             context = context,
             credentialsProvider = credentialsProvider,


stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
```diff
@@ -133,6 +137,26 @@ public class StreamVideoImpl(
                             cancelCall(state.callGuid.cid)
                         }
                     }
+                    is StreamCallState.Idle -> {
+                        clearCallState()
+                    }
```
filbabic:
```suggestion
                    is StreamCallState.Idle -> clearCallState()
```

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -63,39 +59,13 @@ public class CallViewModel(
...
-        }
-
-    private val _callMediaState: MutableStateFlow<CallMediaState> =
-        MutableStateFlow(
-            CallMediaState(
-                isMicrophoneEnabled = isAudioEnabled.value && hasAudioPermission.value,
-                isCameraEnabled = isVideoEnabled.value && hasVideoPermission.value,
-                isSpeakerphoneEnabled = false
-            )
-        )
```
filbabic: Why did we remove this entire block? We will need this information to know if we can show the camera/turn on the mic

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -166,68 +135,48 @@ public class CallViewModel(
...
             withTimeout(CONNECT_TIMEOUT) {
-                val state = streamCallState.first { it is State.InCall } as State.InCall
                 logger.v { "[connectToCall] received: ${streamCallState.value}" }
-                client = streamVideo.createCallClient(
-                    state.callUrl.removeSuffix("/twirp"),
-                    state.sfuToken,
-                    state.iceServers,
-                )
+                val client = streamVideo.awaitCallClient().also {
+                    client = it
```
filbabic:
```suggestion
                    this.client = it
```

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -151,7 +121,6 @@ public class CallViewModel(
                         _participants.value = state.users.values
                             .filter { it.id != streamVideo.getUser().id }
                             .toList()
-                        joinCall()
                     }
                     is State.Joining -> {
```
filbabic, line 154: Maybe these remaining states can just be ignored together? or using an else block

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -166,68 +135,48 @@ public class CallViewModel(
...
             logger.d { "[connectToCall] state: ${streamCallState.value}" }
             withTimeout(CONNECT_TIMEOUT) {
-                val state = streamCallState.first { it is State.InCall } as State.InCall
                 logger.v { "[connectToCall] received: ${streamCallState.value}" }
-                client = streamVideo.createCallClient(
-                    state.callUrl.removeSuffix("/twirp"),
-                    state.sfuToken,
-                    state.iceServers,
-                )
+                val client = streamVideo.awaitCallClient().also {
```
filbabic: I personally don't like this `awaitCallClient` and it being automatically created based on our state. It feels like there might be race conditions or different states in which this will crash.

For example, we only start our internal screens based on the state, where we create the CallViewModel after the client has been created, but customers can create the ViewModel at any point in time, with a `StreamVideo` and `PermissionsManager` instance and they can call `connectToCall` whenever they want.

With this, there is a possibility that they'll create the ViewModel and attempt to `connectToCall` but the CallClient doesn't exist, so the await will either suspend indefinitely or crash.

Additionally, if the user attempts to manually create the CallClient now without following the Join Flow (e.g. if they have UI to directly connect to the client, the app can crash as the token is not set in the credentials manager.

There's going to be more and more of these pitfalls for our customers unless they 100% follow the flow that we provide for them.

The whole call engine and state automation, out of the box initializers etc is becoming more and more convoluted and complex to teach to someone.

Right now - if people use our default launchers and everything (and it's hard and complex to teach them to build it custom as everything is required and using default is kind of forced) and they have no control over the lifecycle, initialization order and times, but they can mess up by calling functions which are public that do the initialization themselves. E.g. they can try to `createCallClient` even though it's been internally created - wasting resources and clogging up memory/CPU power.

Alternatively, they can create the VM and call public functions that crash because they went to work around the state.

This is quite possible if users don't want to use our UI componetns and the default launchers and want to build their own UI and logic.

I also think this will be extremely hard to teach to our customers - this doesn't seem like SDK material, it more looks like it's built _specifically_ for our sample app and the way it looks and works, unlike our chat components.

We should provide:
- Easy out of the box components that can be used or avoided
- Easy alternatives that use the same flow, but are not constrained

Right now we do have out of the box components, but it seems like you will have to use them, otherwise it's too much work to customize and that's bad developer experience.

Take for example the StreamVideoBuilder:

```kotlin
public class StreamVideoBuilder(
    private val context: Context,
    private val credentialsProvider: CredentialsProvider,
    private val config: StreamVideoConfig = StreamVideoConfigDefault,
    private val androidInputs: Set<CallAndroidInput> = emptySet(),
    private val inputLauncher: CallAndroidInputLauncher = DefaultCallAndroidInputLauncher,
    private val loggingLevel: LoggingLevel = LoggingLevel.NONE
)
```

You have to teach customers about the `androidInputs` and what the `AndroidInput` is. Also I just realized it's a sealed class, meaning you can't override you can only provide instances of Activity or Service inputs. What if they want to start something else or have different kind of logic?

Then they have to learn about the `CallAndroidInputLauncher` - which they have to override and implement custom behavior of, then plug in the inputs to that behavior, to achieve that custom behavior. Or override as empty if they want to disable some behavior (the automatic activity launches and all).

They have to do all of this just to avoid our UI, which seems like a lot of work and many things they have to learn about. It'll be very hard to teach this in guides/the tutorial and explain the CallEngine, its state, the StreamVideo state, the CallClient creation and CallViewModel.

Especially if people want to build things in a custom way. We should really try to build custom UI and implementation with this and see if it makes sense because for me it seems like it might be too complex and too constricting of an API. 🤔

It might be just me and looking at this PR and all the changes to the engine and the flow, but it surely feels complex to me, so I imagine it'll be equally (if not more) complex to our customers.

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
kanat: `callMediaState` is still there.
I found like we're duplicating things, since `callMediaState` already contains everything

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
kanat: > With this, there is a possibility that they'll create the ViewModel and attempt to connectToCall but the CallClient doesn't exist, so the await will either suspend indefinitely or crash.

In order to create `CallClient` is required to have `signalUrl`, `sfuToken`, `iceServers`. Those are available only when our SDK is joined to Coordinator. So customers won't be able to create `CallClient` whenever they want, since they will be restricted by the requirement mentioned above, meaning that they have to complete the flow prior `connectToCall`

```kotlin
    public fun createCallClient(
        signalUrl: String,
        sfuToken: SfuToken,
        iceServers: List<IceServer>
    ): CallClient
```

> Right now we do have out of the box components, but it seems like you will have to use them, otherwise it's too much work to customize and that's bad developer experience.

`androidInputs` is empty by default, which means none of our components (activity or service) will be launched by default. Customers may omit doing this and launch any component they want.
Let's say you want a fully custom UI and you don't want our launchers, so just keep `androidInputs` empty and reuse our `CallViewModel`, that's it.
We don't force them to use something, we just provide ways to make it simpler for those who don't want lots of customization.


In the most simple case customers will need to do the following steps to have a full functional calls:
- extend `AbstractStreamCallService` and implement `getStreamVideo` function
- extend `AbstractComposeCallActivity` and implement `getStreamVideo` function
- initialize `StreamVideo` with inputs above

Then the `caller/meeting_creator` starts a `call/meeting` and everything is gonna work out of the box. It is a bare minimum to make it work.

filbabic: Alright - Looking into this, it makes more sense to be easily ignored, but we will have to provide extra content on how to set these up or to customize the UI.

Once we clear up the rest of the comments and discuss the automatic side with the rest of the SDKs, we'll have a better picture of the level of documentation we need to explain our SDK to customers and what the best approach to it is.

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
filbabic: The reason why we wanted to have a complex combination of permissions + user's video/audio opt in is because there will be case of the user allowing a permission but not allowing A/V, or them trying to publish A/V but not having permisisons.

This is just going to be a part of the permissions refactor that Tomislav is working on - to allow users to join calls without permissions.

We'll have to observe both if the user allowed the A/V and if they have permissions, to treat the UI actions different

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/AbstractComposeCallActivity.kt
```diff
@@ -54,24 +57,10 @@ public abstract class AbstractComposeCallActivity : AppCompatActivity(), StreamV
...
-            deniedCamera || deniedMicrophone -> showPermissionsDialog()
-            else -> {
-                checkPermissions()
-            }
-        }
-    }
+    private val requiredPermissions = arrayOf(
+        Manifest.permission.CAMERA,
+        Manifest.permission.RECORD_AUDIO
+    )
```
TGazica: I would leave this as is for now and just request permission if you need it for ease of use while testing. We will need to use registerForActivityResult, onPermissionResult is deprecated, and the internal implementation for it is suppressing the deprecation warning. Proper way to request will be implemented in #60

kanat: lemme undo the permissions changes, so we could both rely on the changes from your branch 👍

stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoImpl.kt
```diff
@@ -427,34 +443,43 @@ public class StreamVideoImpl(
         sfuToken: SfuToken,
         iceServers: List<IceServer>
     ): CallClient {
+        logger.i { "[createCallClient] signalUrl: $signalUrl, sfuToken: $sfuToken, iceServers: $iceServers" }
+        /**
+         * TODO [sfuToken] can be set as soon as we get [JoinedCall] in [joinCallInternal] function.
```
filbabic: I'll apply this!

PR #110, opened 2022-11-23
Task/call permissions

TGazica:
Adds the ability to join calls without granted permissions. Permissions can be granted mid-call or when ringing.
Refactors handling states in CallViewModel.

### DIFF
--- a/stream-video-android/src/main/kotlin/io/getstream/video/android/call/CallClientImpl.kt
+++ b/stream-video-android/src/main/kotlin/io/getstream/video/android/call/CallClientImpl.kt
@@ -661,7 +661,7 @@ internal class CallClientImpl(
         logger.v { "[createUserTracks] #sfu; audioTrack: ${audioTrack.stringify()}" }
 
         if (autoPublish) {
-            publisher?.addTrack(localAudioTrack!!, listOf(sessionId))
+            publisher?.addAudioTransceiver(localAudioTrack!!, listOf(sessionId))
         }
     }
 
--- a/stream-video-android/api/stream-video-android.api
+++ b/stream-video-android/api/stream-video-android.api
@@ -763,6 +763,13 @@ public abstract interface class io/getstream/video/android/api/ClientRPCService 
 	public abstract fun upsertCallMembers (Lstream/video/coordinator/client_v1_rpc/UpsertCallMembersRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
+public abstract interface class io/getstream/video/android/api/EgressRPCService {
+	public abstract fun startBroadcast (Lstream/video/coordinator/server_v1_rpc/StartBroadcastRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public abstract fun startRecording (Lstream/video/coordinator/server_v1_rpc/StartRecordingRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public abstract fun stopBroadcast (Lstream/video/coordinator/server_v1_rpc/StopBroadcastRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public abstract fun stopRecording (Lstream/video/coordinator/server_v1_rpc/StopRecordingRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+}
+
 public abstract interface class io/getstream/video/android/api/GossipRPCService {
 	public abstract fun getCallState (Lstream/video/sfu/gossip/GetCallStateRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public abstract fun joinCall (Lstream/video/sfu/gossip/JoinCallRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
--- a/stream-video-android/src/main/kotlin/io/getstream/video/android/PermissionManagerProvider.kt
+++ b/stream-video-android/src/main/kotlin/io/getstream/video/android/PermissionManagerProvider.kt
@@ -20,10 +20,17 @@ import io.getstream.video.android.permission.PermissionManager
 
 public interface PermissionManagerProvider {
 
+    /**
+     * Used for [PermissionManager] setup and registration with the Activity that it's used in.
+     *
+     * @return The initialized and registered [PermissionManager].
+     */
     public fun initPermissionManager(): PermissionManager
 
     /**
-     * Allows for a custom implementation of the [PermissionManager].
+     * Allows fetching of the [PermissionManager] instance created by [initPermissionManager].
+     *
+     * @return The initialized [PermissionManager] instance.
      */
     public fun getPermissionManager(): PermissionManager
 }
--- a/stream-video-android/src/main/kotlin/io/getstream/video/android/call/state/CallAction.kt
+++ b/stream-video-android/src/main/kotlin/io/getstream/video/android/call/state/CallAction.kt
@@ -31,6 +31,9 @@ public data class ToggleSpeakerphone(
     val isEnabled: Boolean
 ) : CallAction()
 
+/**
+ * Action to select an audio device for playback.
+ */
 public data class SelectAudioDevice(
     val audioDevice: AudioDevice
 ) : CallAction()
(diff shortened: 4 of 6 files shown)


dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
```diff
@@ -103,51 +101,28 @@ class CallActivity : AppCompatActivity() {
         callViewModel.connectToCall(
             CallSettings(
                 audioOn = false,
```
filbabic: Should be initialized through the hasAudio/Microphone permission

dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
```diff
@@ -19,73 +19,71 @@ package io.getstream.video.android.dogfooding
...
+import io.getstream.video.android.compose.theme.VideoTheme
 import io.getstream.video.android.compose.ui.components.call.activecall.ActiveCallContent
 import io.getstream.video.android.model.CallSettings
-import io.getstream.video.android.permission.PermissionManagerImpl
+import io.getstream.video.android.permission.PermissionManager
+import io.getstream.video.android.permission.StreamPermissionManagerImpl
 import io.getstream.video.android.viewmodel.CallViewModel
 import io.getstream.video.android.viewmodel.CallViewModelFactory
 
 class CallActivity : AppCompatActivity() {
```
filbabic: Since this branch is a bit out of date with Kanat's changes - these will probably go into the AbstractActivity

app/src/main/kotlin/io/getstream/video/android/app/ui/call/CallActivity.kt
```diff
@@ -36,10 +36,10 @@ class CallActivity : AbstractComposeCallActivity() {
     /**
      * Provides a custom factory for the ViewModel, that provides fake users for invites.
      */
-    override fun getCallViewModelFactory(): CallViewModelFactory {
+    override fun getCallViewModelFactory(permissionManager: PermissionManager): CallViewModelFactory {
```
filbabic: Instead of providing it as a parameter, can we expose a function that builds the default instance? That way we don't have to build the manager in case our users don't want to use it and want to rely on a custom impl

stream-video-android/src/main/kotlin/io/getstream/video/android/call/CallClient.kt
```diff
@@ -30,7 +30,7 @@ public interface CallClient {
 
     public suspend fun connectToCall(
         sessionId: String,
-        callSettings: CallSettings
+        getCallSettings: () -> CallSettings
```
filbabic: I've removed this in a branch with my impl for chat + video - I think we can just store the data within the ViewModel actually and then pass it down. (As we initially discussed). 

Then users can provide the state to the VM or change the state from the outside, while still being able to pass it down to the StreamVideo/CallClient directly

stream-video-android/src/main/kotlin/io/getstream/video/android/call/CallClientImpl.kt
```diff
@@ -589,22 +597,33 @@ internal class CallClientImpl(
...
-            publisher?.addTrack(localAudioTrack!!, listOf(sessionId))
             publisher?.addVideoTransceiver(localVideoTrack!!, listOf(sessionId))
         }
     }
 
+    private fun setupAudioTrack(autoPublish: Boolean = true) {
+        if (localAudioTrack != null) return
+
+        val audioTrack = makeAudioTrack()
+        audioTrack.setEnabled(true)
```
filbabic: should be delegated from `callSettings.hasAudio`

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -165,40 +174,45 @@ public class CallViewModel(
         }
     }
 
-    public fun connectToCall(callSettings: CallSettings) {
-        logger.d { "[connectToCall] input: $callSettings" }
+    public fun connectToCall(initialCallSettings: CallSettings) {
+        logger.d { "[connectToCall] input: $initialCallSettings" }
+
+        isAudioEnabled.value = initialCallSettings.audioOn && permissionManager.hasRecordAudioPermission.value
```
filbabic: we set this but then initialize the call with the initial settings, can we just use these values instead?

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -165,40 +174,45 @@ public class CallViewModel(
...
                 client = streamVideo.awaitCallClient()
                 _isVideoInitialized.value = true
-                initializeCall(callSettings = callSettings)
+                initializeCall(initialCallSettings)
             }
         }
     }
 
-    private suspend fun initializeCall(callSettings: CallSettings) {
+    private suspend fun initializeCall(initialCallSettings: CallSettings) {
```
filbabic: or alternatively just use autoPublish here

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -208,21 +222,11 @@ public class CallViewModel(
         }
     }
 
-    public fun toggleSpeakerphone(enabled: Boolean) {
-        client.setSpeakerphoneEnabled(enabled)
+    private fun toggleSpeakerphone(enabled: Boolean) {
+        if (::client.isInitialized) client.setSpeakerphoneEnabled(enabled)
```
filbabic: Also on my branch I've added some checks for this as it used to crash with the engine flow in outgoing/incoming screens

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
```diff
@@ -208,21 +222,11 @@ public class CallViewModel(
...
         onSpeakerphoneChanged(enabled)
     }
 
-    public fun toggleCamera(enabled: Boolean) {
```
filbabic, line 211: Any reason these are removed?

dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
TGazica: I can add it if we want to enable it by default.

dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
TGazica: This would require us to change which class the `CallActivity` extends. Currently, it is `AppCompatActivity`, should I change it to `AbstractComposeCallActivity`?

filbabic: Yes I think in the examples we have it's also AbstractComposeActivity - we will eventually provide an XML variant, but all the base logic for an out of the box experience should go in base classes if possible

dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
filbabic: oh, fair enough - we probably don't want audio on by default for echo reasons, leave as-is

app/src/main/kotlin/io/getstream/video/android/app/ui/call/CallActivity.kt
TGazica: I can refactor it. We also need to make sure that the `registerForActivityResult` is called before the Activity/Fragment has entered STARTED state, so there would be one init function and one get function.

stream-video-android/src/main/kotlin/io/getstream/video/android/viewmodel/CallViewModel.kt
TGazica: we would have this and a private `onCameraChanged` method. We also have `onCallAction` which uses actions to change these. I think it is redundant to have multiple ways to change the same thing, as it automatically makes the API larger and we need to maintain more logic. I think we should use `onCallAction` and all of the actions should go through it, and it will call private VM methods. It reduces the number of lambdas we need to pass to composables and we get a smaller public API while keeping all of the functionalities.

dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
TGazica: refactored inside #119

PR #112, opened 2022-11-24
[111] outline engine state flow with Coordinator & SFU only

### LINKED ISSUE #111: Outline engine state flow with Coordinator & SFU only

kanat:
Closes #111

Simplifies state flow & removes `PeerConnection` tracking from `CallEngine`. 

Main changes:
- removed `Starting` state
- `Outgoing` flow is fixed (not getting set for Meetings)
- state stays `Idle` when you create call
- `Joining` = joining to Coordinator
- `Joined` = joined to Coordinator
- `Connecting` = connecting/joining to SFU (keep it connecting to distinguish from Coordinator joining)
- `Connected` = connected/joined to SFU

### DIFF
--- a/stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
+++ b/stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
@@ -22,6 +22,7 @@ import io.getstream.video.android.model.IceServer
 import io.getstream.video.android.model.SfuToken
 import io.getstream.video.android.model.StreamCallGuid
 import io.getstream.video.android.model.StreamCallKind
+import io.getstream.video.android.model.StreamSfuSessionId
 import java.io.Serializable
 import java.util.Date
 
@@ -47,15 +48,6 @@ public sealed interface StreamCallState : Serializable {
         public abstract val callGuid: StreamCallGuid
         public abstract val callKind: StreamCallKind
     }
-
-    /**
-     * Signifies the caller starts the call.
-     */
-    public data class Starting(
-        override val callGuid: StreamCallGuid,
-        override val callKind: StreamCallKind,
-        val memberUserIds: List<String>
-    ) : Active(), Joinable
 
     public sealed class Started : Active() {
         public abstract val createdByUserId: String
@@ -137,7 +129,7 @@ public sealed interface StreamCallState : Serializable {
      * @see [io.getstream.video.android.engine.StreamCallEngine.onSfuJoinSent]
      */
     public data class Connecting(
-        val sfuSessionId: String,
+        val sfuSessionId: StreamSfuSessionId,
         override val callGuid: StreamCallGuid,
         override val callKind: StreamCallKind,
         override val createdByUserId: String,
@@ -157,7 +149,7 @@ public sealed interface StreamCallState : Serializable {
      * @see [io.getstream.video.android.engine.StreamCallEngine.onSfuEvent]
      */
     public data class Connected(
-        val sfuSessionId: String,
+        val sfuSessionId: StreamSfuSessionId,
         override val callGuid: StreamCallGuid,
         override val callKind: StreamCallKind,
         override val createdByUserId: String,
--- a/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/internal/ActiveCallAppBar.kt
+++ b/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/internal/ActiveCallAppBar.kt
@@ -95,6 +95,5 @@ private fun StreamCallState.formatAsTitle() = when (this) {
     is StreamCallState.Incoming -> "Incoming"
     is StreamCallState.Joining -> "Joining"
     is StreamCallState.Outgoing -> "Outgoing"
-    is StreamCallState.Starting -> "Starting"
     StreamCallState.Idle -> "Idle"
 }
--- a/stream-video-android/api/stream-video-android.api
+++ b/stream-video-android/api/stream-video-android.api
@@ -2347,21 +2347,6 @@ public abstract class io/getstream/video/android/model/state/StreamCallState$Sta
 	public abstract fun getUsers ()Ljava/util/Map;
 }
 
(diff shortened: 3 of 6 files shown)


stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -138,29 +132,13 @@ public sealed interface StreamCallState : Serializable {
...
-    ) : InCall()
-
-    /**
-     * Signifies that one of the peer connections (subscriber or publisher) is in Connecting state.
+     * Set when SFU join request is sent.
+     * @see [stream.video.sfu.event.JoinRequest]
+     * @see [io.getstream.video.android.engine.StreamCallEngine.onSfuJoinSent]
      */
     public data class Connecting(
+        val sfuSessionId: String,
```
TGazica: Is there a reason it is now a `String` type instead of `StreamSfuSessionId`, so it is aligned with the rest?

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
```diff
@@ -174,11 +152,13 @@ public sealed interface StreamCallState : Serializable {
     ) : InCall()
 
     /**
-     * Signifies that one of the peer connections (subscriber or publisher) is in Connected state.
+     * Set when SFU join response is received.
+     * @see [io.getstream.video.android.events.JoinCallResponseEvent]
+     * @see [io.getstream.video.android.engine.StreamCallEngine.onSfuEvent]
      */
     public data class Connected(
+        val sfuSessionId: String,
```
TGazica: Same here

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
filbabic: Does it also make sense to maybe remove teh StreamSfuSessionId if we're just naming variables like that

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
```diff
@@ -391,55 +381,44 @@ internal class StreamCallEngineImpl(
...
+        if (state !is State.Idle) {
             logger.w { "[onCallStarted] rejected (state is not Starting): $state" }
             return@launchWithLock
         }
-        if (state.callGuid.cid != call.cid) {
-            logger.w {
-                "[onCallStarted] rejected (callCid is not valid); expected: ${state.callGuid.cid}, actual: ${call.cid}"
-            }
+        if (call.kind != StreamCallKind.RINGING) {
+            logger.w { "[onCallStarted] rejected (call.kind is not RINGING)" }
```
filbabic, line 410:
```suggestion
            logger.w { "[onCallStarted] ignored (call.kind is not RINGING) to answer" }
```

This might make a bit more sense, because we're not technically rejecting the call, we're just not accepting it., or we're ignoring the start event.

Can the start event trigger this function throuhg, if we're not ringing (the event shouldn't arrive)

stream-video-android/src/main/kotlin/io/getstream/video/android/model/StreamCallCid.kt
```diff
@@ -16,10 +16,24 @@
 
 package io.getstream.video.android.model
 
+/**
+ * Represents call cid.
+ */
 public typealias StreamCallCid = String
```
filbabic, line 22: As mentioned we could remove these if they act as extra overhead to the SDK, if it's clear enough through the property names what we're dealing with. IDK if htere is overhead but just wondering

What do you think?

stream-video-android/src/main/kotlin/io/getstream/video/android/model/StreamCallKind.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import io.getstream.video.android.model.StreamCallKind.MEETING
+import io.getstream.video.android.model.StreamCallKind.RINGING
+import java.io.Serializable
+
+/**
+ * The kind of call, either a [RINGING] or a [MEETING].
+ */
+public enum class StreamCallKind : Serializable {
+    MEETING, RINGING;
```
filbabic, line 27: We should double check with the team if these are the names of the call types, or if it's "conferencing", "ringing", "conferencing", "regular", " meeting", "regular" or something else

stream-video-android/src/main/kotlin/io/getstream/video/android/model/StreamSfuSessionId.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.model
+
+/**
+ * Represents SFU session id.
+ */
+public typealias StreamSfuSessionId = String
```
filbabic, line 22: unused as tomislav mentioned

stream-video-android/src/main/kotlin/io/getstream/video/android/model/state/StreamCallState.kt
kanat: Thanks for catching that. Fixed.

Basically I've added those type aliases for easier navigation in IDE (find usage)

stream-video-android/src/main/kotlin/io/getstream/video/android/model/StreamCallCid.kt
kanat: I would keep those type aliases for easier navigation in IDE (find usage)

stream-video-android/src/main/kotlin/io/getstream/video/android/model/StreamSfuSessionId.kt
kanat: fixed

stream-video-android/src/main/kotlin/io/getstream/video/android/engine/StreamCallEngineImpl.kt
kanat: I've treated it as the normal function flow got `rejected` at this place. So not something related to calls glossary.

So for the sack of consistency I use `rejected` everywhere to scan logs easily.

PR #120, opened 2022-11-28
Add static persistence of user credentials to the SDK

filbabic:
### 🎯 Goal

Added a way to persist user credentials within the SDK in a static way, so that we can use the credentials for our push notification integration. 

### 🛠 Implementation details

Object exposed: UserCredentialsManager that allows for initialization and accessing the underlying UserPreferences.

Users can use the app or any other contex to create and access it.

#### Case 1

1. Run the dogfooding/sample app.
2. Log in. 
3. Kill the app.
4. Run the app again.

Result: User should be logged in and the second time it won't ask for credentials/logging in.

#### Case 2

1. Run the dogfooding/sample app.
2. Log in. 
3. Kill the app.
4. Run the app again.
5. Log out & kill the app.
6. Open app.

Result: User should be logged out and the app wil l ask for credentials/logging in.

#### General

#### Code & documentation
- [x] Affected documentation updated (KDocs, docusaurus, tutorial)
- [x] Tutorial starter kit updated
- [x] Examples/guides starter kits updated (`stream-video-examples`)

### ☑️Reviewer Checklist

### DIFF
--- a/dogfooding/src/main/java/io/getstream/video/android/dogfooding/LoginActivity.kt
+++ b/dogfooding/src/main/java/io/getstream/video/android/dogfooding/LoginActivity.kt
@@ -215,14 +215,13 @@ class LoginActivity : ComponentActivity() {
     }
 
     private fun startHome(token: String, user: User) {
-        dogfoodingApp.initializeStreamCalls(
+        dogfoodingApp.initializeStreamVideo(
             AuthCredentialsProvider(
                 "key10", token,
                 user = user
             ),
             loggingLevel = LoggingLevel.BODY
         )
-        dogfoodingApp.userPreferences.storeUserCredentials(user)
         startActivity(HomeActivity.getIntent(this))
     }
 
--- a/app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
+++ b/app/src/main/kotlin/io/getstream/video/android/app/VideoApp.kt
@@ -56,7 +56,7 @@ class VideoApp : Application() {
     /**
      * Sets up and returns the [streamVideo] required to connect to the API.
      */
-    fun initializeStreamCalls(
+    fun initializeStreamVideo(
         credentialsProvider: CredentialsProvider,
         loggingLevel: LoggingLevel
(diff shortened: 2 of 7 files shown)


app/src/main/kotlin/io/getstream/video/android/app/user/UserWrapper.kt
```diff
@@ -16,13 +16,9 @@
...
 
-interface UserPreferences {
-
-    fun getCachedCredentials(): UserCredentials
-
-    fun storeUserCredentials(userCredentials: UserCredentials)
-
-    fun clear()
-}
+data class UserWrapper(
```
TGazica: Maybe `AppUser` or something else that better describes the usage?

filbabic: It's only for internal app usage, not exposed to the customer, so I think it should be okay, but LMK if you'd still like to rename

dogfooding/src/main/java/io/getstream/video/android/dogfooding/LoginActivity.kt
```diff
@@ -215,18 +214,15 @@ class LoginActivity : ComponentActivity() {
...
+    private fun startHome(token: String, user: User) {
         dogfoodingApp.initializeStreamCalls(
             AuthCredentialsProvider(
                 "key10", token,
                 user = user
             ),
             loggingLevel = LoggingLevel.BODY
         )
-        dogfoodingApp.userPreferences.storeUserCredentials(user.toCredentials())
+        dogfoodingApp.userPreferences.storeUserCredentials(user)
```
JcMinarro: Can't we have it internally in our SDK?
I mean, When an user is set into our VideoSDK, we could obtain the `userPreferences` (if it was set by our customers) and then store the user.
It will simplify how our customers integrate our SDK

filbabic: Added a way to do this automatically!

PR #137, opened 2022-12-02
Handle PN

JcMinarro:
### 🎯 Goal
Handle PN and show IncomingCall Notification in case the needed activities are provided.
In this [branch](https://github.com/GetStream/stream-video-android/tree/incoming-activities-sample) you have an example about how add an activity that is used by this implementation.

Basically, what you need to do to handle a this kind of PN is to provide a couple of activities that can be started by our actions:
```
<activity
    android:name=".AcceptCallActivity"
    android:exported="false"
    >

    <!-- We use negative priority to allow our customer to use their own one -->
    <intent-filter
        android:priority="-1"
        >
        <action android:name="io.getstream.video.android.action.ACCEPT_CALL" />
    </intent-filter>
</activity>
<activity
    android:name=".IncomingCallActivity"
    android:exported="false"
    >

    <!-- We use negative priority to allow our customer to use their own one -->
    <intent-filter
        android:priority="-1"
        >
        <action android:name="io.getstream.video.android.action.INCOMING_CALL" />
    </intent-filter>
</activity>
``` 

If you want to test the PN without needed to receive a PN from backend, you can run the following commands:
```
adb root

adb shell am broadcast -a com.google.android.c2dm.intent.RECEIVE -n io.getstream.video.android.dogfooding.debug/com.google.firebase.iid.FirebaseInstanceIdReceiver -a "com.google.android.c2dm.intent.RECEIVE" --es "sender" "stream.video" --es "type" "call_incoming" --es "call_cid" "call_123" --es "user_names" "Jc\,\ Filip"
```

Remember you need to provide your own activities like in the sample branch [I pushed](https://github.com/GetStream/stream-video-android/tree/incoming-activities-sample)

#### General
- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] Assigned a person / code owner group (required)
- [x] Thread with the PR link started in a respective Slack channel (required internally)
- [x] PR targets the `develop` branch
- [x] PR is linked to the GitHub issue it resolves

#### Code & documentation
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (KDocs, docusaurus, tutorial)
- [ ] Tutorial starter kit updated
[...]

PR #142, opened 2022-12-07
Feature/push notification guide and cleanup

### LINKED ISSUE #48: Push Notification integration guide
Write a guide that helps people integrate push providers for our video SDK and lets them integrate custom handlers and states.

filbabic:
### 🎯 Goal

Resolves #48 

Added some ease of use API for the push notification integration, cleaned up the API and wrote a guide on PN integration.

### 🛠 Implementation details

See #137 about the core implementation of push and its integration, how to test it etc. This PR just adds an easy way to integrate a push notification consuming activity as well as a guide that helps users to set it up.

### 🎨 UI Changes

See the documentation for the UI changes/features.

### 🧪 Testing

See #137

#### General
- [x] PR is linked to the GitHub issue it resolves

#### Code & documentation
- [x] Affected documentation updated (KDocs, docusaurus, tutorial)

### ☑️Reviewer Checklist
- [ ] XML sample runs & works


docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
```diff
@@ -0,0 +1,81 @@
...
+class NotificationActivity : AbstractNotificationActivity() {
+
+    override fun getStreamVideo(context: Context): StreamVideo {
+        return // TODO - provide video
+    }
+
+    override fun initializeVideoIfNeeded() {
+        // TODO - initialize SDK
+    }
+}
```
skydoves, line 59: It would be great if we can inform them that they make sure to override the `onCreate` function:

```kotlin
        super.onCreate(savedInstanceState)
}
```

stream-video-android/src/main/kotlin/io/getstream/video/android/notifications/AbstractNotificationActivity.kt
```diff
@@ -0,0 +1,103 @@
...
+ */
+public abstract class AbstractNotificationActivity : AppCompatActivity(), StreamVideoProvider {
+
+    private val streamVideo: StreamVideo by lazy { getStreamVideo(this) }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        initializeVideoIfNeeded()
+        processNotificationData()
+    }
```
skydoves, line 45: What do you think using `addOnContextAvailableListener` instead of forcing to override `onCreate` method like the one below?

```kotlin
    init {
        addOnContextAvailableListener {
            initializeVideoIfNeeded()
            processNotificationData()
        }
    }
```

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+Thanks to our [Stream Push](https://github.com/GetStream/stream-android-push) artifacts, it's very easy to integrate push notification handling within your Chat and Video apps.
+
+Additionally, for Video, we support notifications when engaging in **ringing calls**, where you can accept or decline calls with just a click of a button. This behavior is included in the `stream-video-android` dependency, so you don't have to add any other artifacts to your project.
+
+There are only a few smaller steps to integrate push notifications with Stream Video. Let's see how.
+
+### Adding The NotificationActivity
+
+The first step to integrating Push support is providing your app and our SDK with an `Activity` that handles push notification payload.
```
JcMinarro: There is a previous step needed for PN.
They need to add, at least, one `PushDeviceGenerator` when they build their `StreamVideo` (https://github.com/GetStream/stream-video-android/blob/develop/stream-video-android/src/main/kotlin/io/getstream/video/android/StreamVideoBuilder.kt#L45)

It will register the device on Stream Server

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+We provide an `AbstractNotificationActivity` that implements most of the logic for you. You only need to provide it with an instance of `StreamVideo` and a way to intitialize the SDK in your app, if it hasn't been initialized.
```
JcMinarro, line 41: Aren't we going to provide a "full implementation" that creates all needed instances out of the box?
I think with the improvements we did over `AuthCredentialsProvider` class, we should be able to create and initialize StreamVideo by ourselves, and provide a default implementation to our customers. They can always provide their own implementation that will override our default one

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
filbabic: They shouldn't have to override it since the super will be called anyway, right?

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
filbabic: Good point, I'll add this to the guide!

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
filbabic: We cannot provide a full implementation as stream video instance is not a singleton and sharing the state across multiple instances is not possible.

This is blocked by our stream engine and the states that are handled under the hood.

Since they have to add an activity to the manifest anyway, it's very easy to just extend the AbstractNotificationActivity and have it handle everything

JcMinarro: I know we don't have a singleton instance, but, Why can't we create a `StreamVideo` with the `AuthCredentialsProvider`?
It should be enough to accept the incoming call, right?
Maybe we can't add all customization it could have (I mean, customers could add more stuff to the `StreamVideo` instance) but at least it works out of the box.

We can keep the `AbstractNotificationActivity` to facility our customer their own implementation

filbabic: Because internally the StreamCallEngine keeps the state of the call and dictates what's going to happen in which part of state - so if we create the instance ourselves, instead of, e.g. fetching it from the customer's App class, we will potentially have multiple instances of StreamVideo and StreamCallEngine and in that case the state would be misaligned.

An alternative is to rewrite the engine to be a singleton, but I feel like this is simple enough given that all of our other activities and ease-of-use features (like the CallService) follow the same pattern.

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
```diff
@@ -0,0 +1,81 @@
...
+Our Push Notification system automatically detects and starts the `Activity` that declares it can consume the notification actions. For this to work, you have to declare the `NotificationActivity` in your manifest and add the appropriate intent filters used to find it  
+
+Add the following code to the `AndroidManifest.xml`:
+
+```xml
+<activity
+    android:name=".NotificationActivity"
+    android:exported="false">
+
+    <intent-filter android:priority="-1">
```
JcMinarro: Negative priority should only be used by our SDK, allowing our customers to override it with a higher priority (If they omit it, the priority will be 0, that is bigger than any negative value)

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
skydoves: Yes, but just in case, if they override the `onCreate` function and don't call the `super` function. Or we can add the [CallSuper](https://developer.android.com/reference/kotlin/androidx/annotation/CallSuper) annotation to the `onCreate` function.

filbabic: From what Kanat pointed out - it'll crash if super isn't called, so in that sense we might not have to! But I can add a note to the text!

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
filbabic: Resolving for now - we can look into this later!

docusaurus/docs/Android/05-advanced/02-push-notifications.mdx
filbabic: I'll change the guide, thanks!

PR #187, opened 2023-01-26
Improve API around call controls and our other components

filbabic:
General API improvements around the CallControls and some of the customization API.

To test if this works or not, the easiest way is to extend the `AbstractComposeCallActivity` and then overriding the content function and its CallControls slot.

I don't have specific repro steps as most of this is tied to the Video + Chat guide, so as soon as that's up and running, the guide will cover all the necessary steps.

Alternatively, you can test that nothing changed with these changes (in default):
- Run app/dogfooding and join a call with > 1 person (suggested with one web client)
- Test call controls and the rest of the UI works fine in both landscape and portrait
- Start screen sharing
- Test call controls and rest of the UI works fine, when screen sharing, in both landscape and portrait

There shouldn't be any changes or issues with the default impl.

### DIFF
--- a/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/controls/RegularCallControls.kt
+++ b/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/controls/RegularCallControls.kt
@@ -85,7 +85,7 @@ public fun RegularCallControlsActions(
         horizontalArrangement = Arrangement.SpaceEvenly
     ) {
         items(actions) { action ->
-            val isEnabled = (action.callAction is FlipCamera && isScreenSharing).not()
+            val isEnabled = !(action.callAction is FlipCamera && isScreenSharing)
 
             Card(
                 modifier = Modifier.size(VideoTheme.dimens.callControlButtonSize),


stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/controls/RegularCallControls.kt
```diff
@@ -0,0 +1,106 @@
...
+    actions: List<CallControlAction>,
+    isScreenSharing: Boolean,
+    onCallAction: (CallAction) -> Unit
+) {
+    LazyRow(
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalArrangement = Arrangement.SpaceEvenly
+    ) {
+        items(actions) { action ->
+            val isEnabled = (action.callAction is FlipCamera && isScreenSharing).not()
```
TGazica: Should we use `.not()` moving forward or keep using `!`? We can adopt the codebase as we move along.

TGazica:
```suggestion
            val isEnabled = !(action.callAction is FlipCamera && isScreenSharing)
```

If you wish to apply it like this.

PR #205, opened 2023-02-16
Remove complex auth providers and managers, simplify process

### LINKED ISSUE #193: Improve the way credentials are stored
### Context & proposal

One issue we can have currently with the credentials implementation, is that they're exposed separately from the StreamVideo instance. Because of that, we can come into a situation where the user is changed and the StreamVideo instance is persisted, causing the wrong token/user to be stored within.

One thing we could do is store the CredentialsProvider within the StreamVideo and expose it through there. Then the user, whenever they create a StreamVideo instance, can just fetch it from SV.

### Acceptance criteria

To accept this task we need to complete the following items:

- [x] Expose CredentialsProvider or the User through StreamVideo
- [x] Remove the need to persist the CredentialsProvider in App classes
- [x] (Optional) Look into removing the CredentialsProvider and instead using only UserPreferences (singleton) for any credentials.

filbabic:
### 🎯 Goal

It was a bit complex to set up credentials and think about them. They gave way for errors in initializing the video client, so I figured we'd better simplify it for the initial release to make the process much easier for customers.

Closes #193 

### 🛠 Implementation details

Removed AuthCredentialsProvider, CredentialsProvider, CredentialsManager and similar. Only relying on Preferences right now. 

### 🧪 Testing

To test, make sure to run the Dogfooding, App and Chat + Video final apps. None of these should be crashing or broken, things should work like before.

#### General
- [x] PR is linked to the GitHub issue it resolves

#### Code & documentation

TODO - will update the docs probably at a later stage. Need to thoroughly test this to make sure it's clean and working properly.

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/HttpModule.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/HttpModule.kt
@@ -24,18 +24,18 @@ import okhttp3.logging.HttpLoggingInterceptor
 
 /**
  * @property loggingLevel Log level used for all HTTP requests towards the API.
- * @property credentialsProvider Provider used to fetch user based credentials.
+ * @property preferences Used to fetch user based credentials.
  */
 internal class HttpModule(
     private val loggingLevel: HttpLoggingInterceptor.Level = HttpLoggingInterceptor.Level.NONE,
-    private val credentialsProvider: UserPreferences,
+    private val preferences: UserPreferences,
 ) {
 
     /**
      * Cached instance of the HTTP client.
      */
     internal val okHttpClient: OkHttpClient by lazy {
-        buildOkHttpClient(credentialsProvider)
+        buildOkHttpClient(preferences)
     }
 
     /**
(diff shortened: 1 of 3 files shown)


examples/chat-with-video/chat-with-video-final/src/main/java/io/getstream/video/chat_with_video_final/application/ChatWithVideoApp.kt
```diff
@@ -90,45 +89,35 @@ class ChatWithVideoApp : Application() {
...
+    private var video: StreamVideo? = null
 
-    lateinit var streamVideo: StreamVideo
-        private set
+    val streamVideo: StreamVideo
+        get() = requireNotNull(video)
 
     fun initializeStreamVideo(
-        credentialsProvider: CredentialsProvider,
+        user: io.getstream.video.android.core.model.User,
```
TGazica, line 98: add import

filbabic: We can't do it because there are two User objects (Chat + Video). We could import alias as VideoUser or something but it could be a bit confusing

PR #215, opened 2023-02-28
Update to latest coordinator, added missing endpoints to CallCoordinatorClient as well as missing events

filbabic:
Updated the coord, added some events, parsed them, added API calls to the client (haven't built a special ViewModel or CallViewModel handles)

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/coordinator/CallCoordinatorClient.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/coordinator/CallCoordinatorClient.kt
@@ -215,6 +215,7 @@ internal interface CallCoordinatorClient {
      *
      * @param id Call ID.
      * @param type Call type.
+     * @param muteUsersRequest The request containing information about muting users.
      */
     suspend fun muteUsers(
         id: String,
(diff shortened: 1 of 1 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/coordinator/CallCoordinatorClient.kt
```diff
@@ -137,4 +146,172 @@ internal interface CallCoordinatorClient {
...
+    /**
+     * Attempts to mute users and their tracks in a call.
+     *
+     * @param id Call ID.
+     * @param type Call type.
+     */
+    suspend fun muteUsers(
+        id: String,
+        type: String,
+        muteUsersRequest: MuteUsersRequest
```
TGazica, line 223: docs

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/coordinator/CallCoordinatorClient.kt
```diff
@@ -137,4 +146,172 @@ internal interface CallCoordinatorClient {
...
+     * @param updateCallRequest The request to update the call.
+     */
+    suspend fun updateCall(
+        id: String,
+        type: String,
+        updateCallRequest: UpdateCallRequest
+    ): Result<CallInfo>
+
+    /**
+     * Queries calls with a given filter predicate and pagination.
```
TGazica, line 240: Are filter predicates and pagination inside `queryCallsRequest`? If so ti would be good to make that more clear.

PR #238, opened 2023-03-23
Improve some APIs, Compose UI components, and dogfooding app

### LINKED ISSUE #236: Improve & polish dogfooding app
We should entirely improve & polish dogfooding app.

skydoves:
Improve some APIs, Compose UI components, and dogfooding app.

I've fixed some crash issues on dogfooding app, handled 401 (unauthorized error), fix UI components for Jetpack Compose.

Close #236

Screenshot 2023-03-23 at 3 40 23 PM

Screenshot 2023-03-23 at 3 40 34 PM

Screenshot 2023-03-23 at 3 40 42 PM

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideo.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideo.kt
@@ -43,7 +43,7 @@ import io.getstream.video.android.core.utils.Result
 import kotlinx.coroutines.flow.StateFlow
 
 /**
- * The main interface to control the Video calls. [StreamVideoImpl] is implements this interface.
+ * The main interface to control the Video calls. [StreamVideoImpl] implements this interface.
  */
 public interface StreamVideo {
 


dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
```diff
@@ -18,11 +18,45 @@ package io.getstream.video.android.dogfooding
...
 import io.getstream.video.android.compose.ui.AbstractComposeCallActivity
 import io.getstream.video.android.core.StreamVideo
+import io.getstream.video.android.core.model.state.StreamCallState
+import kotlinx.coroutines.launch
 
 class CallActivity : AbstractComposeCallActivity() {
 
+    private val mediaPlayer by lazy {
+        MediaPlayer.create(this, R.raw.ringing).apply { isLooping = true }
+    }
```
TGazica, line 35: Any reason why we play a ringing sound while joining a meet? This should be google meets like experience.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideo.kt
```diff
@@ -42,6 +42,9 @@ import io.getstream.video.android.core.socket.SocketListener
 import io.getstream.video.android.core.utils.Result
 import kotlinx.coroutines.flow.StateFlow
 
+/**
+ * The main interface to control the Video calls. [StreamVideoImpl] is implements this interface.
```
TGazica:
```suggestion
 * The main interface to control the Video calls. [StreamVideoImpl] implements this interface.
```

dogfooding/src/main/java/io/getstream/video/android/dogfooding/CallActivity.kt
skydoves: Nothing special; I wanted to implement some feeling like the application is working well. (joining-connecting-connected flow). We all know the flows so we think it's working well without any indications, but first-access users might be confused if the app is working well or not while joining a call.

TGazica: Could be a nice little feature.

PR #249, opened 2023-03-28
Update models and protos

TGazica:
Updates models, protos and api calls


stream-video-android-core/src/main/kotlin/org/openapitools/client/models/CallTypeResponse.kt
```diff
@@ -0,0 +1,49 @@
...
+)
+
+package org.openapitools.client.models
+
+import com.squareup.moshi.Json
+
+/**
+ * *
+ * @param createdAt * @param grants * @param name * @param settings * @param updatedAt */
+
```
skydoves: Comment formats look broken here

stream-video-android-core/src/main/kotlin/org/openapitools/client/models/ConnectUserDetailsRequest.kt
```diff
@@ -0,0 +1,46 @@
...
+)
+
+package org.openapitools.client.models
+
+import com.squareup.moshi.Json
+
+/**
+ * *
+ * @param id * @param custom * @param image * @param name */
+
```
skydoves: Same

PR #367, opened 2023-05-10
Dynascale 0.1

tschellenbach:
If the UI renders a participant, it should

* Mark the participant as visible
* Update the resolution that's shown as the display change

The updateSubscriptions call is called whenever

* visibility changes
* resolution changes
* you join a call, or someone else joins

The call is debounced and only runs if the tracks you're subscribing to changed. (This prevents lots of API calls for a livestream or something where people join but it doesn't affect your track subscriptions)

We should handle the updatePublish event, currently, we only support turning/on/off any of the 3 layers.
So you can decide to only publish the q and not the f. We don't support changing the bitrate or codec, but afaik we currently don't use that in the SFU

TODO 1:

Improvement/ Future changes

* We need the VideoRenderer component to
* A. Mark the video as visible. This triggers the track subscription
* B. While the track isn't ready yet we should show the fallback
* Right now that doesn't work 100% right. We don't show anything while the track is loading

The proposed solution for the VideoRenderer compose component:

* Pass a fallback component to the VideoRender (that people can customize)
* Mark track visible immediately (we already do this)
* Show the fallback while the track isn't ready yet

What do you think @skydoves ?

TODO 2

Investigate this: https://github.com/GetStream/stream-video-android/pull/350/commits/e6eca766e735519d225a49d419cc539f8ecf1366#diff-446db17ffd4e9feed351715c6efde07660d9406b7da69ec09cfd16c8ffe170d2R350


stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/VideoRenderer.kt
```diff
@@ -75,32 +75,39 @@ public fun VideoRenderer(
 
     var view: VideoTextureViewRenderer? by remember { mutableStateOf(null) }
 
+    // inform the call that we want to render this video track. (this will trigger a subscription to the track)
+    call.setVisibility(sessionId, trackType, true)
```
skydoves, line 79: We can move this inside the `DisposableEffect` if we don't need to inform whenever a recomposition arises.

tschellenbach: just has to happen once when we show the video, and once when remove the video from the UI

skydoves: Sounds clear! Then let's move it inside the `DisposableEffect` like the example below:

```kotlin
    DisposableEffect(call, mediaTrack) {
        // inform the call that we want to render this video track. (this will trigger a subscription to the track)
        call.setVisibility(sessionId, trackType, true)

        onDispose {
            cleanTrack(view, mediaTrack)
            // inform the call that we no longer want to render this video track
            call.setVisibility(sessionId, trackType, false)
        }
    }
```

PR #565, opened 2023-06-27
AVS-4 Fix: screensharing doesn't stops if participant disconnects

DanielNovak:
Fixes a bug where the screensharing doesn't stop if the participant abruptly leaves the call.

**How to reproduce:**
1) Make a call between Android and browser
2) Share a screen from browser call
3) Close the browser tab / close the browser
**Issue**: Android app will still be showing the screen share.


tschellenbach:
awesome, first pull request merged

PR #580, opened 2023-06-28
[Docs] Use shared call types page

DaemonLoki:
### 🎯 Goal

Use the shared Call types page.

### 🛠 Implementation details

It takes the shared page that we have and loads it into the Flutter docs. The content is not fully there, but once it is and the shared page has the content, it will automatically be distributed across SDKs.

It still requires a bit of a better integration into the existing content, I think!

Want to run it locally? Check [this page](https://www.notion.so/stream-wiki/Documentation-knowledge-d37b53d3ce6e42bcaaa0cabf3bc125c5) for details.

### 🎨 UI Changes

Call Types | Stream Video - Android SDK Docs

### DIFF
--- a/docusaurus/docs/Android/03-guides/05-call-types.mdx
+++ b/docusaurus/docs/Android/03-guides/05-call-types.mdx
@@ -3,9 +3,12 @@ title: Call Types
 description: How Call Types control features and permissions
 ---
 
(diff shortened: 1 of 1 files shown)


skydoves (approved):
Looks good to me!

PR #590, opened 2023-06-29
AVS-13 Add speaking-while-muted detection into the SDK

DanielNovak:
This change makes the `CallState.speakingWhileMuted` StateFlow in the SDK functional. 
UI can listen to this `StateFlow` and display a warning to the user if he is speaking while the microphone is muted. 
**Note**: The flag will reset itself automatically after 2 seconds.

We intercept the `AudioSamples` callback from the `JavaAudioDeviceModule` in WebRTC and then do basic average decibel level analysis. The flag is set if the average decibel value surpasses a certain level. The analysis is only done while in call and only when the microphone is muted.

A separate PR should be create later to connect the application UI with the flag and display the actual warning.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/DecibelThresholdDetection.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/DecibelThresholdDetection.kt
@@ -16,7 +16,6 @@
 
 package io.getstream.video.android.core.call.utils
 
-import androidx.annotation.WorkerThread
 import kotlin.math.log10
 import kotlin.math.pow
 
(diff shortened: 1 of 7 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/DecibelThresholdDetection.kt
```diff
@@ -0,0 +1,114 @@
...
+    val thresholdCrossedCallback: () -> Unit
+) {
+
+    private val decibelSamples = LinkedHashMap<Long, Double>()
+
+    /**
+     * Forward the input audio data from the microphone to this function.
+     * Only 16bit PCM audio format is accepted.
+     */
+    @WorkerThread
```
skydoves: Do we need to add this annotation? I think we can just remove this or add several other methods that depend on `processSoundInput`, because this annotation already lost the meaning in the `processAudioSample`.

skydoves (approved):
Overall looks good to me. I will implement some UI components after merging this PR. Well done! 👍

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/DecibelThresholdDetection.kt
DanielNovak: We don't probably need it - I will remove it.

PR #594, opened 2023-06-29
AVS-17 Use Twilio AudioSwitch as dependency

DanielNovak:
Replace our "fork" of [AudioSwitch library](https://github.com/twilio/audioswitch) with a direct dependency. This makes it also easier to maintain and keep updated. 

At some point we may need to make a fork anyway if we want to add logging into AudioSwitch. The built-in logger is not customisable and uses the `Log` class only.

**TODO**: Test with real bluetooth headsets before merging.


skydoves (approved):
Looks good to me! Nice contribution 👍

Issue #625, opened 2023-07-06
XML version of VideoRenderer 


skydoves:
XML version of VideoRenderer

PR #629, opened 2023-07-10
AVS-25 Implement VideoRendererView for supporting VideoRenderer of XML

skydoves:
Implement VideoRendererView for supporting VideoRenderer of XML.


skydoves:
#625

Issue #635, opened 2023-07-12
Popup for reactions


skydoves:
Popup for reactions.

Issue #634, opened 2023-07-12
Popup for Ice restart & switch sfu


skydoves:
Popup for Ice restart & switch sfu.

PR #636, opened 2023-07-12
AVS-29 Implement a popup to restart Ice, switch sfu, and send reactions

skydoves:
AVS-29 Implement a popup to restart Ice, switch sfu, and send reactions.


skydoves:
#634 #635

tschellenbach:
image

PR #642, opened 2023-07-14
AVS-35 Fix participant count before call is joined

DanielNovak:
This adds handling of new Session events (`CallSessionStartedEvent`, `CallSessionEndedEvent`, `CallSessionParticipantLeftEvent`, `CallSessionParticipantJoinedEvent`). The count of participants can now be observed before the call is joined, as long as you execute `call.get()` at least once. Use `call.state.session.participants` to get the list (which is different than `call.state.participants`).


tschellenbach:
great

PR #643, opened 2023-07-14
AVS-36 Fix screensharing not visible

DanielNovak:
Fixes the screenshare not being displayed.

There are still some resolution related issues. Stopping screensharing will sometimes draw the participant video in wrong resolution (rectangle and top-aligned) and also sharing a larger resolution window with Firefox will just draw a gray screen on Android. We will need to solve these issues in a separate PR.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
@@ -849,29 +849,15 @@ public class RtcSession internal constructor(
         val tracks = participants.map { participant ->
             val trackDisplay = trackDisplayResolution[participant.sessionId] ?: emptyMap()
 
-            val tracks = mutableListOf<TrackSubscriptionDetails>()
-            val entries = trackDisplay.entries.filter { it.value.visible }
-            entries.forEach { entry ->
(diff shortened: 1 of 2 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -849,15 +849,29 @@ public class RtcSession internal constructor(
...
+                }
+
+                if (participant.screenSharingEnabled.value) {
+                    tracks.add(TrackSubscriptionDetails(
+                        user_id = participant.user.value.id,
+                        track_type = TrackType.TRACK_TYPE_SCREEN_SHARE,
+                        dimension = entry.value.dimensions,
+                        session_id = participant.sessionId
+                    ))
+                }
```
tschellenbach: if the video is visible, but the screenshare isn't visible, this will make us subscribe to the screenshare won't it? 

also won't it add it to the tracks twice?

DanielNovak: AFAIK the tracks need to be added twice - in the bottom part of the screen you can still see the participant + at the top you see the screen share. So up to two tracks per participant.

PR #647, opened 2023-07-17
AVS-37 Handle pagination of participants if count > 6

DanielNovak:
Use `LazyVerticalGrid` if participant size > 6. I left the custom participant "grid" code for participant size up to 6 - this will give us more control over the layout and the LazyVerticalGrid is also more complex to render, so we can defer it until we have 7 and more participants.

### DIFF
--- a/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/LandscapeVideoRenderer.kt
+++ b/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/LandscapeVideoRenderer.kt
@@ -305,7 +305,10 @@ internal fun BoxScope.LandscapeVideoRenderer(
                     modifier = Modifier.fillMaxSize(),
                     columns = GridCells.Fixed(3),
                     content = {
-                        items(callParticipants.size) { key ->
+                        items(
+                            count = callParticipants.size,
+                            key = { callParticipants[it].sessionId }
+                        ) { key ->
                             // make 2 items exactly fit available height
                             val itemHeight = with(LocalDensity.current) {
                                 (constraints.maxHeight / 2).toDp()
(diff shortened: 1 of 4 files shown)


stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/LandscapeVideoRenderer.kt
```diff
@@ -293,6 +297,33 @@ internal fun BoxScope.LandscapeVideoRenderer(
...
+                            }
+                            val participant = callParticipants[key]
+                            videoRenderer.invoke(
+                                modifier = modifier.height(itemHeight),
+                                call = call,
+                                participant = participant,
+                                style = style.copy(
+                                    isFocused = dominantSpeaker?.sessionId == participant.sessionId
+                                )
+                            )
```
skydoves: We can improve here by using items (list + key) like the example below:

```kotlin
items(
    items = callParticipants,
    key = { it.sessionId }
) { participant ->
    // make 2 items exactly fit available height
    val itemHeight = with(LocalDensity.current) {
        (constraints.maxHeight / 2).toDp()
    }
    videoRenderer.invoke(
        modifier = modifier.height(itemHeight),
        call = call,
        participant = participant,
        style = style.copy(
            isFocused = dominantSpeaker?.sessionId == participant.sessionId
        )
    )
}
```

Providing a stable key enables item state to be consistent across data-set changes.

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/PortraitVideoRenderer.kt
```diff
@@ -311,6 +315,33 @@ internal fun BoxScope.PortraitVideoRenderer(
...
+
+                LazyVerticalGrid(
+                    modifier = Modifier.fillMaxSize(),
+                    columns = GridCells.Fixed(2),
+                    content = {
+                        items(callParticipants.size) { key ->
+                            // make 3 items exactly fit available height
+                            val itemHeight = with(LocalDensity.current) {
+                                (constraints.maxHeight / 3).toDp()
+                            }
```
skydoves: Same here.

skydoves (changes requested):
Looks good to me. I left a small suggestion.

PR #655, opened 2023-07-18
AVS-39 Socket implementation will crash if JSON format of event doesn't match

DanielNovak:
Application would crash if the VideoEvent format didn't match or the error message was missing.


skydoves (approved):
Looks good to me! 👍

PR #661, opened 2023-07-19
Temporary fix for notification format validation (+ PiP fixes)

DanielNovak:
The PR fixes three issues and one of them is a temporary fix to unblock further work:

- `IncomingCallActivity` was missing `supportsPictureInPicture=true` (causing a crash if home button was clicked on a call that was joined from notification)
- `IncomingCallActivity` would crash if re-created for example during PiP mode, `call.join` could be called repeatedly
- The notification format checks in `VideoDelegate` are not valid anymore. The `call_display_name` in the push bundle is empty. Other platforms are not using this field - they use `created_by_display_name` instead. We need to clarify this inconsistency and make a proper fix later, but for now this will unblock us from further testing.

### DIFF
--- a/stream-video-android-core/api/stream-video-android-core.api
+++ b/stream-video-android-core/api/stream-video-android-core.api
@@ -1,6 +1,5 @@
 public final class io/getstream/video/android/BuildConfig {
 	public static final field BUILD_TYPE Ljava/lang/String;
-	public static final field CORE_VIDEO_DOMAIN Ljava/lang/String;
 	public static final field DEBUG Z
 	public static final field LIBRARY_PACKAGE_NAME Ljava/lang/String;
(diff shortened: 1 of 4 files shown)


stream-video-android-core/api/stream-video-android-core.api
```diff
@@ -1,6 +1,5 @@
 public final class io/getstream/video/android/BuildConfig {
 	public static final field BUILD_TYPE Ljava/lang/String;
-	public static final field CORE_VIDEO_DOMAIN Ljava/lang/String;
```
DanielNovak, line 3: This is an unrelated change but this entry reappeared here in some previous commit by accident. So I will use the opportunity to also remove it (the CI API check would complain otherwise).

PR #662, opened 2023-07-19
Update to latest openapi

DanielNovak:
Update to latest openapi.


stream-video-android-core/src/main/kotlin/org/openapitools/client/models/CallSessionParticipantJoinedEvent.kt
```diff
@@ -56,20 +55,16 @@ data class CallSessionParticipantJoinedEvent (
...
     /* Call session ID */
     @Json(name = "session_id")
     val sessionId: kotlin.String,
 
     /* The type of event: \"call.session_participant_joined\" in this case */
     @Json(name = "type")
-    val type: kotlin.String = "call.session_participant_joined",
-
-    @Json(name = "user")
-    val user: UserResponse,
```
DanielNovak, line 68: (breaking change here, `user` and `user_session_id` were removed)

PR #672, opened 2023-07-20
Init StreamVideo properly into VideoPushDelegate

JcMinarro:
### 🎯 Goal
Our Push Notification Implementation needs a StreamVideo instance to work. We should reuse the one our customers already created, but could be the case our customers didn't initialized it properly.
If that is the case, we should try to create one from the persisted data, and after all our process is completed, "uninstall" this instance to ensure we don't override any customer config


DanielNovak (approved):

skydoves (approved):
Looks good to me!

PR #678, opened 2023-07-21
AVS-58 Incoming call intents have a wrong action

DanielNovak:
The incoming call intents were not created correctly. All PendingIntents had the same "key" which was determined by the last Intent in `PendingIntent.getActivities` and in our case this created always the same `Intent(context, DismissNotificationActivity::class.java)`. The system would then reuse existing PendingIntents and deliver an intent with the wrong action. For example instead of getting `INCOMING_CALL` we would get `ACCEPT_CALL`.

For more information look at the docs of `PendingIntent.getActivities()`. Especially the part about the last intent being the key.

How to reproduce the bug:
1. Start dogfooding application
2. Call application from other device (e.g. iOS)
3. Click on the notification body (not Answer or Reject). And check in `IncomingCallActivity` that the `intent.action` is incorrect. It's `ACCEPT_CALL` instead of `INCOMING_CALL`.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/DismissNotificationActivity.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/DismissNotificationActivity.kt
@@ -32,10 +32,20 @@ internal class DismissNotificationActivity : Activity() {
 
     companion object {
         private const val KEY_NOTIFICATION_ID = "KEY_NOTIFICATION_ID"
-        fun createIntent(context: Context, notificationId: Int): Intent =
+
+        /**
+         * @param baseIntentAction - Important - the action must be set. We are creating multiple
+         * dismiss intents (accept, reject, ...) and the system will "think" the Pending intents
+         * we create in [DefaultNotificationHandler.getActivityForIntent] are identical. This can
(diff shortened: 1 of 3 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/DismissNotificationActivity.kt
```diff
@@ -32,10 +32,21 @@ internal class DismissNotificationActivity : Activity() {
...
+         * we create in [DefaultNotificationHandler.getActivityForIntent] are identical. This can
+         * lead to a reused intent with a wrong action.
+         * See documentation of [PendingIntent.getActivities] - the last intent in the list is the key.
+         *
+         */
+        fun createIntent(context: Context, notificationId: Int, baseIntentAction: String): Intent =
             Intent(context, DismissNotificationActivity::class.java)
                 .apply {
                     putExtra(KEY_NOTIFICATION_ID, notificationId)
+                }.apply {
```
JcMinarro:
```suggestion
                .apply {
                    putExtra(KEY_NOTIFICATION_ID, notificationId)
```

JcMinarro (approved):
LGTM

Issue #680, opened 2023-07-23
Clean up unused notification managers and abstract services


skydoves:
Clean up unused notification managers and abstract services.

PR #681, opened 2023-07-23
AVS-59 Remove unused notification managers and abstract services

skydoves:
AVS-59 Remove unused notification managers and abstract services.


skydoves:
Close #680

PR #683, opened 2023-07-24
Fix Flow/StateFlow issues in StreamUserDataStore

DanielNovak:
The initial value of the our `StreamUserDataStore` was always null - we can't use `asStateFlow` - the DataStore will not automatically just emit a value, you need to `collect()` it first. So instead of using a `StateFlow` I just used the original `Flow` and mapped it.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
```diff
@@ -17,6 +17,7 @@
 package io.getstream.video.android.core.notifications.internal
 
 import android.content.Context
+import android.util.Log
```
JcMinarro: Looks like this import is unused

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
```diff
@@ -102,7 +104,7 @@ internal class VideoPushDelegate(
                     context = context,
                     user = user,
                     token = userToken,
-                    apiKey = userDataStore.apiKey.value,
+                    apiKey = userDataStore.apiKey.first(),
```
JcMinarro, line 106: Are we sure it won't be null?
Shouldn't we follow the same approach as previous values?

stream-video-android-datastore/src/main/kotlin/io/getstream/video/android/datastore/delegate/StreamUserDataStore.kt
```diff
@@ -114,20 +115,19 @@ public class StreamUserDataStore constructor(
     public suspend fun clear(): StreamUserPreferences? = updateData { null }
 
     /** A state that contains a persisted [User] data. */
-    public val user: StateFlow<User?> = data.map { it?.user }.asStateFlow(null, scope)
```
JcMinarro, line 117: After removing the `StateFlow` from this class, the `scope` instance is not used at all

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -375,9 +378,9 @@ public class CallState(private val call: Call, private val user: User, internal
     val createdBy: StateFlow<User?> = _createdBy
 
     private val _ingress: MutableStateFlow<CallIngressResponse?> = MutableStateFlow(null)
-    val ingress: StateFlow<Ingress?> = _ingress.mapState {
-        val token = call.clientImpl.dataStore.userToken.value
-        val apiKey = call.clientImpl.dataStore.apiKey.value
+    val ingress: StateFlow<Ingress?> = _ingress.mapState(initialValue = null, scope = CoroutineScope(Dispatchers.IO)) {
```
JcMinarro: A `scope` is injected on the constructor, we should create a new one into the class, it will difficult tests in the future

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -277,12 +278,15 @@ public class RtcSession internal constructor(
 
     init {
         val dataStore = StreamUserDataStore.instance()
-        val user = dataStore.user.value
-        val apiKey = dataStore.apiKey.value
-        if (apiKey.isBlank() || user?.id.isNullOrBlank()) {
-            throw IllegalArgumentException("The API key, user ID and token cannot be empty!")
+        coroutineScope.launch {
+            val user = dataStore.user.firstOrNull()
+            val apiKey = dataStore.apiKey.first()
```
JcMinarro, line 283: `firstOrNull()` instead?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/RejectCallBroadcastReceiver.kt
```diff
@@ -54,24 +57,27 @@ internal class RejectCallBroadcastReceiver : BroadcastReceiver() {
...
+                val streamVideo: StreamVideo = if (StreamVideo.isInstalled) {
+                    StreamVideo.instance()
+                } else {
+                    userDataStore.user.firstOrNull()?.let { user ->
+                        userDataStore.userToken.firstOrNull()?.let { userToken ->
+                            StreamVideoBuilder(
+                                context = context,
+                                user = user,
+                                token = userToken,
+                                apiKey = userDataStore.apiKey.first(),
```
JcMinarro, line 74: Same here

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
DanielNovak: This is never null and the function doesn't take null. 

```
    public val apiKey: Flow<ApiKey> =
        data.map { it?.apiKey.orEmpty() }
```
It creates an empty API key - I haven't changed this logic. That's why I have to use `first()` here so that it compiles and expects a non-null value.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/RejectCallBroadcastReceiver.kt
DanielNovak: See above - API key is non-null always.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
DanielNovak: See above - API key is non-null always.

stream-video-android-datastore/src/main/kotlin/io/getstream/video/android/datastore/delegate/StreamUserDataStore.kt
DanielNovak: Good catch.

PR #685, opened 2023-07-24
WIP: Implement RingingState and a single CallContentScreen for call state handling

DanielNovak:
**Work In Progress - do not merge**


stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/ringing/outgoingcall/OutgoingCallContent.kt
```diff
@@ -58,19 +58,19 @@ import io.getstream.video.android.mock.mockParticipantList
 public fun OutgoingCallContent(
     modifier: Modifier = Modifier,
     call: Call,
-    isVideoType: Boolean,
+    isVideoType: Boolean = true,
```
skydoves: I think giving this value automatically here is not a good idea - because Incoming/Outgoing calls have different UI/UX (You can check the out in our internal design guidelines) depending on this value.

dogfooding/src/main/kotlin/io/getstream/video/android/ui/call/CallActivity.kt
```diff
@@ -68,36 +58,33 @@ class CallActivity : ComponentActivity() {
...
         setContent {
-            CallScreen(
+            CallContentScreen(
                 call = call,
-                showDebugOptions = io.getstream.video.android.BuildConfig.DEBUG,
                 onLeaveCall = {
                     call.leave()
                     finish()
-                }
-            )
```
skydoves: I'm not sure if this is a good approach because we should cover more use cases for joining a specific call (you can imagine Disco, Clubhouse, Twitter Live, or something) instead of ringing calls. We need to discuss more about this on the channel. So even if we may need to have some duplicated codes for Control Actions or something, I'd prefer separating them and making their dedicated role clear.

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/lifecycle/MediaPiPLifecycle.kt
```diff
@@ -66,22 +66,7 @@ public fun MediaPiPLifecycle(
...
-                // TODO: There's not way to onUserLeaveHint in Compose for now.
-                // https://developer.android.com/reference/android/app/Activity#onUserLeaveHint()
-                try {
-                    Handler(Looper.getMainLooper()).post {
-                        enterPictureInPicture(context = context, call = call)
-                    }
-                } catch (e: Exception) {
-                    StreamLog.d("MediaPiPLifecycle") { e.stackTraceToString() }
-                }
-            }
```
skydoves: Is there any reason to remove the whole PIP support here? 🤔

skydoves (changes requested):
I left several comments, which seems to we need more discussion here, but overall approach looks a great trial 😄

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/ringing/outgoingcall/OutgoingCallContent.kt
DanielNovak: True. I removed the "common call UI" commit from this PR in the end.

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/lifecycle/MediaPiPLifecycle.kt
DanielNovak: Just by accident when testing background calls (PiP makes the app always foreground).

dogfooding/src/main/kotlin/io/getstream/video/android/ui/outgoing/DebugCallScreen.kt
```diff
@@ -0,0 +1,144 @@
...
+                    text = "Ring"
+                )
+            }
+        }
+    }
+}
+
+@Preview
+@Composable
+private fun CallLobbyScreenPreview() {
```
skydoves: `DebugCallScreenPreview`? 😄

skydoves (approved):
This is a lot of great works. Nicely done! 🙌

dogfooding/src/main/kotlin/io/getstream/video/android/ui/outgoing/DebugCallScreen.kt
DanielNovak: Fixed :-)

PR #686, opened 2023-07-24
Prevent flicker in CallJoinScreen (frequent loading state emit)

DanielNovak:
Small UI fix - we shoudln't be emitting the Loading state every time a value from the datastore is delivered. We need to only emit it once before we start loading. This prevents UI flicker.


skydoves (approved):
This is a great catch, Daniel! 👍

PR #694, opened 2023-07-26
Upload unit test results in build artifacts

DanielNovak:
We are missing the unit test reports in the CI build runs. It's hard to know what exactly failed without the report - especially if it's only failing on CI.
Source article for the


skydoves (approved):
Looks good to me!

PR #697, opened 2023-07-27
[WIP] Livestream tutorials

skydoves:
[WIP] Livestream tutorials.


docusaurus/docs/Android/05-ui-cookbook/15-watching-livestream.mdx
```diff
@@ -0,0 +1,360 @@
+---
+title: Watching a livestream
+description: How to watch a livestream on android with Kotlin
+---
+
+This cookbook tutorial walks you through how to render the UI for watching a livestream on Android.
+
+:::note
+This cookbook tutorial will skip how to join a livestream call. If you didn't learn about the ![Livestream Tutorial](../02-tutorials/03-livestream.mdx), we're highly recommend you to read about the tutorial first.
```
DanielNovak: `we're highly recommend` should be `we highly recommend`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -306,6 +316,22 @@ public class CallState(
...
+            val started = _session.value?.startedAt
+            val ended = _session.value?.endedAt ?: OffsetDateTime.now()
+            val difference = if (started == null) null else {
+                ended.toInstant().toEpochMilli() - started.toInstant()?.toEpochMilli()!!
+            }
+            emit(difference)
+        }
+    }
+
+    /** how many MS the call has been running, null if the call didn't start yet */
```
DanielNovak: Instead of `MS` I would either use `ms` or maybe it will be more clear with `milliseconds`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -333,9 +359,22 @@ public class CallState(
 
     /** if we are in backstage mode or not */
     val backstage: StateFlow<Boolean> = _backstage
+
     /** the opposite of backstage, if we are live or not */
     val live: StateFlow<Boolean> = _backstage.mapState { !it }
 
+    /** how many MS the call has been running, null if the call didn't start yet */
```
DanielNovak: Same as above: Instead of `MS` I would either use `ms` or maybe it will be more clear with `milliseconds`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -306,6 +316,22 @@ public class CallState(
     private val _settings: MutableStateFlow<CallSettingsResponse?> = MutableStateFlow(null)
     public val settings: StateFlow<CallSettingsResponse?> = _settings
 
+    private val _durationInMs = flow {
+        while (scope.isActive) {
```
DanielNovak: I think we could use `currentCoroutineContext().isActive` here - this way we don't need to specify  `CoroutineScope`. I did some quick tests and it seems to behave the same.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -306,6 +316,22 @@ public class CallState(
...
+            delay(1000)
+            val started = _session.value?.startedAt
+            val ended = _session.value?.endedAt ?: OffsetDateTime.now()
+            val difference = if (started == null) null else {
+                ended.toInstant().toEpochMilli() - started.toInstant()?.toEpochMilli()!!
```
DanielNovak: If you remove the `?` in this line, then you can write it without `!!` - the non-null is inferred.
```
ended.toInstant().toEpochMilli() - started.toInstant().toEpochMilli()
```

stream-video-android-core/src/androidTest/kotlin/io/getstream/video/android/core/LivestreamTest.kt
```diff
@@ -89,60 +85,26 @@ class LivestreamTest : IntegrationTestBase() {
...
+        // counts and startedAt
+        assertThat(call.state.participants.value).isNotEmpty()
+        assertThat(call.state.startedAt).isNotNull()
+        assertThat(call.state.participantCounts.value?.total).isEqualTo(1)
+        assertThat(call.state.participantCounts.value?.anonymous).isEqualTo(0)
     }
 
     @Test
-    @Ignore
     fun timeRunning() = runTest {
```
DanielNovak, line 99: Running just this test alone (from AS) will result in:
```
kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 10000 ms
at kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:191)
```
but it runs fine if you run all the LiveStreamTest tests at  once. I think it relies on something / some state from previous tests.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
skydoves: Sounds better!

docusaurus/docs/Android/05-ui-cookbook/15-watching-livestream.mdx
skydoves: Nice catch!

stream-video-android-core/src/androidTest/kotlin/io/getstream/video/android/core/LivestreamTest.kt
DanielNovak: But sometimes it runs fine 🤷‍♂️ . Something is flaky here. I think this was fixed with my PR (https://github.com/GetStream/stream-video-android/pull/700) today, so maybe we just need to rebase.

DanielNovak: Ok, this is solved with my PR - so after locally rebasing with `develop` it runs fine.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
tschellenbach: thx

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
tschellenbach: updated the docs.

tschellenbach:
we should get this live so other SDK teams can review. 

@tbarbugli we do need to decide on how to expose the duration? (bit tricky since you have the .live thing)

PR #699, opened 2023-07-27
Cancel ringing if timeout period is met

DanielNovak:
This adds auto-cancel functionality to ringing. We will need to write a test for it, but I ran into some issues with the `delay` in the timer and I will do it in a separate PR (it may need some further changes in our code to allow this level of testing coroutines with `advanceTimeBy`).


skydoves (approved):
Looks good to me!

PR #700, opened 2023-07-27
Add way to supply mock RtcSession / MediaManager instances

DanielNovak:


stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/base/IntegrationTestBase.kt
```diff
@@ -59,25 +61,28 @@ object IntegrationTestState {
...
 
     var nextEventContinuation: Continuation<VideoEvent>? = null
     var nextEventCompleted: Boolean = false
 
     init {
         MockKAnnotations.init(this, relaxUnitFun = true)
+    }
 
+    @Before
+    fun setupVideo() {
```
DanielNovak, line 87: I moved all this code deliberately from `init { ... }` to `@Before fun ...` because the test dispatcher was set too late in the `TestBase`, the `init {...}` was already finished by that time and it caused issues (e.g. the `StreamVideoImpl.scope` was a real Dispatchers.IO and not a test dispatcher.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/stories/RingTest.kt
```diff
@@ -60,8 +65,37 @@ class RingTest : IntegrationTestBase() {
...
+
+        // This is needed because we need to wait for call.reject() internally to finish
+        // Is there a way we could wait for the suspend function?
+        runBlocking {
+            delay(500)
+        }
+
+        // We verify that the call doesn't ring anymore and the state is back to Idle
+        assertThat(client.state.ringingCall.value).isNull()
+        assertThat(call.state.ringingState.value is RingingState.Idle)
```
skydoves: We should use [turbine](https://github.com/cashapp/turbine) to wait to get the emitted flow values instead of delaying with static durations.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/CallStateTest.kt
```diff
@@ -176,12 +175,12 @@ class CallStateTest : IntegrationTestBase() {
...
-        val speakingWhileMuted = call.state.speakingWhileMuted.testIn(backgroundScope)
+        val speakingWhileMuted = call.state.speakingWhileMuted
         call.state.markSpeakingAsMuted()
         call.state.markSpeakingAsMuted()
         call.state.markSpeakingAsMuted()
-        assertTrue(speakingWhileMuted.expectMostRecentItem())
         // The flag should automatically reset to false 2 seconds
-        assertFalse(speakingWhileMuted.awaitItem())
+        advanceTimeBy(2500)
+        assertFalse(speakingWhileMuted.value)
```
skydoves: This is not a good idea - writing a test code based on very static delaying instead of observation, which means if we have much of this code, we will need to wait much time to run unit test codes. We can nicely resolve this with [turbine](https://github.com/cashapp/turbine).

skydoves (commented):
After reviewing the unit tests, I suggest considering an improvement by adopting Turbine, as described below. This approach will eliminate the need for manual delay and value manipulation of `StateFlows`, streamlining our testing process and enhancing its effectiveness

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/base/IntegrationTestBase.kt
skydoves: This is a good point.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/CallStateTest.kt
DanielNovak: This [advanceTimeBy](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/advance-time-by.html) advances the virtual time instantly, so the test actually runs instantly. This makes it possible to test for example timers in our code - I verify that after 2000+ seconds the timer resets the value. With Turbine I would just verify that the value resets, but not if the time is correct (it would also pass the test if the value reset after 500ms instead of 2500ms).

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/stories/RingTest.kt
DanielNovak: Yes, but in this is quite problematic - I was not able to do it yesterday, I will continue today. The StateFlow in this case is emitting multiple values as it progresses through the call state. Is there an easy way to wait for a specific StateFlow value? (or maybe verify that the StateFlow emits a specific sequence of values).

I was not able to remove the delay because internally our `wrapAPICall` is a `suspend` function and it's a real API call to our endpoints. We use a `UnconfinedTestDispatcher` in our `DispatcherRule` class therefore the execution order of queued coroutines is not guaranteed. And I don't yet know how we can wait for the API call to finish until we proceed.

DanielNovak: Ok, I fixed this. I used `waitForNextEvent<CallRejectedEvent>()` - so the code waits until the call is automatically rejected by our internal implementation. I had to modify our `waitForNextEvent` implementation because it was only advancing the virtual time while waiting for events - this doesn't work if we do real API calls in unit tests. So I added ` withContext(Dispatchers.Default.limitedParallelism(1))` inside `waitForNextEvent`.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/CallStateTest.kt
skydoves: Sounds sweet! Nicely done.

skydoves (approved):
Looks good to me! 👍

PR #702, opened 2023-07-31
Fix crash if same call is rejoined

DanielNovak:
The `StreamVideoImpl` is keeping a list of calls we created / joined. The problem is that we never remove calls from the list once we leave/close them. This causes an exception if you re-join the same call again while the application is running.


skydoves (approved):
Looks good to me!

PR #707, opened 2023-08-01
Configure different environments

JcMinarro:
### 🎯 Goal
Configure different environment per flavor


skydoves (approved):
Looks really good to me!

PR #708, opened 2023-08-01
Use local microphone for volume indicator during call

DanielNovak:
The `audioLevelChanged` events are too slow for displaying the local participant (user) volume. For this case (same like on iOS) we will use a the local microphone for getting the audio level without a delay.

I refactored the code so that the `Call` object will now have a `localMicrophoneAudioLevel` StateFlow which can be observed to get this value. But it is not necessary if you use the SDK Compose views - the `ParticipantVideo` drawn by the `CallContent` will already use this local audio level automatically.

I added a `RampValueUpAndDownHelper` class which smoothes out the audio levels and gradually ramps up the audio level to the new value and then goes back to 0. This greatly simplifies the UI part - you can now just observe the value and draw the indicator directly (the value is 0.0 - 1.0f, so you can just use it as height percent) without having to create custom animations. 

And similar like iOS we have now disabled the audio levels for other participants due to the lag (until we have a fix).

Note: We will need to make a separate task for adding this volume indicator to the pre-join view. At this point the `Call.localMicrophoneAudioLevel` will be returning 0f because the call isn't connected yet and WebRTC isn't yet capturing the mic. We will probably need to attach the mic and then detach it - just for the sake of the volume indicator.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/SoundInputProcessor.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/SoundInputProcessor.kt
@@ -16,8 +16,9 @@
 
 package io.getstream.video.android.core.call.utils
 
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlin.math.log10
-import kotlin.math.pow
+import kotlin.math.sqrt
 
 /**
  * The average decibel value is calculated within the last X milliseconds defined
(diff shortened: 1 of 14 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/SoundInputProcessor.kt
```diff
@@ -54,26 +62,24 @@ internal class DecibelThresholdDetection(
...
-    private fun calculateAverageDecibelPower(pcmData: ByteArray): Double {
+    private fun calculateAverageDecibelPower(soundData: ByteArray): Double {
         // For code clarity the following code expects the pcmData to be in PCM 16bit format.
         // And this is the default format used by the JavaAudioDeviceModule, so unless someone
         // in this class will override the default value then we are safe.
+        val samples = ShortArray(soundData.size / 2) { i ->
+            ((soundData[i * 2 + 1].toInt() and 0xFF) shl 8 or (soundData[i * 2].toInt() and 0xFF)).toShort()
+        }
 
-        val referencePressure = 0.00002 // 20 µPa (reference pressure for 0 dB)
```
DanielNovak, line 62: (I used this opportunity here to simplify my original code for decibel calculations)

skydoves (approved):
Looks great to me! Nice improvements 💯

skydoves:
@DanielNovak We need to resolve some conflicts before merging this. (maybe it's because of my spotless updates 😄 )

PR #709, opened 2023-08-01
enable ice restarts

tschellenbach:
Ice restarts on both the publisher and subscriber peer connection are now supported on the SFU. This pull request enables them on android. This will improve the reliability of calls when the network has some temporary problems.

SFU switching is still disabled.


skydoves (approved):
I've tested restarting publisher/subscriber ice with the debugging options, and seems to work really well! 🎉

PR #721, opened 2023-08-03
Bump Spotless to 6.20.0

### LINKED ISSUE #720: Bump spotless to 6.20.0
Bump spotless to 6.20.0.

skydoves:
Bump Spotless to 6.20.0. Resolve #720


JcMinarro (changes requested):
Could you consider to Squash all commits on this PR into one and then include the Commit Hash to the `.git-blame-ignore-revs file?
This file allows us to skip any commit from git-blame that doesn't apport anything to the history.
In that case, this commit is the one where the check style has been updated, but the code has no feature change.
[Here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) you can read docs about it.

JcMinarro (approved):

PR #724, opened 2023-08-03
Fix: ask for push permission in dogfooding app in QR code flow

DanielNovak:
The meeting can be opened the first time directly from QR code or automatically from Google Play referrer. In this case we jump directly to the meeting - leaving no time to ask for push notification permission. As a quick solution we can wait in the `DeeplinkingActivity` until the permission is given and the proceed. Not waiting causes multiple permission requests to be fired with random / wrong behaviour.


dogfooding/src/main/kotlin/io/getstream/video/android/DeeplinkingActivity.kt
```diff
@@ -50,7 +55,7 @@ class DeeplinkingActivity : ComponentActivity() {
                 Box(
                     modifier = Modifier
                         .fillMaxSize()
-                        .background(VideoTheme.colors.appBackground),
+                        .background(Colors.background),
```
DanielNovak, line 58: Wrong background was used here (white instead of dark)

PR #725, opened 2023-08-03
Use dark theme for demo app

DanielNovak:
The demo/dogfooding apps are dark and the theme should be dark too - it's noticable when you start the app and the screen is white and then switches to dark background once the MainActivity is drawn. This makes the app start animation a bit more pleasing.


skydoves (approved):
Looks good to me! Nice catch.

PR #726, opened 2023-08-03
Fix back button behaviour in demo app

DanielNovak:
Back button was not working during call in demo app because the Chat was capturing the `onBackPressed` event.


skydoves (approved):
Looks good to me

PR #727, opened 2023-08-03
Disable mic in demo app if call opened directly after install

DanielNovak:
For better UX we will disable the microphone in the demo app if you install the app from QR code and you are directly connected to a call (skipping the Lobby screen which gives you the option to enable/disable mic and video).


skydoves (approved):
Looks good to me 💯

Issue #679, opened 2023-07-21
Scale type for video rendering don't work


kosancicivan:
**Describe the bug**
I added videoScalingType property in MainActivity of tutorial-video sample
```
VideoRenderer(
                modifier = Modifier.weight(1f),
                call = call,
                video = remoteVideo,
                videoScalingType = VideoScalingType.SCALE_ASPECT_FIT
      )
```
When running tutorial video sample, video aspect ratio of web sample participant, stays the same. Nothing changes, video is largely clipped, same as if videoScalingType is not set. 

**SDK version**
- 0.2.0

**Expected behavior**
As stated in documentation:
SCALE_ASPECT_FIT - video frame is scaled to fit the size of the view by maintaining the aspect ratio (black borders may be displayed).

**Device:**
 - Xiami redmi note 8 and emulator
 - Android version: [11 and 13]

skydoves:
Hey @kosancicivan , thanks for reporting this issue. It seems to be related to the rendering TextureView. I will get back to you soon once we find any improvements for this.

DanielNovak:
@kosancicivan - just a little update. We found the problem, it's in our webrtc-android library - the [VideoTextureViewRenderer](https://github.com/GetStream/webrtc-android/blob/main/stream-webrtc-android-ui/src/main/kotlin/io/getstream/webrtc/android/ui/VideoTextureViewRenderer.kt) is missing a `requestLayout()` inside `updateFrameData()`. We need to test it a bit and we will then release a new version of the webrtc-anroid library and then a new SDK version. This shouldn't take long. We will keep you updated.

skydoves:
Hey @kosancicivan, we've released a new version `0.3.1`, which resolves this issue. Would you check out if it works well on your side? Thanks!

skydoves:
I will close this issue since you've resolved it in the recent release. Thanks for reporting this issue! @kosancicivan

kosancicivan:
Works as expected, thanks

PR #741, opened 2023-08-07
Fix aspect ratio not working for VideoContent

DanielNovak:
Changing the `VideoScalingType` didn't have an effect because of a bug in webrtc-android (https://github.com/GetStream/webrtc-android/pull/46). 

Issue originally reported here: https://github.com/GetStream/stream-video-android/issues/679 

Notice that I changed the aspect ration from BALANCED to FILL because that's what we actually want. It looks the same as before. Our VideoContent wasn't actually using BALANCED scale type - it was ignored due to the bug.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -41,7 +41,7 @@ threetenAbp = "1.4.6"
 tink = "1.9.0"
 turbine = "0.13.0"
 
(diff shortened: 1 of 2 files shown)


skydoves (approved):
Looks really great to me. Nicely done! 🙌

PR #742, opened 2023-08-07
Fix encryption failing after restore

DanielNovak:
Client applications that use `allowBackup=true` (such us our demo app) will also back up files and preferences created by the Video SDK. If encryption is enabled (`encryptPreferences` is set to true in `StreamVideoBuilder`) then our encrypted data is backed up and we won't be able to decrypt them on restore (by definition - since it's encrypted with device keys). In this case the only thing we can do is to remove any local data and reinitialise the Datastore again. 

**How to test:**
You can run the script from https://developer.android.com/guide/topics/data/testingbackup#Preparing to back up the application, uninstall it and restore it. Our current develop will fail with:

```
Fatal Exception: java.security.InvalidKeyException: Keystore cannot load the key with ID: master_key
       at com.google.crypto.tink.integration.android.AndroidKeystoreAesGcm.<init>(AndroidKeystoreAesGcm.java:61)
       at com.google.crypto.tink.integration.android.AndroidKeystoreKmsClient.getAead(AndroidKeystoreKmsClient.java:179)
       at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.readMasterkeyDecryptAndParseKeyset(AndroidKeysetManager.java:366)
       at com.google.crypto.tink.integration.android.AndroidKeysetManager$Builder.build(AndroidKeysetManager.java:298)
       at io.getstream.video.android.datastore.delegate.StreamUserDataStore$Companion.install(StreamUserDataStore.kt:174)
       at io.getstream.video.android.datastore.delegate.StreamUserDataStore$Companion.install$default(StreamUserDataStore.kt:155)
       at io.getstream.video.android.core.notifications.internal.VideoPushDelegate$userDataStore$2.invoke(VideoPushDelegate.kt:46)
       at io.getstream.video.android.core.notifications.internal.VideoPushDelegate$userDataStore$2.invoke(VideoPushDelegate.kt:45)
```

### DIFF
--- a/stream-video-android-datastore/src/main/kotlin/io/getstream/video/android/datastore/delegate/StreamUserDataStore.kt
+++ b/stream-video-android-datastore/src/main/kotlin/io/getstream/video/android/datastore/delegate/StreamUserDataStore.kt
@@ -25,6 +25,7 @@ import com.google.crypto.tink.KeyTemplates
 import com.google.crypto.tink.aead.AeadConfig
 import com.google.crypto.tink.integration.android.AndroidKeysetManager
 import io.getstream.log.StreamLog
+import io.getstream.log.taggedLogger
 import io.getstream.video.android.datastore.model.StreamUserPreferences
 import io.getstream.video.android.datastore.serializer.UserSerializer
 import io.getstream.video.android.datastore.serializer.encrypted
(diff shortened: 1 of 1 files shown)


skydoves (approved):
Looks good to me! 💯

PR #745, opened 2023-08-08
Fix - only first screenshare is working properly

DanielNovak:
Only first screenshare (from desktop) was working. Stopping and starting the screenshare would not work. I introduced this bug in https://github.com/GetStream/stream-video-android/pull/653 (by mistake I thought that this part of code is only local participant track).


skydoves (approved):
Looks great to me!

PR #746, opened 2023-08-08
Fix screenshare zoom doesn't work

DanielNovak:
The zoomable view will only work correctly with `wrapContentSize()`. Using `fillMaxSize()` will make the image the wrong size and crop the screenshared frame (can't be zoomed out).


skydoves (approved):
Looks good to me!

PR #751, opened 2023-08-09
Don't crash if PiP fails

DanielNovak:
The try/catch needs to be inside the Handler to catch the Exception. Eventually we will need to fix the cause of the crash - we are enabling PiP too late in onPause, it should be done in `onUserLeaveHint()` - but we don't have that callback in Compose.


skydoves (approved):
Looks good to me!

PR #758, opened 2023-08-13
Fix PiP mode in gesture navigation

DanielNovak:
This should fix the PiP issues on devices with gesture navigation. The lifecycle onPause() event is called too late. We need to listen to the Activity's onPause() event and then enter PiP. Ideally we should do it in `Activity.onUserLeaveHint` ([link](https://developer.android.com/reference/android/app/Activity#onUserLeaveHint())), but we haven't found a way how to get this callback into our `MediaPiPLifecycle` composable without asking the developers to forward us those events.


skydoves (approved):
Looks good to me! Perfect! 💯

PR #777, opened 2023-08-19
Remove internal usage of StreamUserDataStore - leave the persistence to client level

DanielNovak:
The purpose of this PR is to remove the internal persistence layer in the SDK which stores the `User` and token related information. Similar to iOS we want to leave the implementation of the storage to the client side. Some clients may want to use special encryption, or only in-memory storage or other strategies. For convenience we will leave the `StreamUserDataStore` in the sample. This data store is completely detached from the Stream Video SDK and the client application has to manually store the data after login and then retrieve it when initialising the SDK.

**Changes in StreamVideoBuilder API:**
- none - except for removal of the optional `encryptPreferences` boolean which now doesn't have a use anymore

**Changes in SDK behaviour**:
- There is a change if push notifications need to be handled. The Stream SDK doesn't know how to initialise automatically because it doesn't know the token or User. The client application should initialise the SDK in `Application.onCreate()` otherwise push notifications will be ignored until the SDK is initialised. In the previous version this would work out-of-box but only if encyption was enabled (otherwise the SDK would fail at trying to decrypt data that is not encrypted).
  - We should later provide a way to "lazily initialise" the Stream Video SDK after a push notification is received. The client could supply a lambda containing the initialisation instead of initialising it. 

**Main changes:** 
- `StreamUserDataStore` usage is completely removed from the SDK
- the only persisted data in the SDK is the device push token information in a new (internal) `DeviceTokenStorage` 
- uncessary singleton usage for token storage replaced with normaln instances instead 
- demo application manages the `User` and token with help of `StreamUserDataStore` (but we could also use SharedPreferences or any other mechanism)
- demo app will now directly jump into the CallJoinScreen (home screen) if the user is already logged in
- demo SDK initialisation simplified (code is in one place in `StreamVideoInitHelper`)


skydoves (approved):
Looks good to me and seems to work well as we discussed 👍

PR #793, opened 2023-08-24
Reduce size of call sound mp3 and avatar placeholder

DanielNovak:
The call wait sound (mp3) is taking up 700kB. I have reduced it to 90kB without any noticeable impact on quality.
And the avatar placeholder is now taking up 21kB instead of 350kB (png -> webp).

Uncessary ~1MB saved in SDK.


skydoves (approved):
Looks perfect 💯

PR #799, opened 2023-08-28
Scan QR code button in demo

DanielNovak:
This adds a simple way to directly scan meeting QR codes from the demo app without having to open the camera app.


dogfooding/src/main/kotlin/io/getstream/video/android/ui/DogfoodingNavHost.kt
```diff
@@ -38,6 +46,34 @@ fun AppNavHost(
...
+                context.startActivity(DeeplinkingActivity.createIntent(context, callId))
+            } else {
+                Toast.makeText(
+                    context,
+                    "Unrecognised meeting QR code format",
+                    Toast.LENGTH_SHORT,
+                ).show()
+            }
+        }
+    }
```
skydoves: I'm curious about the rationale behind exposing the `qrScannedCallback` within the `NavHost` rather than within the `CallJoinScreen`. It seems like we could simplify this by introducing a straightforward composable function named `rememberQRScannedCallback` and locating it within the `CallJoinScreen`.

DanielNovak: No rationale here :-). Good point. I changed it.

skydoves (approved):

PR #829, opened 2023-09-20
Add video filter support

DanielNovak:
Add support for custom Video filters. 

Easy setup - you only need to set `Call.videoFilter` and either create an instance of `BitmapVideoFilter` (easy to use, you get a bitmap which you can mutate) or `RawVideoFilter` which gives you the `VideoFrame` directly and has no overhead.


skydoves (approved):
Looks good to me! Nicely done.

PR #850, opened 2023-10-06
Remove hardcoded channel info

aleksandar-apostolov:
Add channel ID, Title and Description into xml resources. Use them instead of the hard-coded ones.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -73,3 +73,16 @@ public fun Activity.shouldShowRequestPermissionsRationale(permissions: Array<out
...
 }
+
+/**
+ * Get a string or a default value.
+ *
+ * @param context an Android context (possibly applicationContext)
+ * @param resId the resource ID
+ * @param default the default value if the string could not be retrieved from strings.
+ */
+internal fun stringOrDefault(context: Context, resId: Int, default: String) = try {
```
DanielNovak: Do we need this? 🤔 It will never be null because the default string is defined in the strings.xml in the SDK. So if you override it and return null then that's a developer error. 

I wonder if we could just use:

```
open fun getChannelName(): String =  application.applicationContext.getString( R.string.stream_video_incoming_call_notification_channel_title)
```

And we could just delete `CHANNEL_NAME` and the others. 
There is no need to keep the strings defined in the code.

aleksandar-apostolov: True, its defensive measure to account for that exact case. The other option is to use getString directly as you suggest, thus reducing the need to have the CHANNEL_NAME and the rest in the companion object

DanielNovak: In that case I think it's not worth to have this defensive measure - that is more like a developer mistake and we prefer the "fail fast" approach in these cases so that they discover the issue quickly (wrong configuration). Otherwise the code will hide the problem for you and you will think you changed something but the code will silently ignore. So I suggest we just delete the companion object constants.

aleksandar-apostolov: Sure, will do that.

DanielNovak (approved):

PR #854, opened 2023-10-09
Reaction dialog

aleksandar-apostolov:
Add a reaction dialog with the option to visually choose and send the correct reaction

### DIFF
--- a/dogfooding/src/main/kotlin/io/getstream/video/android/ui/call/ReactionsMenu.kt
+++ b/dogfooding/src/main/kotlin/io/getstream/video/android/ui/call/ReactionsMenu.kt
@@ -1,0 +1,201 @@
+/*
+ * Copyright (c) 2014-2023 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-video-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:OptIn(ExperimentalLayoutApi::class)
+
+package io.getstream.video.android.ui.call
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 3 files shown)


dogfooding/src/main/kotlin/io/getstream/video/android/ui/call/ReactionsMenu.kt
```diff
@@ -0,0 +1,201 @@
...
+        shape = RoundedCornerShape(2.dp),
+    ).wrapContentWidth()
+    val onEmojiSelected: (emoji: String) -> Unit = {
+        sendReaction(scope, call, it, onDismiss)
+    }
+
+    Dialog(onDismiss) {
+        Card(
+            Modifier.wrapContentWidth(),
+        ) {
```
skydoves: What about using an explicit parameter here? 'modifier = Modifier...'

dogfooding/src/main/kotlin/io/getstream/video/android/ui/call/ReactionsMenu.kt
```diff
@@ -0,0 +1,201 @@
...
+                        reactionText = "Smile",
+                        onEmojiSelected = onEmojiSelected,
+                    )
+                    ReactionItem(
+                        modifier = modifier,
+                        reactionMapper = reactionMapper,
+                        emojiCode = ReactionsMenuData.reactions[5],
+                        reactionText = "Heart",
+                        onEmojiSelected = onEmojiSelected,
+                    )
```
skydoves: I think we can simplify here using `forEach` like the example below:

```kotlin
ReactionsMenuData.reactions.forEach { reaction ->
  ReactionItem(
                        modifier = modifier,
                        reactionMapper = reactionMapper,
                        emojiCode = reaction.emojiCode,
                        reactionText = reaction.text,
                        onEmojiSelected = onEmojiSelected,
                    )  
}
```

dogfooding/src/main/kotlin/io/getstream/video/android/ui/call/ReactionsMenu.kt
```diff
@@ -0,0 +1,201 @@
...
+import io.getstream.video.android.core.Call
+import io.getstream.video.android.core.mapper.ReactionMapper
+import io.getstream.video.android.mock.StreamMockUtils
+import io.getstream.video.android.mock.mockCall
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+private object ReactionsMenuData {
+    const val raiseHand = ":raise-hand:"
+    val reactions = listOf(":fireworks:", ":hello:", ":like:", ":hate:", ":smile:", ":heart:")
```
skydoves: Let's create a data object, which contains `emojiCode` and `text` to match them together in the code below.

skydoves (commented):
Overall looks good to me! I left some trivial comments related to the object and formats.

PR #859, opened 2023-10-11
Update scroll order

aleksandar-apostolov:
Update scroll order and extract sort order state in separate State object.


skydoves (approved):
Overall looks good to me. Nicely done! 🙌

PR #873, opened 2023-10-17
Foreground service & Ongoing call notification

aleksandar-apostolov:
Add an ongoing call notification backed by foreground service to keep the call alive and allow the user to return to, or leave, the call.

Docs:
Updated the documentation & guides with new page "Keeping the call alive"

Updated "Push notification -> Setup" to include the ONGOING_CALL action.


skydoves (approved):
Overall very looks good to me. Nicely done! 🙌

PR #900, opened 2023-11-02
Use generated ID as name for random user

liviu-timar:
### 🎯 Goal

Set generated ID for random users so that it's used in the Outgoing and Incoming call screens.

### 🛠 Implementation details

Pass the generated user id to the user object.

### 🎨 UI Changes

image


DanielNovak:
You can also use a fallback logic in the screen. It probably always just reads the name - but it should fallback to userId if name is not set (some other platform may forget to set the name). We already do this fallback in some other screens.

DanielNovak (approved):

PR #909, opened 2023-11-09
Dismiss direct call notification when call is cancelled

liviu-timar:
### 🎯 Goal

The direct call notification should be dismissed when the call initiator cancels the call.

### 🛠 Implementation details

Added and handled a new notification type for direct call cancellation.

### ☑️Reviewer Checklist


DanielNovak:
This was also a feature request from RN/Flutter teams I think - but we need to ping the backend devs first because there is no backend support for this yet. I will try to find the thread where we discussed it.

aleksandar-apostolov:
This will be handled in https://github.com/GetStream/stream-video-android/pull/956

PR #910, opened 2023-11-09
Implement Play In-app Updates for demo app

liviu-timar:
### 🎯 Goal

Implement in-app updates in the demo app so we can nudge the user to install the latest version.

### 🛠 Implementation details

- Using the ktx Play In-app Updates library
- Getting update availability in the View Model and exposing it as a Flow
- Handling updates in the MainActivity as an immediate (full-screen) update

### 🎨 UI Changes

| Update exists screen | Update in progress | Update canceled by user | 
| --- | --- | --- |
| image | image | image |

### 🧪 Testing

By using [Play Internal App Sharing](https://play.google.com/console/internal-app-sharing).


dogfooding/src/main/kotlin/io/getstream/video/android/MainActivity.kt
```diff
@@ -66,6 +73,52 @@ class MainActivity : ComponentActivity() {
                     )
                 }
             }
+
+            checkForAppUpdates(viewModel.appUpdateResultFlow)
+        }
+    }
+
+    private suspend fun checkForAppUpdates(appUpdateResultFlow: Flow<AppUpdateResult>) {
```
DanielNovak: Can we move this somewhere into a different class / util object? Only to keep the MainActivity a bit more simple because this code is noise for external developers that want to see how to use our SDK and this code will confuse them.

dogfooding/src/main/kotlin/io/getstream/video/android/MainActivity.kt
```diff
@@ -66,6 +73,52 @@ class MainActivity : ComponentActivity() {
...
+                    result.startImmediateUpdate(appUpdateActivityResultLauncher)
+                }
+                is AppUpdateResult.InProgress -> Log.d(IN_APP_UPDATE_LOG_TAG, "Update in progress")
+                is AppUpdateResult.Downloaded -> Log.d(IN_APP_UPDATE_LOG_TAG, "Update downloaded")
+                is AppUpdateResult.NotAvailable -> Log.i(IN_APP_UPDATE_LOG_TAG, "None available")
+            }
+        }
+    }
+
+    private val appUpdateActivityResultLauncher = registerForActivityResult(
```
DanielNovak: Same - is it maybe possible to move this elsewhere?

dogfooding/src/main/kotlin/io/getstream/video/android/MainActivity.kt
liviu-timar: Good idea. I've created a a helper class and moved as much code as possible into it.

PR #911, opened 2023-11-09
Show avatar when video not available, add padding to grid layout for 6 participants

aleksandar-apostolov:
Add bottom grid padding to improve visibility in 3x3 layout. 
Always render the video fallback behind the video for better experience to avoid black screen instead of e.g. the user avatar.

### DIFF
--- a/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/VideoRenderer.kt
+++ b/stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/VideoRenderer.kt
@@ -93,7 +93,7 @@ public fun VideoRenderer(
         return
     }
 
-    // Sow avatar always behind the video.
+    // Show avatar always behind the video.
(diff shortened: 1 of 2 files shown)


stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/VideoRenderer.kt
```diff
@@ -93,6 +93,9 @@ public fun VideoRenderer(
         return
     }
 
+    // Sow avatar always behind the video.
```
DanielNovak: show :)

DanielNovak (approved):

stream-video-android-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/VideoRenderer.kt
aleksandar-apostolov: :) I need new keyboard

PR #913, opened 2023-11-10
Modules refactoring

aleksandar-apostolov:
Goal: Reduce overall complexity and consolidate modules.
——————
Action:
“stream-video-android-tooling” - merge with dogfooding,
Reason:
Only used there and the code consists of only one debugging activity and a utility function to create “toasts”.
——————
Actions:
“stream-video-android-model” - move into “core” under a model package name.
“stream-video-android-datastore” - move into “core” under a ‘datastore’ package name. If StreamUserDataStore was interface, I would suggest having the interface in core and implementation maybe in separate module.
Reason:
The question I asked to suggest these moves -> Can core be used without datastore / model? If no, then merge the modules. Unless someone wants to use the UI artefacts without the core modules. But current setup doesn’t allow for this.
——————
Actions:
Rename “stream-video-android-xml” to “stream-video-android-ui-xml”
Rename “stream-vdeio-android-compose” to “stream-video-android-ui-compose”
Rename “stream-video-android-ui-common” to “stream-video-android-ui-core”
Reason:
Better clarity on the dependencies between UI modules and what each contains.
——————
Action:
Rename “stream-video-android-mock” to “stream-video-android-ui-previewdata”
Reason:
Only one test uses the “mock” its called BaseComposeTest. The rest usage is into @Preview composable in dogfooding, app and compose modules. Thus the name mock makes no sense.
——————
Action:
“Dogfooding” should became “app” and maybe configure based on flavour if ever.
Reason:
“app” seems to not have reason to exist.
——————


DanielNovak (approved):

skydoves (approved):
Looks good to me! Let me follow up generating baseline profiles once this PR is merged.

PR #929, opened 2023-11-17
Dynascale

aleksandar-apostolov:
Change publishing quality based on sfu events.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
@@ -881,10 +881,9 @@ public class RtcSession internal constructor(
      * This is used for dynsacle
      */
     internal fun updatePublishQuality(event: ChangePublishQualityEvent) = synchronized(this) {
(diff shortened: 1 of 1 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -879,59 +880,36 @@ public class RtcSession internal constructor(
      * Change the quality of video we upload when the ChangePublishQualityEvent event is received
      * This is used for dynsacle
      */
-    internal fun updatePublishQuality(event: ChangePublishQualityEvent) {
-        if (publisher == null) {
-            return
+    internal fun updatePublishQuality(event: ChangePublishQualityEvent) = synchronized(this) {
+        val sender = publisher?.connection?.transceivers?.firstOrNull {
```
DanielNovak, line 884: I would log if this is null probably, something like "received `ChangePublishQualityEvent` but can't handle it because no video transceiver available". We have had quite a few hard to debug problems in the past due to random `return` in the code flow.

DanielNovak (approved):

PR #931, opened 2023-11-17
Use flavor name constant for demo app

DanielNovak:
Just small improvement for demo app - don't use hardcoded flavor names.


skydoves (approved):
Looks good to me!

PR #942, opened 2023-11-22
Use new SimulcastAlignedVideoEncoderFactory

DanielNovak:
Use new `SimulcastAlignedVideoEncoderFactory` from webrtc-android. This is necessary to make sure that the camera stream is always aligned to work with VP8 encoding.
More information here: https://github.com/GetStream/webrtc-android/pull/85


skydoves (approved):
Looks good to me 👍

PR #949, opened 2023-11-29
Refactor call settings menu in demo app

liviu-timar:
### 🎯 Goal

Improve call settings menu. Refactor both UI and code.

### 🛠 Implementation details

- Extracted menu entries as dedicated composable
- Grouped debug entries at the end
- Changed blur effect icon (with on/off variants)
- Replaced mic icon to fix bad alignment
- Moved the menu a bit up to not cover buttons

### 🎨 UI Changes

| Before | After |
| --- | --- |
| image | image |


demo-app/src/main/kotlin/io/getstream/video/android/ui/call/SettingsMenu.kt
```diff
@@ -131,56 +127,57 @@ internal fun SettingsMenu(
...
+                    },
+                    label = if (isBackgroundBlurEnabled) {
+                        "Disable background blur"
+                    } else {
+                        "Enable background blur (beta)"
+                    },
+                    onClick = {
+                        onToggleBackgroundBlur()
+
+                        if (call.videoFilter == null) {
```
DanielNovak, line 161: this code can be removed

DanielNovak (approved):

PR #954, opened 2023-12-05
Improve reactions in demo app

liviu-timar:
### 🎯 Goal

Fix reactions visibility issues. Also, make reactions more similar to iOS and web. Issue with more details [here](https://github.com/GetStream/android-video-issues-tracking/issues/53).

### 🛠 Implementation details

Modified the `CallScreen` composable and started using the `videoRenderer` and `floatingVideoRenderer` parameters. The video renderer has a `reactionContent` parameter, which I used to pass a `CustomReactionContent` composable.

Reaction centered position is temporary until new component design is implemented.

### 🧪 Testing

Send & receive each reaction in a call with web & iOS and with all layouts.


skydoves (approved):
Looks good to me! 💯

PR #956, opened 2023-12-05
Incoming calls handling

aleksandar-apostolov:
Handle incoming and ongoing calls via same CallService.
Update and read call state.
Connect coordinator socket and listen to call events.


demo-app/src/main/kotlin/io/getstream/video/android/DirectCallActivity.kt
```diff
@@ -54,6 +57,7 @@ class DirectCallActivity : ComponentActivity() {
 
     @Inject
     lateinit var dataStore: StreamUserDataStore
+    private var call: Call? = null
```
DanielNovak: I think a `lateinit var` is more appropiate here - then we don't need to have that many `!!` everywhere. Only in `onStop()` you can check for `::call.isInitialized`

aleksandar-apostolov: Great point

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -0,0 +1,266 @@
...
+ */
+internal class CallService : Service() {
+    private val logger by taggedLogger("CallService")
+    private var callId: StreamCallId? = null
+    private var callDisplayName: String? = null
+    private var observeRingingState: Job? = null
+    private var observeCallState: Job? = null
+    private var updateCallJob: Job? = null
+    private var startCoordinatorSocketJob: Job? = null
+    private val supervisorJob = SupervisorJob()
```
kanat: Hey @aleksandar-apostolov,

WDYT about creating a `serviceScope` here and use it instead of a `GlobalScope` later on? 🙂 
```
val serviceScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -0,0 +1,266 @@
...
+            }
+            observeCallState(callId!!, streamVideo!!)
+        }
+        return START_NOT_STICKY
+    }
+
+    @OptIn(DelicateCoroutinesApi::class)
+    private fun observeCallState(callId: StreamCallId, streamVideo: StreamVideo) {
+        // Ringing state
+        observeRingingState = GlobalScope.launch(supervisorJob + Dispatchers.IO) {
```
kanat: Then here we could use that `serviceScope` :)

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    override fun onDestroy() {
+        callId?.let {
+            val notificationId = callId.hashCode()
+            NotificationManagerCompat.from(this).cancel(notificationId)
+        }
+        // Stop any jobs
+        supervisorJob.cancel()
```
kanat: And then here
```kotlin
serviceScope.cancel()
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
aleksandar-apostolov: Its a great idea. Tnx for pointing it out.
We do not even need the SupervisorJob in this case.

DanielNovak (approved):

PR #963, opened 2023-12-08
Improve call state handling when app is killed or device is locked

liviu-timar:
### 🎯 Goal

- End an ongoing call when the app is killed
- Stop/start the video feed when the device is locked/unlocked.

### 🛠 Implementation details

- End an ongoing call when the app is killed: use the `onTaskRemoved` callback in `OngoingCallService` to leave the call.
- Stop/start the video feed: register a broadcast receiver in `OngoingCallService` that listens for `ACTION_USER_PRESENT` and `ACTION_SCREEN_OFF` broadcasts.

TBD:
- Another option for ending the call is to use the `stopWithTask` manifest flag for `OngoingCallService` and leaving the call in `onDestroy()`.
- Should the stop/start video feed functionality be default in the SDK or just in the Demo app?
- Should `ToggleCameraBroadcastReceiver` be placed in the `notifications.internal.receivers` package?
- Should `ToggleCameraBroadcastReceiver` inherit `GenericCallActionBroadcastReceiver` and be found via a `search` method in `DefaultStreamIntentResolver`? The advantage would be the call parameter that it receives, but it would also complicate the code.

### DIFF
--- a/stream-video-android-core/api/stream-video-android-core.api
+++ b/stream-video-android-core/api/stream-video-android-core.api
@@ -4096,6 +4096,12 @@ public final class io/getstream/video/android/core/notifications/NotificationHan
 	public static final field INTENT_EXTRA_NOTIFICATION_ID Ljava/lang/String;
 }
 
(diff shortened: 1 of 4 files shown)


aleksandar-apostolov (approved):
Looks good

PR #964, opened 2023-12-12
Add pre-push spotless check to prevent pushes if spotlessApply has not been run

aleksandar-apostolov:


liviu-timar:
This would be great 😄

skydoves (approved):
Looks very good to me!

PR #970, opened 2023-12-19
Do not leave call in PiP if phone is locked.

aleksandar-apostolov:
Call lifecycle is now handled via the service:
* Remove `CallLifecycle` and do not leave call when phone is locked in PiP.


liviu-timar:
Tested and it works. You're still in the call when you lock-unlock while in PiP. ✅

liviu-timar (approved):

PR #974, opened 2023-12-27
Remote config, new links handling and supported logins per flavor

aleksandar-apostolov:
* Add remote config (can be extended in future with other parameters)
* Add "demo" and "pronto" links in xml and allow for both paths to be handled.
* Default to path variable for call id, fallback to query parameter
* Environment switcher (based on config with available environments per flavor)
* Login buttons on login screen are now based on remote config.


demo-app/src/main/kotlin/io/getstream/video/android/DeeplinkingActivity.kt
```diff
@@ -72,8 +72,9 @@ class DeeplinkingActivity : ComponentActivity() {
             }
         }
 
+        val callIdFromExtra = intent?.getStringExtra(CALL_ID)
         val data: Uri = intent?.data ?: return
-        val callId = data.getQueryParameter("id") ?: return
+        val callId = callIdFromExtra ?: extractCallId(data) ?: return
```
DanielNovak: We need to remove these two `return` here - they cause the activity to just show an endless loading if this is actually null and it's not obvious what is wrong.

I suggest something like this:

```
        if (data == null) {
            logger.e { "Can't open the call from deeplink because intent data is null" }
            finish()
            return
        }


        if (callId == null) {
            logger.e { "Can't open the call from deeplink because call ID is null" }
            finish()
            return
        }
```

demo-app/src/main/AndroidManifest.xml
```diff
@@ -83,25 +83,14 @@
...
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
-
                 <data android:scheme="https" />
-                <data android:host="@string/deeplinking_host_legacy" />
-                <data android:pathPrefix="@string/deeplinking_path_prefix" />
+                <data android:host="getstream.io" />
+                <data android:host="pronto.getstream.io" />
+                <data android:pathPrefix="/join" />
+                <data android:pathPrefix="/video/demos" />
```
DanielNovak: There is a bug - if you enter https://getstream.io/video/ on mobile browser and then click `Explore demo` it will open `https://getstream.io/video/demos` - which will then open our app and but without a call ID.

DanielNovak:
Is the login page with three buttons (email / google / random) also shown in Google Play builds? It shouldn't be shown in this case. Just checking :-).

demo-app/src/main/AndroidManifest.xml
DanielNovak: Maybe we just need to change this:
`<data android:pathPrefix="/video/demos" />`
to this?
`<data android:pathPrefix="/video/demos/join" />`

I tried that change and it seems to fix the issue (the meeting is now correctly opened in the app if I click Explore Demo on the website).

If that is not sufficient then we may need to add `pathPattern` or `pathAdvancedPattern` or `pathSuffix` (https://developer.android.com/guide/topics/manifest/data-element)

aleksandar-apostolov:
> Is the login page with three buttons (email / google / random) also shown in Google Play builds? It shouldn't be shown in this case. I ask because I built the productionDebug version and this now shows the login picker which is not correct. Just checking :-).

If the automatic login fails the page is shown, we can control which options are shown via the configuration. The environment switch is not shown as for production there is only one environment which is "demo".

In the previous case if login fails (i.e. no internet) you are just stuck on the screen. In the new handling, you can try to proceed by choosing a login type. Login types are configurable per flavour, (currently they are the same for both flavours)

DanielNovak (approved):

PR #977, opened 2024-01-09
Add call sounds

liviu-timar:
### 🎯 Goal

Add sounds for incoming/outgoing calls. Added only incoming and outgoing sounds for now, **no busy or cancel**.

### 🛠 Implementation details

- Implemented with a `MediaPlayer` controlled in `CallService`.
- Now the foreground service starts for outgoing calls as well. A specific notification is shown.
- Sounds can be customized through a `Sounds` class that's passed to `StreamVideoBuilder`.
- Some client changes were needed.

### 🎨 UI Changes

- A sounds plays on both the incoming and outgoing devices.
- A notification now appears on the outgoing device. Clicking on it opens the outgoing screen. When the call is accepted, the notification is replaced by the _ongoing call_ notification.

### 🧪 Testing

- [ ] Test all actions to see if sounds start/stop correctly: accept, reject on incoming, cancel on outgoing. Both in foreground and in background.
- [ ] Interact with outgoing call notification. Check strings.
- [ ] Check PR comments.
- [ ] Customize sounds by replacing one with another and passing `null` to disable.
- [ ] Call multiple users.
- [ ] Check docs [here](https://staging.getstream.io/video/docs/android/ui-cookbook/incoming-and-outgoing-call/#sounds) and KDocs in code for new types.

#### General

#### Code & documentation
- [X] Affected documentation updated (KDocs, docusaurus, tutorial)

### ☑️Reviewer Checklist
- [ ] Tutorial starter kit


demo-app/src/main/kotlin/io/getstream/video/android/DirectCallActivity.kt
```diff
@@ -171,13 +171,6 @@ class DirectCallActivity : ComponentActivity() {
         }
     }
 
-    override fun onStop() {
```
liviu-timar, line 174: Do you see any problem in not ending the call here?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
```diff
@@ -110,30 +110,34 @@ class ClientState(client: StreamVideo) {
         removeRingingCall()
     }
 
-    fun addRingingCall(call: Call) {
-        // TODO: behaviour if you are already in a call
+    fun addRingingCall(call: Call, ringingState: RingingState) {
```
liviu-timar, line 113: I am starting `CallService` in this method. I need to know what trigger to send to the service, so I also added a `ringingState` parameter. `addRingingCall` is called in places where we know if the call is incoming or outgoing, so is it ok to add this param?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -266,7 +266,7 @@ public class Call(
         response.onSuccess {
             state.updateFromResponse(it)
             if (ring) {
-                client.state.addRingingCall(this)
+                client.state.addRingingCall(this, RingingState.Outgoing())
```
liviu-timar, line 269: Is `Call.create` with `ring=true` the only way to create an outgoing call? I am relying on it to play the outgoing sound (in the service).

See `addRingingCall` [comment](https://github.com/GetStream/stream-video-android/pull/977/files#r1447059871) below.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -84,6 +86,7 @@ public class StreamVideoBuilder @JvmOverloads constructor(
     private val videoDomain: String = "video.stream-io-api.com",
     private val runForegroundServiceForCalls: Boolean = true,
     private val localSfuAddress: String? = null,
+    private val sounds: Sounds = Sounds(),
```
liviu-timar, line 89: I've added a new `sounds` parameter in `StreamVideoBuilder`. I then pass it to `StreamVideoImpl` so it can be used wherever needed. It was the closest I could get to the iOS implementation.

iOS uses an [Appearance](https://github.com/GetStream/stream-video-swift/blob/main/Sources/StreamVideoSwiftUI/Appearance.swift) class to customize the sounds. It passes it to a [StreamVideoUI](https://github.com/GetStream/stream-video-swift/blob/main/Sources/StreamVideoSwiftUI/StreamVideoUI.swift) class, which is used to initialize the video client, as described [here](https://getstream.io/video/docs/ios/basics/quickstart/#setting-up-the-streamvideoui-object).

The Sounds class is identical to the [iOS one](https://github.com/GetStream/stream-video-swift/blob/main/Sources/StreamVideoSwiftUI/Sounds.swift).

Is this way of providing sounds ok? Another way would have been to use the Compose Theme, but we're using the sounds in `CallService`.

liviu-timar, line 89: How would a client app customize the sounds (or change other parameters, for that matter), once the SDK is initialized?

aleksandar-apostolov: You can always reload the SDK with different parameters.

aleksandar-apostolov: I think here is better because ideally the sounds should also be used in the XML module.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
aleksandar-apostolov: Just as a thought, maybe we need to deprecate `addRingingCall` and use create new API `addIncommingCall` and `addOutgoingCall` call.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -216,14 +216,15 @@ internal class CallService : Service() {
             logger.w { "Foreground service did not start!" }
             stopService()
         } else {
+            initializeCallAndSocket(streamVideo!!, callId!!)
```
liviu-timar, line 249: This makes the outgoing call screen state updates more reliable when the app is in the background (before it sometimes missed events and the screen was still in _calling_ state - will handle in another PR).

liviu-timar: Do you see any problem in always connecting to the web socket when the fg service starts?

DanielNovak (approved):

PR #978, opened 2024-01-09
Use different video filter fo the screen share specifically different…

aleksandar-apostolov:
Due to shared `surfaceTextureHelper` with the camera screenshareing from Android would result in mixed source where random frames are shown from the camera or Screenshare. Participant video would remain blank and after disabling screenshare participant video is no longer available until screenshare is started again.

This PR fixes the above issue by adding a new video filter for screenshare source using the `surfaceTextureHelper` from the `screenShare` manager

The issue: screensharing from the left device (pixel 7), results in camera frames mixed with the screenshare track, while also the participant video track is blank on the receiver end.
image


DanielNovak (approved):
wow, great catch!

PR #984, opened 2024-01-15
Ensure main activity is finished when starting deeplinking and ensure…

aleksandar-apostolov:
Ensure all permissions are granted before going to the call screen.
Also ensure that MainActivity is finished because it causes double socket connection.


DanielNovak (approved):

skydoves (approved):
Looks good to me 🙌

PR #1007, opened 2024-02-05
Redesign components and demo app

aleksandar-apostolov:
This PR aims to redesign the components offered by the SDK as well as the Demo app.
Currently there is a new `VideoTheme` under `base` package because the old theme is used in the tutorials. We should delete the old `VideoTheme` once the tutorials are updated.

The redesign consists of:
- Adding new layer of reusable components to the SDK i.e. `base` components. This includes, `StreamButton` `StreamIconButton` etc.. 
- Adapting existing components to use the `base` components.
- Using the new design and components into demo app.
- Adapting the demo screens to the new design

Additional changes:
- Add "Call recording start/stop" to demo app
- Add "Share call" and "Participant list" to demo app

Screenshot_20240205-135941
Screenshot_20240205-140145
Screenshot_20240205-135846
Screenshot_20240205-135915


demo-app/src/main/kotlin/io/getstream/video/android/ui/call/ShareCall.kt
```diff
@@ -0,0 +1,160 @@
...
+            )
+            Spacer(modifier = Modifier.size(16.dp))
+            Row(
+                modifier = Modifier.fillMaxWidth(),
+                horizontalArrangement = Arrangement.SpaceBetween,
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                Row {
+                    Text(
+                        text = "Call ID:",
```
DanielNovak: add space after `:` (otherwise the actual ID is right next to the text)

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/ShareCall.kt
```diff
@@ -0,0 +1,160 @@
...
+@Composable
+public fun ShareCallWithOthers(
+    modifier: Modifier = Modifier,
+    call: Call,
+    clipboardManager: ClipboardManager?,
+    env: State<StreamEnvironment?>,
+    context: Context,
+) {
+    ShareSettingsBox(modifier, call, clipboardManager) {
+        val link = if (env.value?.env == "demo") {
```
DanielNovak: You can use the `StreamEnvironments.demo` constant here.

DanielNovak (approved):

PR #1023, opened 2024-03-05
Add audio only call content

aleksandar-apostolov:
Transfer the `AudioCallContent` from the sample app and extend our own SDK with this UI.
The UI is almost the same as outgoing call with the difference that there is ongoing timer of the call duration, much like in the dialer app. This is to be used with calls who are "audio only" and where the "audio room" is not the intended use-case.

image


stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/AudioCallContent.kt
```diff
@@ -0,0 +1,140 @@
...
+    },
+    durationPlaceholder: String = "",
+    isShowingHeader: Boolean = true,
+    headerContent: (@Composable ColumnScope.() -> Unit)? = null,
+    detailsContent: (
+        @Composable ColumnScope.(
+            participants: List<MemberState>,
+            topPadding: Dp,
+        ) -> Unit
+    )? = null,
```
skydoves, line 88: We're missing Kdocs here.

skydoves (approved):
Overall looks good to me, @aleksandar-apostolov! 💯

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/AudioCallContent.kt
aleksandar-apostolov: Added

PR #1025, opened 2024-03-06
Update theme

aleksandar-apostolov:
Unify the VideoTheme with the new design.


DanielNovak (approved):

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/StreamColors.kt
```diff
@@ -17,229 +17,99 @@
...
+    val buttonAlertPressed: Color,
+    val buttonAlertDisabled: Color,
+    val iconDefault: Color,
+    val iconPressed: Color,
+    val iconActive: Color,
+    val iconAlert: Color,
+    val iconDisabled: Color,
+    val alertSuccess: Color,
+    val alertCaution: Color,
+    val alertWarning: Color,
```
skydoves, line 64: This PR has already merged, but it would be good to have Kdocs for these properties for providing better developer experiences 😄

PR #1038, opened 2024-03-18
Fix concurrent modification exception

aleksandar-apostolov:
This PR fixes a bug that would cause a crash when modifying the `subscribers` for call events at the same time when `fireEvent` is invoked.


skydoves (approved):
Looks good to me!

PR #1051, opened 2024-04-01
Speakerphone fallback update

aleksandar-apostolov:
When disabling the speakerphone in `setSpeakerPhone(boolean)` the `fallback` device is set to `the first which is NOT a speakerphone`. Extend the api into `setSpeakerPhone(boolean, StreamAudioDevice) to allow for a preference when selecting a fallback.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
@@ -132,7 +132,7 @@ class SpeakerManager(
      * setSpeakerPhone(enable, StreamAudioDevice.Earpiece)
      * ```
      *
-     * @param enable if ture, enables the speakerphone, if false disables it and selects another device.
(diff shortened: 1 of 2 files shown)


DanielNovak (approved):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
```diff
@@ -122,8 +122,20 @@ class SpeakerManager(
...
+     * Enables or disables the speakerphone.
+     *
+     * When the speaker is disabled the device that gets selected next is by default the first device
+     * that is NOT a speakerphone. To override this use [defaultFallback].
+     * If you want the earpice to be selected if the speakerphone is disabled do
+     * ```kotlin
+     * setSpeakerPhone(enable, StreamAudioDevice.Earpiece)
+     * ```
+     *
+     * @param enable if ture, enables the speakerphone, if false disables it and selects another device.
```
liviu-timar: Typo "ture" instead of "true"

liviu-timar (approved):

PR #1060, opened 2024-04-09
Out of the box `CallActivity` (StreamCallActivity, ComposeStreamCallActivity) added

aleksandar-apostolov:
## TL;DR
A default out of the box activity to handle calls. Lots of extension possibilities.
Part of the TL version of this comment will be used as documentation moving forward.

## Goal & Why

The goal of this PR is to add a default activity that will handle most of the call use-cases out of the box.
This is intended to replace `DirectCallActivity`, `IncomingCallActivity` and `CallActivity` in `demo-app` as well as `CallActivity` into the `AudioSampleApp`.
This will allow for customers with a simple use-case to integrate the call UI as a standalone component as straight forward as possible.

## How

- `stream-video-ui-core` - added a `StreamCallActivity` abstract activity class.
This class handles most of the `call` state and `actions`. It handles call connection, disconnection based on intent actions. It also adds an API for possible extension and more fine grained control if extended.
- `stream-video-ui-core` - added `StreamActivityUiDelegate<T>`* interface
This interface has one method. Its `setContent` and is called into `StreamCallActivity.onCreate(..)` to allow to set the content of the activity. The content can be `@Com`posable` or `xml` binding.
- `stream-video-ui-compose` - added `ComposeStreamCallActivity`.
This is an actual activity that can be invoked. It extends `StreamCallActivity` and provides a `StreamCallActivityComposeDelegate` which handles the activity UI in `Compose`.
- `stream-video-ui-compose` - added `StreamCallActivityComposeDelegate` which extends `StreamActivityUiDelegate` and adds a `Compose` content into the activity. This delegate proceeds to further use the already existing compose components basically `RingingCallContent` and provides several extension points.
- `stream-video-ui-compose` - added the `ComposeStreamCallActivity` into `AndroidManifest.xml` as a default action handler for all the call actions.

## Extension possibility
- `StreamCallActivity`
The `StreamCallActivity` abstract class consists of several sections. You can extend the lifecycle method with ensured `Call` object e.g. `onCreate(Call)` instead of the normal `onCreate`, you can extend and choose what happens when `VideoCallEvent` is received or when `CallAction` is taken. Also there is a possibility to extend the `leave()` or `accept()` methods but these are marked with a delicate API warning because if not written correctly they will affect the whole call state and activity. Finally there is a possibility to completely override the UI by providing a different `StreamActivityUiDelegate` by overriding the `uiDelegate()` method.
- `StreamCallActivityComposeDelegate` provides several compose functions that can be overridden (all or one at a time) and you can provide such instance in `uiDelegate()` to extend the `Compose` UI of the `ComposeStreamCallActivity`

## How to use
By default we only support compose out-of-the-box UI so by adding the `stream-video-ui-compose` module the `ComposeCallActivity` will already be added as a default handler for the call intents.

To use out of the box solution one must add `stream-video-ui-compose` as a dependency and then simply use the `ComposeStreamCallACtivity` as any other activity.

Example on how to place an outgoing call.
```kotlin
val intent = StreamCallActivity.callIntent(
                                this,
                                StreamCallId("audio_call", UUID.randomUUID().toString()),
                                members,
                                true,
                                action = NotificationHandler.ACTION_OUTGOING_CALL,
                                clazz = ComposeStreamCallActivity::class.java
                            )
                            startActivity(intent)
```

Both `StreamCallActivity` and `ComposeCallActivity` are open classes and can be extended.

Simple extension of `ComposeCallActivity` can look like:
```kotlin
class MyCallActivity : ComposeStreamCallActivity() {

    override fun onCallEvent(call: Call, event: VideoEvent) {
        super.onCallEvent(call, event)
        // Do something...
    }
}
```

To extend only one composable the minimum implementation is as follows:
```kotlin
class MyUiDelegate : StreamCallActivityComposeDelegate() {
    @Composable
    override fun StreamCallActivity.NoAnswerContent(call: Call) {
        Box(modifier = Modifier.fillMaxSize()) {
            Text(text = "This is custom UI when there is no answer for the call!")
        }
    }
}

class MyCustomUiActivity : ComposeStreamCallActivity() {
    @Suppress("UNCHECKED_CAST")
    override fun <T : StreamCallActivity> uiDelegate(): StreamActivityUiDelegate<T> {
        return MyUiDelegate() as StreamActivityUiDelegate<T>
    }
}
```

## Notes
1. I was thinking of skipping compose delegate and add the methods directly into `StreamCallActivity` but this way we lock the default out of the box handling for compose only. This way at least call management and state can be utilised also for xml UI users.
2. Lots of stuff is open for extension, this is on purpose as closing some methods will limit the functionality and purpose of the class.
3. Contrary to number `2` the default `onCreate` methods of the `StreamCallActivity` are `final`. An alternative methods are provided which have the `Call` as parameter. The reason for this is that if someone were to override the `onCreate` then it would destroy the purpose of the `StreamCallActivity` in which case a copy & paste or complete new activity is better.


stream-video-android-ui-compose/src/main/AndroidManifest.xml
```diff
@@ -14,4 +14,30 @@
...
+
+
+    <application>
+        <activity
+        android:name=".ui.ComposeStreamCallActivity"
+        android:supportsPictureInPicture="true"
+        android:showOnLockScreen="true"
+        android:showWhenLocked="true"
+        android:launchMode="singleTop"
+        android:exported="true">
```
DanielNovak: should be probably `false`

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/StreamCallActivityComposeDelegate.kt
```diff
@@ -0,0 +1,263 @@
...
+        io.getstream.video.android.compose.ui.components.call.activecall.AudioCallContent(
+            onBackPressed = {
+                onBackPressed(call)
+            },
+            call = call,
+            isMicrophoneEnabled = micEnabled,
+            onCallAction = {
+                onCallAction(call, it)
+            },
+            durationPlaceholder = duration ?: "Calling...",
```
DanielNovak: hardcoded text - better to use string resource here

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/StreamCallActivityComposeDelegate.kt
```diff
@@ -0,0 +1,263 @@
...
+    @Composable
+    public open fun StreamCallActivity.NoPermissions(
+        granted: List<String> = emptyList(),
+        notGranted: List<String> = emptyList(),
+        showRationale: Boolean,
+    ) {
+        StreamDialogPositiveNegative(
+            content = {
+                Text(
+                    text = "Some permissions are required",
```
DanielNovak: string resources here too (in whole class)

PR #1074, opened 2024-04-18
Prevent overwrite of User properties on backend

liviu-timar:
### 🎯 Goal

Send `null` if `User` properties are empty.

Thread [here](https://getstream.slack.com/archives/C022N8JNQGZ/p1713367514869789) and [here](https://getstream.slack.com/archives/C06QKTWP9RU/p1713366615668659).

### 🛠 Implementation details

Pass `null` to `ConnectUserDetailsRequest` and `UserRequest` for each empty `User` property.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoImpl.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoImpl.kt
@@ -453,8 +453,8 @@ internal class StreamVideoImpl internal constructor(
             val response = createGuestUser(
                 userRequest = UserRequest(
                     id = user.id,
-                    image = user.image.takeUnless { it.isEmpty() },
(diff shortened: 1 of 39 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoImpl.kt
```diff
@@ -453,9 +453,9 @@ internal class StreamVideoImpl internal constructor(
             val response = createGuestUser(
                 userRequest = UserRequest(
                     id = user.id,
-                    image = user.image,
-                    name = user.name,
-                    custom = user.custom,
+                    image = user.image.takeUnless { it.isEmpty() },
+                    name = user.name.takeUnless { it.isEmpty() },
```
aleksandar-apostolov: Maybe `isBlank()` is better?

skydoves (approved):
Overall looks good to me!

PR #1078, opened 2024-04-23
Call activity replace

aleksandar-apostolov:
Replace the various activities in `demo-app` with the `StreamCallActivity` implementation.


skydoves (approved):
Overall looks good to me. Nicely done!

It would be also good to have some docs about `StreamCallActivity` and  `ComposeStreamCallActivity` using this demo code as an example.

PR #1080, opened 2024-04-24
Allow null User fields

liviu-timar:
### 🎯 Goal

Made some of `User` fields nullable and `null` by default.

In `CoordinatorSocket.authenticate()` we pass:
- `null` if we don't want the field's value to change
- `""` or `emptyMap()` or `emptyList()` if we want to reset the field
- a `value` if we want to set the field
- `null` if the field is blank (whitespace)

### 🧪 Testing

❗ Needs testing

### DIFF
--- a/demo-app/src/main/kotlin/io/getstream/video/android/ui/join/CallJoinScreen.kt
+++ b/demo-app/src/main/kotlin/io/getstream/video/android/ui/join/CallJoinScreen.kt
@@ -237,7 +237,7 @@ private fun CallJoinHeader(
         Text(
             modifier = Modifier.weight(1f),
             color = Color.White,
-            text = user?.name?.ifBlank { user?.id }?.ifBlank { user!!.custom["email"] }.orEmpty(),
+            text = user?.userNameOrId ?: "",
             maxLines = 1,
             fontSize = 16.sp,
         )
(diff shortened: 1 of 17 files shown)


demo-app/src/main/kotlin/io/getstream/video/android/ui/join/CallJoinScreen.kt
```diff
@@ -234,7 +234,7 @@ private fun CallJoinHeader(
         Text(
             modifier = Modifier.weight(1f),
             color = Color.White,
-            text = user?.name?.ifBlank { user?.id }?.ifBlank { user!!.custom["email"] }.orEmpty(),
+            text = user?.userNameOrId ?: "",
```
skydoves: We can make this `user?.userNameOrId.orEmpty()`

demo-app/src/main/kotlin/io/getstream/video/android/util/StreamVideoInitHelper.kt
```diff
@@ -168,8 +168,8 @@ object StreamVideoInitHelper {
 
         val chatUser = io.getstream.chat.android.models.User(
             id = user.id,
-            name = user.name,
-            image = user.image,
+            name = user.name ?: "",
+            image = user.image ?: "",
```
skydoves: We can also utilize `orEmpty()` here as well.

skydoves (approved):
Looks good to me! I left trivial comments for syntax.

demo-app/src/main/kotlin/io/getstream/video/android/ui/join/CallJoinScreen.kt
liviu-timar: Thanks for noticing. Used it in another place, but missed these ones. 👍

aleksandar-apostolov (approved):

PR #1081, opened 2024-04-29
Improve peer connection health check to prevent reconnecting

liviu-timar:
### 🎯 Goal

When user does not have `send-audio` or `send-video` permissions the publisher peer connection (`RtcSession.publisher`) is `null`. Take this into account when checking the peer connection health in `CallHealthMonitor.check()` and prevent infinite reconnections.

[Reported](https://getstream.slack.com/archives/C06QKTWP9RU/p1714071364865939) by Patreon in their audio room setup.

### 🛠 Implementation details

Check if user can publish video or audio and take this into account in `CallHealthMonitor.check()`.


skydoves (approved):
Looks good to me!

aleksandar-apostolov (approved):

PR #1121, opened 2024-06-26
[PBE-4806] Add example to prevent crash that appears when StreamCallActivity is recreated after being in background

liviu-timar:
### 🎯 Goal

Add Demo app code to illustrate how to rebuild SDK in overridden `onPreCreate()`.

### 🛠 Implementation details

Override `onPreCreate` in `CallActivity` and reload SDK if needed.


skydoves (approved):
Looks good to me

PR #1123, opened 2024-06-27
[PBE-4805] Trust additional CAs on older Android versions

liviu-timar:
### 🎯 Goal

Add new ISRG Root X1 and ISRG X2 CAs to support Let's Encrypt shorter certificate chain on Android 7.0+.

### 🛠 Implementation details

- Added certificates as `pem` files to `res/raw`.
- Edited `network_security_config.xml` file to use the two new certificates.
- `AndroidManifest` uses the `network_security_config.xml` file in the `application` tag: `android:networkSecurityConfig="@xml/network_security_config"`


aleksandar-apostolov:
Don't merge this, its just an example.

PR #1127, opened 2024-07-03
Refresh token on the coordinator socket

skydoves:
Refresh token on the coordinator socket.


aleksandar-apostolov:
This will not fix the issue we have.

liviu-timar:
Solved in another way due to error appearing on a different code path (socket `onFailure` callback).

PR #1137, opened 2024-07-10
Catch unhandled exception in video client scope

aleksandar-apostolov:
# TL;DR 
Added scope error handler to catch all `scope` exceptions and don't crash the SDK.

# Details
When a socket error would occur be it 401 or any other, it would eventually end up in one of our error flows. But in case the socket is disconnected with a valid token which in the meantime becomes invalid trying to connect the token would return `connection.error` event. This event is propagated correctly to the error flows, however the `continuation` also needs to resume and is resumed with this exception. Since there is no exception handling for this, the coroutine would crash and take the SDK down as well as the integrating app.
This was happening with very limited tokens i.e. one that expires in 1-2min.

### DIFF
--- a/stream-video-android-core/api/stream-video-android-core.api
+++ b/stream-video-android-core/api/stream-video-android-core.api
@@ -4459,9 +4459,6 @@ public final class io/getstream/video/android/core/sounds/Sounds {
 
 public final class io/getstream/video/android/core/utils/AndroidUtilsKt {
 	public static final fun safeCall (Ljava/lang/Object;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
-	public static final fun safeCall (Lkotlin/jvm/functions/Function0;)V
-	public static final fun safeSuspendingCall (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public static final fun safeSuspendingCall (Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun shouldShowRequestPermissionsRationale (Landroid/app/Activity;[Ljava/lang/String;)Z
 }
 
--- a/stream-video-android-core/api/stream-video-android-core.api
+++ b/stream-video-android-core/api/stream-video-android-core.api
@@ -269,6 +269,7 @@ public final class io/getstream/video/android/core/ClientState {
 	public final fun getRingingCall ()Lkotlinx/coroutines/flow/StateFlow;
 	public final fun getUser ()Lkotlinx/coroutines/flow/StateFlow;
 	public final fun handleEvent (Lorg/openapitools/client/models/VideoEvent;)V
+	public final fun hasActiveOrRingingCall ()Z
 	public final fun removeActiveCall ()V
 	public final fun removeRingingCall ()V
 	public final fun setActiveCall (Lio/getstream/video/android/core/Call;)V
@@ -4460,6 +4461,10 @@ public final class io/getstream/video/android/core/sounds/Sounds {
 }
 
 public final class io/getstream/video/android/core/utils/AndroidUtilsKt {
+	public static final fun safeCall (Ljava/lang/Object;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public static final fun safeCall (Lkotlin/jvm/functions/Function0;)V
+	public static final fun safeSuspendingCall (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun safeSuspendingCall (Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun shouldShowRequestPermissionsRationale (Landroid/app/Activity;[Ljava/lang/String;)Z
 }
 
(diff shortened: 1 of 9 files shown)


stream-video-android-core/api/stream-video-android-core.api
```diff
@@ -4460,6 +4458,10 @@ public final class io/getstream/video/android/core/sounds/Sounds {
 }
 
 public final class io/getstream/video/android/core/utils/AndroidUtilsKt {
+	public static final fun safeCall (Ljava/lang/Object;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
```
aleksandar-apostolov, line 4461: We need to make this `internal` along with the other safe calls.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -84,3 +93,75 @@ internal fun <T> combineComparators(vararg comparators: Comparator<T>): Comparat
         comparator.thenComparing(combined)
     }
 }
+
```
aleksandar-apostolov, line 93: Make these functions `internal`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
```diff
@@ -16,10 +16,12 @@
 
 package io.getstream.video.android.core
 
-import android.content.Context
+import android.util.Log
```
aleksandar-apostolov: Use `logger by taggedLogger` instead

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
```diff
@@ -80,11 +84,21 @@ class ClientState(client: StreamVideo) {
...
+    /**
+     * Returns true if there is an active or ringing call
+     */
+    fun hasActiveOrRingingCall(): Boolean = safeCall(false) {
+        val hasActiveCall = _activeCall.value != null
+        val hasRingingCall = _ringingCall.value != null
+        val activeOrRingingCall = hasActiveCall || hasRingingCall
+        logger.d { "[hasActiveOrRingingCall] active: $hasActiveCall, ringing: $hasRingingCall" }
+        activeOrRingingCall
+    }
```
skydoves, line 96: Is there a specific reason we need to use `safeCall` here? It seems unlikely that an exception would occur in this context.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/PersistentSocket.kt
```diff
@@ -415,30 +419,94 @@ public open class PersistentSocket<T>(
...
+        logger.d { "[sendHealthCheck] Sending..." }
         val healthCheckRequest = HealthCheckRequest()
         socket?.send(healthCheckRequest.encodeByteString())
     }
 
+    /**
+     * Check if the socket is in a state that can connect.
+     */
+    internal fun canConnect(): Boolean = safeCall(false) {
+        val connectionState = connectionState.value
```
skydoves, line 441: Same here, do we need to use `safeCall` here and the similar methods below?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+@file:OptIn(
+    ExperimentalContracts::class,
+    ExperimentalContracts::class,
+    ExperimentalContracts::class,
```
skydoves: It seems to be duplicated by mistake

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -84,3 +93,75 @@ internal fun <T> combineComparators(vararg comparators: Comparator<T>): Comparat
...
+/**
+ * Safely call a function and handle exceptions.
+ *
+ * @param block the function to call.
+ */
+fun safeCall(block: () -> Unit) {
+    contract {
+        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
+    }
+    try {
```
skydoves: We can make this inline function

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -84,3 +93,75 @@ internal fun <T> combineComparators(vararg comparators: Comparator<T>): Comparat
...
+ * Safely call a function and handle exceptions.
+ *
+ * @param default the default value to return in case of an exception.
+ * @param block the function to call.
+ */
+fun <T> safeCall(default: T, block: () -> T): T {
+    contract {
+        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
+    }
+    return try {
```
skydoves: We can make this inline function as well

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -84,3 +93,75 @@ internal fun <T> combineComparators(vararg comparators: Comparator<T>): Comparat
...
+ */
+suspend fun safeSuspendingCall(block: suspend () -> Unit) {
+    contract {
+        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
+    }
+    try {
+        block()
+    } catch (e: Exception) {
+        // Handle or log the exception here
+        println("Exception occurred: ${e.message}")
```
skydoves: What about using the stream logger instead? Other methods are the same.

skydoves (commented):
The primary idea for fixing this issue looks good! I left some trivial comments.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
aleksandar-apostolov: Because we are accessing the state flows which can crash, I don't want to have a case where the flow is unavailable and it crashes, just want to return false in that case.

PR #1140, opened 2024-07-12
[PBE-5147] Prevent exception caused by CallService not being promoted to foreground fast enough

liviu-timar:
### 🎯 Goal

In some cases, `startForeground` is not called fast enough and `stopSelf` is called before it, which causes a crash.

JIRA issue: https://stream-io.atlassian.net/browse/PBE-5147

### 🛠 Implementation details

- Added new `maybePromoteToForegroundService` method that calls `startForeground` early in `onStartCommand` if there isn't already an active call.

### 🎨 UI Changes

- Refactored and cleaned strings for incoming, outgoing, ongoing notifications and notification channels.

### 🧪 Testing

Tested:
- Notifications get updated correctly (incoming, outgoing, ongoing).
- _Incoming call while already in a call_ scenario. 
- If notification ID is correct for the temporary notification.  
- If channel is correctly set.

#### General
- [x] PR is linked to the Jira issue it resolves


skydoves (approved):
Looks good to me 👍

PR #1148, opened 2024-07-23
[PBE 5272] Stop foreground service when uninstalling StreamVideo instance

liviu-timar:
### 🎯 Goal

Calling `StreamVideo.removeClient()` doesn’t end up calling `maybeStopForegroundService()`, so any ongoing call notification will still show after removeClient()

Ideally all calls are explicitly left first, but it would be nice if `removeClient()` was a reliable way to fully nuke everything Stream Video related.

### 🎫 JIRA Issue

https://stream-io.atlassian.net/browse/PBE-5272

### 🛠 Implementation details

- Called `leave` instead of `cleanup` for active call in `StreamVideoImpl#cleanup`.
- `leave` will eventually stop the foreground service. 
- Added `safeCall` to `Call#leave()`.


kanat (approved):
LGTM

PR #1152, opened 2024-08-05
[PBE-5431] Meet Android 14 foreground service requirements

liviu-timar:
### 🎯 Goal

Meet foreground service requirements introduced in Android 14.

### 🛠 Implementation details

- Create new utility extension function that calls `startForeground` with a service type for Android 14+.
- Use this function where needed in `CallService` and `StreamScreenShareService`
- Checked manifest, `startForeground` calls and permissions.
- Checked:
  - `foregroundServiceType` in `Manifest`
  - Foreground service permissions
  - `serviceType` in `startForeground` calls
  - Related runtime permissions: `MANAGE_OWN_CALLS`, calling `createScreenCaptureIntent()`.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
@@ -171,7 +171,14 @@ inline fun <T> safeCall(default: T, block: () -> T): T {
     }
 }
 
-internal fun Service.customStartForeground(
+/**
+ * Start a foreground service with a service type to meet requirements introduced in Android 14.
+ *
+ * @param notificationId The notification ID
+ * @param notification The notification to show
+ * @param trigger The trigger that started the service: [TRIGGER_ONGOING_CALL], [TRIGGER_OUTGOING_CALL], [TRIGGER_INCOMING_CALL], [TRIGGER_SHARE_SCREEN]
+ */
+internal fun Service.startForegroundWithServiceType(
     notificationId: Int,
     notification: Notification,
     trigger: String,
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
@@ -183,20 +183,25 @@ internal fun Service.startForegroundWithServiceType(
     notification: Notification,
     trigger: String,
 ) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
-        val foregroundServiceType = when (trigger) {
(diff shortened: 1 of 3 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -162,3 +170,33 @@ inline fun <T> safeCall(default: T, block: () -> T): T {
...
+        ServiceCompat.startForeground(
+            this,
+            notificationId,
+            notification,
+            foregroundServiceType,
+        )
+    } else {
+        startForeground(notificationId, notification)
+    }
+}
```
skydoves: I believe we can just use the `ServiceCompat.startForeground` function instead of separating it into using `startForeground,` as `ServiceCompat.startForeground` already implemented the details depending on the Android version.

```kotlin
    public static void startForeground(@NonNull Service service, int id,
            @NonNull Notification notification, int foregroundServiceType) {
        if (Build.VERSION.SDK_INT >= 34) {
            Api34Impl.startForeground(service, id, notification, foregroundServiceType);
        } else if (Build.VERSION.SDK_INT >= 29) {
            Api29Impl.startForeground(service, id, notification, foregroundServiceType);
        } else {
            service.startForeground(id, notification);
        }
    }
```

So following the internal implementation, we can pass the `foregroundServiceType` after S(`Build.VERSION.SDK_INT >= 29`) and we also need to handle the `ForegroundServiceStartNotAllowedException` following the [official documentation](https://developer.android.com/develop/background-work/services/foreground-services#start).

skydoves (commented):
Looks good to me! I just left a comment about the `ServiceCompat`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
liviu-timar: At first I intended to just use `ServiceCompat`, judging by it's name, but I got a lint error in AS. But probably I missed something. Good catch, I'll have a second look!

aleksandar-apostolov: Hey @liviu-timar can we update this to only use the `ServiceCompat` after which we can merge the PR.

liviu-timar: Yep, will do it soon/today. 👍

liviu-timar: Did the changes, ready for merging with the caveat that I'm not sure that the `PHONE_CALL` type is appropriate, as per [docs](https://developer.android.com/about/versions/14/changes/fgs-types-required#phone-call).

aleksandar-apostolov: We will update this separately by providing new service types.

PR #1154, opened 2024-08-08
Do not leave the call when `StreamCallActivity` goes into background

aleksandar-apostolov:
Fix a bug that would cause a call to `call.leave()` in `StreamCallActivity.onStop()`

### DIFF
--- a/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
+++ b/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
@@ -383,11 +383,8 @@ public abstract class StreamCallActivity : ComponentActivity() {
      * @param call the call
      */
     public open fun onStop(call: Call) {
+        // Extension point only.
         logger.d { "Default activity - stopped (call -> $call)" }
-        if (isVideoCall(call) && isInPictureInPictureMode) {
-            logger.d { "Default activity - stopped: PiP detected, will leave call. (call -> $call)" }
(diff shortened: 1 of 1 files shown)


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -384,8 +384,8 @@ public abstract class StreamCallActivity : ComponentActivity() {
      */
     public open fun onStop(call: Call) {
         logger.d { "Default activity - stopped (call -> $call)" }
-        if (isVideoCall(call) && !isInPictureInPictureMode) {
-            logger.d { "Default activity - stopped: No PiP detected, will leave call. (call -> $call)" }
+        if (isVideoCall(call) && isInPictureInPictureMode) {
```
liviu-timar: - Lock while in the foreground: does not leave call ✅ 
- Lock while in PiP: leaves call ❓ 

Video [here](https://drive.google.com/file/d/1jUKKryD9w6btLFNuc7WZQdcMdwVBRBky/view?usp=sharing).

aleksandar-apostolov: I think we need to not do anything in this case. I will update the PR.

liviu-timar: Does not leave the call anymore. ✅

liviu-timar (approved):

PR #1164, opened 2024-08-23
[PBE-5674] Add multiple services with different types for livestreaming calls

aleksandar-apostolov:
# Goal
The goal of this PR is to make available different types of `CallService` because not all services require all the permissions that `CallService` requires. 

For example if you are having an app only as a `livestream` guest you don't need `phoneCall` service type, but rather `mediaPlayback`. Then you can remove the unnecessary permission in the manifest with, for example:

```xml
<uses-permission android:name="android.permission.USE_FULL_SCREEN_INTENT" tools:node="remove"/>
<uses-permission android:name="android.permission.MANAGE_OWN_CALLS" tools:node="remove"/>
<uses-permission android:name="android.permission.FOREGROUND_SERVICE_PHONE_CALL" tools:node="remove"/>
```

### DIFF
--- a/stream-video-android-core/src/main/AndroidManifest.xml
+++ b/stream-video-android-core/src/main/AndroidManifest.xml
@@ -104,7 +104,7 @@
             android:exported="false" />
 
         <service
-            android:name=".notifications.internal.service.LivestreamMemberCallService"
+            android:name=".notifications.internal.service.LivestreamViewerService"
             android:foregroundServiceType="mediaPlayback"
             android:exported="false" />
     </application>
--- a/stream-video-android-core/src/main/AndroidManifest.xml
+++ b/stream-video-android-core/src/main/AndroidManifest.xml
@@ -48,7 +48,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_CAMERA" />
 
-    <!-- If the app is only  -->
+    <!-- If the app is only livestream viewer/guest type -->
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
 
     <!-- If the calls support screensharing -->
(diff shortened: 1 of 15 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/LivestreamCallService.kt
```diff
@@ -0,0 +1,37 @@
...
+package io.getstream.video.android.core.notifications.internal.service
+
+import android.content.pm.ServiceInfo
+import io.getstream.log.TaggedLogger
+import io.getstream.log.taggedLogger
+
+/**
+ * Due to the nature of the livestream calls, the service that is used is of different type.
+ */
+internal open class LivestreamCallService : CallService() {
```
liviu-timar, line 26: Maybe `LivestreamHostCallService` and `LivestreamViewerCallService` are better names, quicker to understand?

aleksandar-apostolov: `Viewer` I agree, for the other I had it named exactly like that :) but I changed it since if you don't have two apps this is the only one that you need more or less.

stream-video-android-core/src/main/AndroidManifest.xml
```diff
@@ -34,12 +34,26 @@
...
+    <uses-permission android:name="android.permission.USE_FULL_SCREEN_INTENT" />
     <uses-permission android:name="android.permission.MANAGE_OWN_CALLS"/>
-    <uses-permission android:name="android.permission.VIBRATE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_PHONE_CALL" />
+
+    <!-- If the app is only livestream  host -->
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_CAMERA" />
+
+    <!-- If the app is only  -->
```
liviu-timar: "If the app is only livestream guest/viewer"

liviu-timar (approved):

PR #1171, opened 2024-08-29
Add audio only `CallService` to avoid requiring `CAMERA` permission

aleksandar-apostolov:
# Goal
Add an `audio only` service type for livestream to avoid requiring the `CAMERA` permission.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallServiceConfig.kt
```diff
@@ -62,6 +62,19 @@ public fun livestreamCallServiceConfig(): CallServiceConfig {
     )
 }
 
+/**
+ * Return a default configuration for the call service configuration for livestream which has no camera
```
liviu-timar, line 66: Wouldn't this type also apply to `audio_rooms`, so the description and name should reflect that?

skydoves (approved):
Looks good to me :)

PR #1178, opened 2024-09-13
[PBE-6015] Add extra info to request client header

liviu-timar:
### 🎯 Goal

Add more info to the `X-Stream-Client` header.

### 🛠 Implementation details

- Added app name to header. Will be provided by client app via new `StreamVideoBuilder` `appName` parameter. Falls back to `packageName` if argument is not provided.
- Added docs for new parameter.
- Removed _duplicated_ header that was added when creating socket. Left just the one added in `HeadersInterceptor#intercept`.

### 🎨 Changes

Example header before:
`stream-video-android-1.0.13|os=Android 14|api_version=34|device_vendor=samsung|device_model=SM-A155F`

Example header now:
`stream-video-android-1.0.13|os=Android 14|api_version=34|device_vendor=samsung|device_model=SM-A155F|app_name=Stream Video Demo App`
or
`stream-video-android-1.0.13|os=Android 14|api_version=34|device_vendor=samsung|device_model=SM-A155F|app_name=io.getstream.video.android.debug`


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -102,6 +103,7 @@ public class StreamVideoBuilder @JvmOverloads constructor(
     private val crashOnMissingPermission: Boolean = false,
     private val permissionCheck: StreamPermissionCheck = DefaultStreamPermissionCheck(),
     private val audioUsage: Int = defaultAudioUsage,
+    private val appName: String? = null,
```
liviu-timar, line 106: `appName` is optional. Should we make it mandatory?

aleksandar-apostolov: Can the packagename be the default if not supplied?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/PersistentSocket.kt
```diff
@@ -277,7 +276,6 @@ public open class PersistentSocket<T>(
             .url(url)
             .addHeader("Connection", "Upgrade")
             .addHeader("Upgrade", "websocket")
-            .addHeader("X-Stream-Client", StreamVideo.buildSdkTrackingHeaders())
```
aleksandar-apostolov, line 280: Why did we remove this?

liviu-timar: The header was duplicated when connecting the socket. It was added here and also in `HeadersInterceptor#intercept`. Tested with this line removed and it's added only once, both in `video/connect` and other requests, like `video/call`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
liviu-timar: I added the packageName as a fallback. The header when `appName` is not provided looks like: `stream-video-android-1.0.13|os=Android 14|api_version=34|device_vendor=samsung|device_model=SM-A155F|app_name=io.getstream.video.android.debug`

aleksandar-apostolov (approved):

PR #1179, opened 2024-09-16
Use device ringtone for incoming call sound

liviu-timar:
### 🎯 Goal

Use device ringtone for incoming call sound.

### 🛠 Implementation details

- Added `Sounds.DEVICE_INCOMING_RINGTONE` static property.
- Used `DEVICE_INCOMING_RINGTONE` as default value for `Sounds#incomingCallSoundResId`.
- Added getters to incoming and outgoing properties for better encapsulation and to provide fallbacks - see code.
- Exposed `Uri`s from `Sounds`.
- Used new `Sounds` properties in `CallService`.

### 🧪 Testing

Cases:
- Default `Sounds` in `Builder`: outgoing - default tone provided by SDK, incoming - device ringtone.
- Pass custom sound res to `Sounds` property: the custom sound will be used.
- Pass `null` to `Sounds` property: that sound will not exist (muted).
- Device ringtone returns `null`: fallback to default SDK incoming sound.
- Outgoing sound was set to `DEVICE_INCOMING_RINGTONE`: fallback to default SDK outgoing sound.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/Sounds.kt
```diff
@@ -16,16 +16,52 @@
...
+class Sounds(
+    private val context: Context,
+    @RawRes private val incomingCallSoundResId: Int? = DEVICE_INCOMING_RINGTONE,
+    @RawRes private val outgoingCallSoundResId: Int? = R.raw.call_outgoing_sound,
+) {
+    private val logger by taggedLogger("StreamVideo:Sounds")
+
+    internal val incomingCallSoundUri: Uri?
+        get() = if (incomingCallSoundResId == DEVICE_INCOMING_RINGTONE) {
+            RingtoneManager.getActualDefaultRingtoneUri(context, RingtoneManager.TYPE_RINGTONE) ?: run {
```
aleksandar-apostolov: wrap this in `safeCall` or `runCatching` just in case we are failing the loading of the default ringtone.

aleksandar-apostolov: Actually, why are we parsing this every time we `get` it? 
I think its better to do an `init` block or inside the constructor to initialise these sounds.
Or at least set the `field`.

I would rather add a `SoundsConfig` class instead of the `Context` parameter and then be flexible enough on what to add there, including the context. Furthermore we can add a boolean instead of relying on the `DEVICE_INCOMING_RINGTONE`

Why not do the following:
1. `@Deprecate` the current `Sounds` constructor. Add a version to remove.
2. Create a new constructor with a `SoundsConfig` class.
3. The default `SoundsConfig` is to use our default sounds.
4. Add two methods to provide these configuration and sounds.

e.g. 
`streamDefaultSounds()`
`deviceRingtoneSounds(context)` 

The definition of these functions can be the following:
```kotlin
fun defaultCallSounds(
   incommingCallSound: Int = R.raw.call_incomming_sound,
   outgoingCallSound: Int = R.raw.call_outgoing_sound)
) = Sounds(ResSoundConfig(incommingCallSound, outgoingCallSound))
```

and the other one

```kotlin
fun deviceRingtoneSounds(
   context: Context,
   outgoingCallSound: Int = R.raw.call_outgoing_sound)
) {
      val ringtoneUri = RingtoneManager.getActualDefaultRingtoneUri(context, RingtoneManager.TYPE_RINGTONE)
      // Since the context was used up we are not leaking it in the Sounds object anymore
      Sounds(UriSoundConfig(ringtoneUri, defaultOutgoingSoundUri)
}
```

E.g. 
```kotlin
interface SoundConfig {
   fun incommingSoundUri() : Uri?
   fun outhoingSoundUri(): Uri?
}

ResSoundConfig(
    incommingCallSound: Int = R.raw.call_incomming_sound,
    outgoingCallSound: Int = R.raw.call_outgoing_sound)
) : SoundConfig { 
 // The usual implementation with cached values for the URI.
}

UriSoundConfig(
     incommingSoundUri: Uri,
     outgoingSoundUri: Uri
) {
// Just expose via the interface.
}
```

This way we expose I think an unlimited number of possibilities with configuring the sounds, while having two easy default options.
The class `Sounds` is also questionable, but lets keep it for now.

WDYT?

aleksandar-apostolov (changes requested):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/Sounds.kt
liviu-timar: > wrap this in safeCall or runCatching just in case we are failing the loading of the default ringtone.

The docs say that method returns `null` if the loading fails, e.g. if the ringtone is on external storage. In that case, we fall back to the default sound.

> Actually, why are we parsing this every time we get it?

That was because the ringtone can change since the last time the SDK was initialized.

> Since the context was used up we are not leaking it in the Sounds object

It uses the `context` that's already passed to the builder and which is stored anyway in `StreamVideoImpl`.

> This way we expose I think an unlimited number of possibilities with configuring the sounds, while having two easy default options. WDYT?

Overall looks like a good idea! We could use this opportunity and make it more flexible. 👍Maybe small changes like: the default should use the device ringtone etc. I'll start looking into it. ⚙️

aleksandar-apostolov: Also this `getActualDefaultRingtoneUri` are we sure we can always read it?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -490,13 +490,13 @@ internal open class CallService : Service() {
...
-    private fun playCallSound(@RawRes sound: Int?) {
-        sound?.let {
+    private fun playCallSound(soundUri: Uri?, mediaPlayer: MediaPlayer?) {
+        if (soundUri != null && mediaPlayer != null) {
             try {
-                mediaPlayer?.let {
-                    if (!it.isPlaying) {
-                        setMediaPlayerDataSource(it, sound)
-                        it.start()
+                with(mediaPlayer) {
```
aleksandar-apostolov: This part where we are playing the ringtone. Should we not use the `val reingtone = RingtoneManager.getRingtone(context, uri)` and then play the ringtone after getting audio focus.
I think using `MediaPlayer` here is wrong, lets change this.

We can also do this for our own `Uri`s.

```kotlin
fun requestAudioFocus(context: Context): Boolean {
    audioManager = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager

    val audioAttributes = AudioAttributes.Builder()
        .setUsage(AudioAttributes.USAGE_NOTIFICATION_RINGTONE)
        .setContentType(AudioAttributes.CONTENT_TYPE_MUSIC)
        .build()

    focusRequest = AudioFocusRequest.Builder(AudioManager.AUDIOFOCUS_GAIN_TRANSIENT)
        .setAudioAttributes(audioAttributes)
        .setAcceptsDelayedFocusGain(false)
        .setOnAudioFocusChangeListener { /* Handle focus change if needed */ }
        .build()

    val result = audioManager.requestAudioFocus(focusRequest)
    return result == AudioManager.AUDIOFOCUS_REQUEST_GRANTED
}

// Then
fun playRingtone(context: Context) {
    if (requestAudioFocus(context)) {
        val ringtone = getDefaultRingtone(context)
        ringtone.play()
    } else {
        // Handle failure to gain audio focus
    }
}
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/Sounds.kt
liviu-timar: If we can't read it (external storage & no permission etc), then it returns `null` and we fall back to the default incoming sound. Also I've wrapped it in `safeCall` and returned `null` if exception occurs.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
liviu-timar: Good find, I'll look into it. What I know:
- Using `RingtoneManager` and it's `repeatSound` method requires a Android version bigger than our `minSDK` so for now I just kept it simple.
- It seems that `RingtoneManager` uses `MediaPlayer` under the hood.
- Maybe `requestAudioFocus` helps with reported issues about not playing sound on some devices.

I'll test the volume controls again and look into it. Then decide if/what to leave for a future PR, like we said.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/Sounds.kt
liviu-timar: Refactored the code. Docs for overview [here](https://staging.getstream.io/video/docs/android/advanced/ringing/#sounds).

aleksandar-apostolov (approved):

aleksandar-apostolov (approved):

PR #1181, opened 2024-09-17
Remove debug info

aleksandar-apostolov:
### 🎯 Goal

The `DebugInfo` has been making more trouble than it was worth having. Removing it for now. We will create a new `Profiler` class in the future to allow us having debugging and timing info in the future.

### 🛠 Implementation details

Removed the `Timer` and `DebugInfo` classes. Also removed `developmentMode` boolean in `StreamVideoClientIml`.


skydoves (approved):
Looks good to me :) I like this cleaning

PR #1183, opened 2024-09-18
Improve notifications

liviu-timar:
### 🎯 Goal

Refactor notifications to show up-to-date text and icons.

### 🛠 Implementation details

- Added new `getOngoingCallNotification` method in `NotificationHandler` with an `onUpdate` lambda parameter.
- The default implementation of `getOngoingCallNotification` observes the ringing state and number of members/participants and calls `onUpdate` with a refreshed notification that reflects the call state.
- `getOngoingCallNotification` is used in `CallService`. Will also be used in the `Telecom` implementation in the future.
- Various changes to `DefaultNotificationHandler` and `CallService`
- Changed notification titles, desc, icons.
- ❗ Breaking change: the `callDisplayName` parameter in `NotificationHandler#getRingingCallNotification()` is now nullable (`String` -> `String?`).

### 🎨 UI Changes

image

### 🧪 Testing

- Go through the entire ringing flow and watch all notifications: incoming, outgoing, ongoing.
- Start a group call and add/remove participants: incoming, outgoing, ongoing.


liviu-timar:
Found a bug. Pressing on the ongoing call `Hang up` button does not work on the caller device, only on receiver.

liviu-timar (commented):
Found a bug. Pressing on the ongoing call `Hang Up` button does not work on the caller device, only on receiver. Fixing it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/DefaultNotificationHandler.kt
```diff
@@ -332,7 +300,11 @@ public open class DefaultNotificationHandler(
             callId,
             notificationId,
         )
-        val endCallIntent = intentResolver.searchEndCallPendingIntent(callId = callId)
+        val hangUpIntent = if (isOutgoingCall) {
```
liviu-timar, line 310: Bug here, hang-up not working for outgoing device. Looking into it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/NotificationHandler.kt
```diff
@@ -18,22 +18,36 @@ package io.getstream.video.android.core.notifications
...
+        callId: StreamCallId,
+        callDisplayName: String?,
+        isOutgoingCall: Boolean = false,
+        remoteParticipantCount: Int = 0,
+    ): Notification?
     fun getRingingCallNotification(
         ringingState: RingingState,
         callId: StreamCallId,
-        callDisplayName: String,
+        callDisplayName: String?,
```
aleksandar-apostolov: can we have a default `null` value?

liviu-timar: Yes, would be good. Also did it for `getOngoingCallNotification`.

aleksandar-apostolov (approved):

PR #1184, opened 2024-09-18
Improvements in `StreamCallActivity` and `demo-app`

aleksandar-apostolov:
### 🎯 Goal

1. Improve the `StreamCallActivity` to be able to better handle `onConfigurationChange` without destroying the call or entering PiP. 
2. Add a wake-lock to hold the screen on if this is configured in the intent.
3. Add additional controls in the landscape menu to allow for chat access and settings.


demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -556,6 +569,18 @@ fun CallScreen(
...
+    val context = LocalContext.current
+    return remember {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            (context as? Activity)?.isInPictureInPictureMode ?: false
+        } else {
+            false
+        }
+    }
+}
+
```
skydoves: What about improving this method following [the official docs](https://developer.android.com/develop/ui/compose/system/picture-in-picture#handle-ui) like the code below?

```kotlin
internal fun Context.findActivity(): ComponentActivity {
    var context = this
    while (context is ContextWrapper) {
        if (context is ComponentActivity) return context
        context = context.baseContext
    }
    throw IllegalStateException("Picture in picture should be called in the context of an Activity")
}

@Composable
fun rememberIsInPipMode(): Boolean {
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
        val activity = LocalContext.current.findActivity()
        var pipMode by remember { mutableStateOf(activity.isInPictureInPictureMode) }
        DisposableEffect(activity) {
            val observer = Consumer<PictureInPictureModeChangedInfo> { info ->
                pipMode = info.isInPictureInPictureMode
            }
            activity.addOnPictureInPictureModeChangedListener(
                observer
            )
            onDispose { activity.removeOnPictureInPictureModeChangedListener(observer) }
        }
        return pipMode
    } else {
        return false
    }
}
```

This is out of topic of this PR, we can also apply the same approach [for our SDK code](https://github.com/GetStream/stream-video-android/blob/7080ea26e05ab4490c5c68f68d71a2b2aedcba80/stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/pip/PictureInPicture.kt#L61) as well.

skydoves (approved):
Overall looks good to me!

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
aleksandar-apostolov: I added the official `rememberIsInPipMode()` to the SDK and we are now using it in our components as well apart from `MediaPiPLifecycle` since that one uses `java` callback code.

PR #1185, opened 2024-09-19
Improve which participant is shown in PiP, allow for mirroring the `me` video stream for better experience.

aleksandar-apostolov:
### 🎯 Goal

The goal of this PR is to improve how PiP participant is determined and also improve the video rotation for `me` participant.

### 🛠 Implementation details

1. PiP participant rules
- If we have two participants only, always take the not-`me` one.
- If we have a list of activeSpeakers find the first one who is not `me`
- If we have a dominant speaker who is not `me` show it
- If we currently have no other option, show `me`

2. VideoConfig changes
- The `VideoRenderer` now accepts a `VideoRendererConfig` which apart from the scale type and the fallback content also defines a parameter called `mirrorStream` if yes, the stream will be mirrored for a `me` participant in case the front camera is selected.

### DIFF
--- a/stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/config/VideoRendererConfig.kt
+++ b/stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/config/VideoRendererConfig.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2014-2024 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-video-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/config/VideoRendererConfig.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.ui.Modifier
+import io.getstream.video.android.compose.ui.components.video.DefaultMediaTrackFallbackContent
+import io.getstream.video.android.compose.ui.components.video.VideoScalingType
+import io.getstream.video.android.core.Call
+
+@Immutable
+public data class VideoRendererConfig(
+    var mirrorStream: Boolean = false,
+    val scalingType: VideoScalingType = VideoScalingType.SCALE_ASPECT_FILL,
+    val fallbackContent: @Composable (Call) -> Unit = {},
```
skydoves: Do we need to make `mirrorStream` as mutable? If so, we should remove the `@Immutable` annotation.

skydoves (approved):
Overall looks good to me!

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/config/VideoRendererConfig.kt
aleksandar-apostolov: It shouldn't be a `var`. I will change it to `val` and keep the annotation.

PR #1186, opened 2024-09-20
Add config to skip notification permission on `build()`

aleksandar-apostolov:
### 🎯 Goal

Allow to configure if permission request about notification is going to be made when the `StreamVideo.build()` method is called.

### 🛠 Implementation details

Add a new parameter in `NotificationConfig` called `requestPermissionOnDeviceRegistration`.
Do not call `notificationPermissionManager?.start()` if `requestPermissionOnDeviceRegistration == false`.


skydoves (approved):
Looks good to me!

PR #1192, opened 2024-09-27
Add User-Agent with the App Data

JcMinarro:
### 🎯 Goal

Include a custom User-Agent to any HTTP Call our SDK does. The User-Agent will looks like:
```
"User-Agent:Stream Video Calls (Development) / 1.1.7-DEBUG(1); StandAloneInstall; (Google; sdk_gphone64_x86_64; SDK 32; Android 12)"
```

The same we did on https://github.com/GetStream/stream-chat-android/pull/5415

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/ConnectionModule.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/ConnectionModule.kt
@@ -18,6 +18,7 @@ package io.getstream.video.android.core.internal.module
 
 import android.content.Context
 import android.net.ConnectivityManager
+import android.os.Build
 import io.getstream.video.android.core.StreamVideo
 import io.getstream.video.android.core.api.SignalServerService
 import io.getstream.video.android.core.dispatchers.DispatcherProvider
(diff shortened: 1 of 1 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/ConnectionModule.kt
```diff
@@ -288,12 +289,57 @@ internal class CoordinatorAuthInterceptor(
...
+internal class HeadersInterceptor(
+    context: Context,
+) : Interceptor {
+    private val userAgent by lazy { buildUserAgent(context) }
+
     override fun intercept(chain: Interceptor.Chain): Response {
         val request = chain.request()
             .newBuilder()
+            .addHeader("User-Agent", userAgent)
             .addHeader("X-Stream-Client", StreamVideo.buildSdkTrackingHeaders())
```
liviu-timar: The `x-stream-client` header contains similar info, maybe we can add the extra data in it?

aleksandar-apostolov: On chat and iOS its the same so its fine

aleksandar-apostolov (approved):

skydoves (approved):
Looks good to me!

JcMinarro:
Do not merge it yet, a customer is having an issue with the same implementation on Chat SDK when the app name contains strange characters

JcMinarro:
> Do not merge it yet, a customer is having an issue with the same implementation on Chat SDK when the app name contains strange characters

Ok, it is fixed on the last commit

aleksandar-apostolov:
Tests are failing on this even after merging. Needs to be checked in more details.

aleksandar-apostolov (changes requested):
We need to fix unit tests before merging this.

aleksandar-apostolov:
We are going to re-do the headers for all SDKs / platforms in Q1

PR #1200, opened 2024-10-08
[PBE-6041] Fix mic muted in background bug

liviu-timar:
### 🎯 Goal

Fix bug: microphone is muted when the app is in the background. Happens on some devices with Android 13 and 14.

### 🛠 Implementation details

- Changed `CallService` type from `PHONE_CALL` to `CAMERA` and `MICROPHONE`.
- Created new `AudioCallService` and related `audioCallServiceConfig`.
- Improved inline docs, naming and `AndroidManifest` comments.

### 🧪 Testing

- Test with default call type (camera + mic) to avoid regressions.
- Remove camera permissions and fg service type to simulate audio-only calls. App will crash when making call.
- Replace default config with new `audioCallServiceConfig` in `Builder`. App will not crash anymore.
- Test microphone while app is in background.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallServiceConfig.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallServiceConfig.kt
@@ -25,9 +25,15 @@ internal const val ANY_MARKER = "ALL_CALL_TYPES"
 
 // API
 /**
- * Configuration class for the call service.
+ * Configuration class for the call foreground service.
  * @param runCallServiceInForeground If the call service should run in the foreground.
  * @param callServicePerType A map of call service per type.
+ *
+ * @see callServiceConfig
+ * @see livestreamCallServiceConfig
+ * @see livestreamAudioCallServiceConfig
+ * @see livestreamGuestCallServiceConfig
+ * @see audioCallServiceConfig
  */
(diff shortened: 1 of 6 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallServiceConfig.kt
```diff
@@ -38,9 +44,9 @@ public data class CallServiceConfig(
 )
 
 /**
- * Return a default configuration for the call service configuration.
+ * Returns a default call foreground service configuration.
  */
-public fun callServiceConfig(): CallServiceConfig {
+public fun defaultCallServiceConfig(): CallServiceConfig {
```
aleksandar-apostolov: Does it really need to change to `default` since its a public function. Plus this returns a different `CallService` config based on the call but a name of `callCallServiceConfig` sounds strange hence the `callServiceConfig`.

Maybe a correct name would be `callTypeServiceConfig` but I would not change it since its public already.

liviu-timar: Exactly, `callCallServiceConfig` sounds strange, wanted to put it in line with the others, but agree, better not to change it. I'll revert this name.

aleksandar-apostolov (approved):

PR #1217, opened 2024-10-30
Fix call reconnection issue

liviu-timar:
### 🎯 Goal

Fix issue when call reconnects indefinitely when connectivity breaks and is restored.

### 🛠 Implementation details

- Changed `Call#handleSignalChannelDisconnect` to enable rejoin retries.
- Generating a new session id and using the client impl as a source of truth for it.

### 🧪 Testing

- Livestream
- Normal call


aleksandar-apostolov:
This is updated in the reconnect V2 PR. Closing this one.

PR #1219, opened 2024-10-31
Reconnect v2

aleksandar-apostolov:
### 🎯 Goal

Implement a reconnect v2 handling on Android.

### 🛠 Implementation details

New coordinator and sfu socket implementation.
Refactoring on the `RTCSession` and `Call` to better manage the connection state.

## Breaking changes 

• `tokenProvider` is now a class instead of lambda. To continue using the old implementation use `legacyTokenProvider` field in the builder.
• Removed `Migrating` as a `RealtimeConnection` state. Use `Reconnecting` instead.
• `sessionId` is no longer available via the StreamVideo instance instead it is available in the `Call` instance.
• Renamed `Call.reconnect` to `Call.fastReconnect()`
• Renamed `Call.switchSfu` to `Call.migrate()`


demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -331,33 +331,35 @@ fun CallScreen(
...
-                                                ),
-                                            )
-                                        },
-                                    )
-                                },
-                            )
+                            me?.let {
+                                FloatingParticipantVideo(
+                                    call = call,
+                                    participant = me!!,
```
martinmitrevski: isn't it better to use the value from the closure now, I believe it was called `it`?

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/base/DynamicMenu.kt
```diff
@@ -217,7 +218,8 @@ private fun DynamicMenuPreview() {
                 onToggleAudioFilterClick = { },
                 onRestartSubscriberIceClick = { },
                 onRestartPublisherIceClick = { },
-                onKillSfuWsClick = { },
+                onSfuRejoinClick = { },
```
martinmitrevski, line 221: what's the difference between "{ }" and "{}"? We should make them all consistent, preferably "{}".

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -220,6 +224,44 @@ public class Call(
...
+    private val listener = object : NetworkStateProvider.NetworkStateListener {
+        override suspend fun onConnected() {
+            leaveTimeoutAfterDisconnect?.cancel()
+            logger.d { "[onConnected] no args" }
+            val elapsedTimeMils = System.currentTimeMillis() - lastDisconnect
+            if (lastDisconnect > 0 && elapsedTimeMils < reconnectDeadlineMils) {
+                logger.d {
+                    "[onConnected] Reconnecting (fast) time since last disconnect is ${elapsedTimeMils / 1000} seconds. Deadline is ${reconnectDeadlineMils / 1000} seconds"
+                }
+                rejoin()
```
martinmitrevski: both if cases are calling the same method, how do you distinguish the fast one from the full one?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -433,31 +479,28 @@ public class Call(
...
-        }
-
-        monitor.start()
         client.state.setActiveCall(this)
-        startCallStatsReporting(result.value.statsOptions.reportingIntervalMs.toLong())
+        monitorSession(result.value)
+        return Success(value = session!!)
+    }
 
+    private suspend fun Call.monitorSession(result: JoinCallResponse) {
```
martinmitrevski, line 482: is the `Call.` prefix needed here?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -493,84 +536,93 @@ public class Call(
...
-            // Wait a little for clean-up
-            delay(250)
-
-            // Re-join the call
-            val result = _join()
-            if (result.isFailure) {
-                // keep trying until timeout
-                handleSignalChannelDisconnect(isRetry = true)
+    internal suspend fun rejoin() {
+        reconnectAttepmts++
```
martinmitrevski, line 620: Is this method safe? What if it's called multiple times and there's reconnection in progress - shouldn't we have some checks based on the state?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -493,84 +536,93 @@ public class Call(
...
+        monitor.stop()
+        state._connection.value = RealtimeConnection.Reconnecting
+        location?.let {
+            val joinResponse = joinRequest(location = it)
+            if (joinResponse is Success) {
+                // switch to the new SFU
+                val cred = joinResponse.value.credentials
+                val session = this.session!!
```
martinmitrevski, line 627: shouldn't we be safely unwrapping the session here?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -493,84 +536,93 @@ public class Call(
...
+                logger.i { "Rejoin SFU ${session?.sfuUrl} to ${cred.server.url}" }
+
+                this.sessionId = UUID.randomUUID().toString()
+                val (prevSessionId, subscriptionsInfo, publishingInfo) = session.currentSfuInfo()
+                val reconnectDetails = ReconnectDetails(
+                    previous_session_id = prevSessionId,
+                    strategy = WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN,
```
martinmitrevski, line 634: why is this hardcoded to rejoin? In the code path above it was also being called from fast reconnect.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -493,84 +536,93 @@ public class Call(
...
+                    },
+                )
+                this.session?.connect(reconnectDetails)
+                monitorSession(joinResponse.value)
             } else {
-                sfuSocketReconnectionTime = null
+                logger.e {
+                    "[rejoin] Failed to get a join response ${joinResponse.errorOrNull()}"
+                }
+                state._connection.value = RealtimeConnection.Reconnecting
```
martinmitrevski, line 661: Isn't the state set to reconnecting at the beginning?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -581,20 +633,25 @@ public class Call(
...
-        // mark us as reconnecting
-        val connectionState = state._connection.value
-
-        if (connectionState is RealtimeConnection.Joined || connectionState == RealtimeConnection.Connected) {
-            state._connection.value = RealtimeConnection.Reconnecting
-        }
-
-        // see if we are online before attempting to reconnect
-        val online = network.isConnected()
+    suspend fun reconnect() {
```
martinmitrevski: The names are a bit misleading - it's all reconnect. I would rename this one to `fastReconnect`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
```diff
@@ -104,37 +101,59 @@ class ClientState(client: StreamVideo) {
...
-            val call = clientImpl.call(type, id)
-        } else if (event is CallRingEvent) {
-            // get or create the call, update is handled by CallState
-            val (type, id) = event.callCid.split(":")
-            val call = clientImpl.call(type, id)
-            _ringingCall.value = call
+        when (event) {
+            is ConnectedEvent -> {
+                _connection.value = ConnectionState.Connected
+                registerPushDevice()
```
martinmitrevski: We shouldn't register the device for push when we're connected. This is customer action that should be done separately.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
```diff
@@ -104,37 +101,59 @@ class ClientState(client: StreamVideo) {
...
+            }
+
+            is CallCreatedEvent -> {
+                // what's the right thing to do here?
```
martinmitrevski, line 107: We should update the internal state of the call from the response. Here, it seems like we're not doing anything.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -93,25 +96,31 @@ public class StreamVideoBuilder @JvmOverloads constructor(
...
+    } ?: ConstantTokenProvider(token),
     private val loggingLevel: LoggingLevel = LoggingLevel(),
     private val notificationConfig: NotificationConfig = NotificationConfig(),
     private val ringNotification: ((call: Call) -> Notification?)? = null,
     private val connectionTimeoutInMs: Long = 10000,
     private var ensureSingleInstance: Boolean = true,
     private val videoDomain: String = "video.stream-io-api.com",
     private val runForegroundServiceForCalls: Boolean = true,
-    private val callServiceConfig: CallServiceConfig? = null,
+    private val callServiceConfig: CallServiceConfig? = livestreamGuestCallServiceConfig(),
```
martinmitrevski: what's the purpose of this?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -139,16 +136,16 @@ internal const val defaultAudioUsage = AudioAttributes.USAGE_VOICE_COMMUNICATION
...
+    internal val scope: CoroutineScope = ClientScope(),
     override val user: User,
     internal val apiKey: ApiKey,
     internal var token: String,
     private val lifecycle: Lifecycle,
     private val loggingLevel: LoggingLevel,
-    internal val connectionModule: ConnectionModule,
-    internal val tokenProvider: (suspend (error: Throwable?) -> String)?,
+    internal val coordinatorConnectionModule: CoordinatorConnectionModule,
+    internal val tokenProvider: TokenProvider = ConstantTokenProvider(token),
```
martinmitrevski, line 148: what is a constant token provider doing? Ideally, this should be `nil` and no-op - customers should provide it from the outside.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -205,7 +190,7 @@ internal class StreamVideoImpl internal constructor(
         // stop all running coroutines
         scope.cancel()
         // stop the socket
-        socketImpl.cleanup()
+        // socketImpl.cleanup()
```
martinmitrevski: best to remove it over commenting it out (if not needed). Although, shouldn't we do some cleanup of the socket as well?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -228,42 +213,30 @@ internal class StreamVideoImpl internal constructor(
...
+            } else {
+                throw e
             }
         }
     }
 
+    private fun HttpException.isAuthError(): Boolean {
+        val failure = parseError(this)
+        val parsedError = failure.value as Error.NetworkError
+        return parsedError.serverErrorCode == VideoErrorCode.TOKEN_EXPIRED.code
```
martinmitrevski: You should check for more codes - 40, 41, 42, not only one.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -657,19 +591,13 @@ internal class StreamVideoImpl internal constructor(
         }
     }
 
-    private suspend fun waitForConnectionId(): String? =
+    private suspend fun waitForConnectionId(): String {
```
martinmitrevski: why is it non-optional now? Do we return some default value?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -328,24 +344,20 @@ public class RtcSession internal constructor(
...
-                sfuToken,
-                getSdp,
-                sfuFastReconnectListener,
-            )
+        val sfuConnectionModule = SfuConnectionModule(
+            context = clientImpl.context,
+            apiKey = apiKey,
+            apiUrl = sfuUrl,
+            wssUrl = sfuWsUrl,
+            connectionTimeoutInMs = 2000L,
```
martinmitrevski, line 356: 2 seconds? Isn't it too low?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -368,20 +380,75 @@ public class RtcSession internal constructor(
...
+                when (sfuSocketState) {
+                    is SfuSocketState.Connected ->
+                        call.state._connection.value =
+                            RealtimeConnection.Connected
+
+                    is SfuSocketState.Connecting ->
+                        call.state._connection.value =
+                            RealtimeConnection.InProgress
+                    else -> {
+                        // Ignore it
```
martinmitrevski, line 405: shouldn't we listen to disconnected?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -420,19 +487,20 @@ public class RtcSession internal constructor(
...
-            }
-        } catch (e: TimeoutCancellationException) {
-            throw RtcException(
-                message = "RtcSession.connect() timed out. JoinCallResponseEvent not received in time.",
-                cause = e,
-            )
+        val request = JoinRequest(
+            session_id = sessionId,
+            token = sfuToken,
+            fast_reconnect = false,
```
martinmitrevski, line 497: is the method used only for initial connect? If not, we should look into this being always false.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1335,69 +1388,68 @@ public class RtcSession internal constructor(
...
+                            )
+
+                            // If successful, emit the result and break the loop
+                            result.getOrThrow()
+                            break
+                        } catch (cause: Throwable) {
+                            val sameValue = mangledSdp == publisherSdpOffer.value
+                            val sameSfu = currentSfu == sfuUrl
+                            val isPermanent = isPermanentError(cause)
+                            val willRetry =
```
martinmitrevski, line 1430: this is a bit strange logic - why do we need to do this?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1335,69 +1388,68 @@ public class RtcSession internal constructor(
...
-                        ),
-                    )
-
-                    // start listening to ICE candidates
-                    launch {
-                        sfuConnectionModule.sfuSocket.pendingPublisherIceCandidates.collect { iceCandidates ->
-                            publisher?.addIceCandidate(iceCandidates)
+                            if (willRetry) {
+                                delay(delayInMs[attempt])
+                                attempt++
```
martinmitrevski, line 1439: the loop is: `while (attempt <= maxAttempts)`, `maxAttempts` is 3, `attempt`'s initial value is 0. This means this will run 4 times, while there are 3 values in the list `delayInMs`. Will this crash on the 4th attempt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/CoordinatorAuthInterceptor.kt
```diff
@@ -0,0 +1,61 @@
...
+    var apiKey: String,
+    var token: String,
+    var authType: String = "jwt",
+) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val original = chain.request()
+
+        val updatedUrl = if (original.url.toString().contains("video")) {
```
martinmitrevski, line 36: This check is also a bit unsafe and might impact things somewhere else, can't we do it differently?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/ConnectionConf.kt
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.socket.common
+
+import io.getstream.video.android.model.SfuToken
+import io.getstream.video.android.model.User
+import stream.video.sfu.event.JoinRequest
+
+public sealed class ConnectionConf {
```
martinmitrevski, line 23: Why not use `Config` instead of `Conf`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/HealthMonitor.kt
```diff
@@ -0,0 +1,171 @@
...
+        consecutiveFailures = 0
+        postponeHealthMonitor()
+    }
+
+    /**
+     * Postpone the action to check if connection keeps alive.
+     * If the connection is not alive anymore, an action to reconnect is postponed.
+     * In another case the healthCheck is postponed.
+     */
+    private fun postponeHealthMonitor() {
```
martinmitrevski, line 93: what triggers the execution of the action, after it being postponed?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/HealthMonitor.kt
```diff
@@ -0,0 +1,171 @@
...
+    ): Job = launch {
+        delay(delayMilliseconds)
+        block()
+    }
+
+    internal fun interface RetryInterval {
+        fun nextInterval(consecutiveFailures: Int): Long
+    }
+
+    object ExponencialRetryInterval : RetryInterval {
```
martinmitrevski: *Exponential

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/parser2/adapters/CustomObjectDtoAdapter.kt
```diff
@@ -0,0 +1,116 @@
...
+import kotlin.reflect.KClass
+
+/**
+ * Base class for implementing Moshi adapters that support our API's dynamic
+ * JSON models.
+ */
+internal open class CustomObjectDtoAdapter<Value : Any>(private val kClass: KClass<Value>) {
+
+    private companion object {
+        private const val EXTRA_DATA = "extraData"
```
martinmitrevski: shouldn't this be "custom"?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/SfuSocket.kt
```diff
@@ -0,0 +1,444 @@
...
+                            is SfuSocketState.Disconnected.WebSocketEventLost -> {
+                                streamWebSocket?.close()
+                                connectionConf?.let {
+                                    sfuSocketStateService.onReconnect(
+                                        it,
+                                    )
+                                }
+                            }
+                        }
+                        // callListeners { listener -> listener.onDisconnected(cause = state.cause) }
```
martinmitrevski: can we remove this?

stream-video-android-core/src/main/proto/video/sfu/event/events.pb.go
```diff
@@ -71,7 +71,7 @@ func (x VideoLayerSetting_Priority) Number() protoreflect.EnumNumber {
 
 // Deprecated: Use VideoLayerSetting_Priority.Descriptor instead.
 func (VideoLayerSetting_Priority) EnumDescriptor() ([]byte, []int) {
-	return file_video_sfu_event_events_proto_rawDescGZIP(), []int{26, 0}
+	return file_video_sfu_event_events_proto_rawDescGZIP(), []int{29, 0}
```
martinmitrevski, line 74: why do you keep this in the repo?

martinmitrevski (commented):
Added some comments, we also need to add tests.

PR #1220, opened 2024-10-31
[PBE-6219] Make number of participants more accurate in call session (lobby)

liviu-timar:
### 🎯 Goal

We need to handle the new `call.session_participant_count_updated` event & overall improve the participant count logic.

### 🛠 Implementation details

- Handling new `CallSessionParticipantCountsUpdatedEvent` event
- Improved existing participant events to update the `byRole` map in the call session
- Updating `call.state.participantCounts` from new event in lobby state and from SFU Health Check after join.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/org/openapitools/client/models/CallSessionParticipantCountsUpdatedEvent.kt
+++ b/stream-video-android-core/src/main/kotlin/org/openapitools/client/models/CallSessionParticipantCountsUpdatedEvent.kt
@@ -23,7 +23,6 @@
 
 package org.openapitools.client.models
 
-import org.openapitools.client.models.CallParticipantResponse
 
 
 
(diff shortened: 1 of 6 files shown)


skydoves (approved):
Looks very good to me!

stream-video-android-core/src/main/kotlin/org/openapitools/client/models/CallSessionParticipantCountsUpdatedEvent.kt
```diff
@@ -0,0 +1,81 @@
...
+ * @param anonymousParticipantCount
+ * @param callCid
+ * @param createdAt
+ * @param participantsCountByRole
+ * @param sessionId Call session ID
+ * @param type The type of event: \"call.session_participant_count_updated\" in this case
+ */
+
+
+data class CallSessionParticipantCountsUpdatedEvent (
```
aleksandar-apostolov, line 50: Is this manually written or generated?

liviu-timar: It's generated.

PR #1229, opened 2024-11-20
[AND-26] Fix sound still audible when device muted issue

liviu-timar:
### 🎯 Goal

Fix issue that causes sound to still be heard even if device is muted.

### 🛠 Implementation details

- Remove `audioUsage` from `CallServiceConfig`
- Rely  only on `audioUsage` `Builder` param as it's easier to deprecate the `CallServiceConfig` constructor and audio usage isn't really related to the call service.
- `audioUsage` is used when building the `PeerConnectionFactory`.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
@@ -208,6 +208,7 @@ public class StreamVideoBuilder @JvmOverloads constructor(
             callServiceConfig = callServiceConfig
                 ?: callServiceConfig().copy(
                     runCallServiceInForeground = runForegroundServiceForCalls,
+                    audioUsage = audioUsage,
                 ),
             testSfuAddress = localSfuAddress,
             sounds = sounds,
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
@@ -208,7 +208,6 @@ public class StreamVideoBuilder @JvmOverloads constructor(
             callServiceConfig = callServiceConfig
                 ?: callServiceConfig().copy(
                     runCallServiceInForeground = runForegroundServiceForCalls,
-                    audioUsage = audioUsage,
                 ),
             testSfuAddress = localSfuAddress,
             sounds = sounds,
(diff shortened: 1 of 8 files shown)


tutorials/tutorial-livestream/src/main/kotlin/io/getstream/video/android/tutorial/livestream/LiveGuest.kt
```diff
@@ -63,6 +63,7 @@ fun LiveAudience(
         callServiceConfig = livestreamGuestCallServiceConfig(),
         ensureSingleInstance = false,
         loggingLevel = LoggingLevel(priority = Priority.VERBOSE),
+        audioUsage = AudioAttributes.USAGE_MEDIA,
```
aleksandar-apostolov: Isn't it better to just set this to `audioUsage = callServiceConfig.audioUsage` and deprecate the `audioUsage` parameter in the builder?
We always take it from the config for the peer connection as well.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallServiceConfig.kt
```diff
@@ -35,13 +35,27 @@ internal const val ANY_MARKER = "ALL_CALL_TYPES"
  * @see livestreamGuestCallServiceConfig
  * @see audioCallServiceConfig
  */
-public data class CallServiceConfig(
+
+public data class CallServiceConfig
+@Deprecated(
```
aleksandar-apostolov: Should we deprecate the builder field instead?

tutorials/tutorial-livestream/src/main/kotlin/io/getstream/video/android/tutorial/livestream/LiveGuest.kt
liviu-timar: Yes, we can do it either way. I thought of both options and came to the conclusion that it's simpler to deprecate the audioUsage prop here in a way it's visible. If I deprecate the Builder constructor parameter, you have no visual cue, you see it only if you read the inline docs for it. It would require deprecating the whole Builder constructor and adding a secondary one, which seems too much. Besides this, the Builder param is older and used more.

If you think it's more useful for the future to have audioUsage only in the callServiceConfig, I can change it. And I'll deprecate just the Builder constructor parameter, not the entire constructor.

> We always take it from the config for the peer connection as well.

Actually, the builder param is used in the peer connection, but that's easy to change.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -113,6 +113,10 @@ public class StreamVideoBuilder @JvmOverloads constructor(
     private val sounds: Sounds = defaultResourcesRingingConfig(context).toSounds(),
     private val crashOnMissingPermission: Boolean = false,
     private val permissionCheck: StreamPermissionCheck = DefaultStreamPermissionCheck(),
+    @Deprecated(
```
liviu-timar, line 116: This will not add a visual indicator (a strikethrough) when using the constructor parameter. But it will show a message in the definition popup:

image

liviu-timar: Just fyi.

Anyway, the parameter is ignored and the one from `callServiceConfig` is used.

aleksandar-apostolov (approved):

PR #1236, opened 2024-11-28
Bump WebRTC to 1.3.2

skydoves:
Bump WebRTC to 1.3.2.


aleksandar-apostolov (changes requested):
Merge will wait until we have a new patch for WebRTC

PR #1239, opened 2024-12-02
[AND-153] `ForegroundServiceStartNotAllowedException` will now show the notification instead of failing the whole process of starting a service

aleksandar-apostolov:
### 🎯 Goal

Sometimes starting a foreground service from the notification may fail on higher android versions. In this case we just show the notification and delay the service for when the UI is visible.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -164,60 +169,89 @@ internal open class CallService : Service() {
...
+                            callId,
+                            TRIGGER_INCOMING_CALL,
+                            callDisplayName,
+                            callServiceConfiguration,
+                        ),
+                    )
+                }
+                result!!
+            }.onError {
+                // Show notification
```
liviu-timar, line 206: So when the service cannot be started, we just show the incoming call notification. And the service will be eventually started with the ONGOING trigger when the call is accepted. Is that correct? And what guarantees that the fg service can be started this second time? Can't it fail for a similar reason, like: "if you start the app from a notification, but you go into background while the service is starting" or if we try to start the fg service before the app reaches the foreground?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -235,6 +235,9 @@ public class StreamVideoBuilder @JvmOverloads constructor(
             scope.launch {
                 try {
                     val result = client.connectAsync().await()
+                    if (notificationConfig.autoRegisterPushDevice) {
```
liviu-timar, line 238: I remember I moved this to the block that runs when we get the Connected event, but I can't remember why exactly. I'll check tomorrow and run a test.

liviu-timar (commented):
Added some comments and a "clean up" commit.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
aleksandar-apostolov: The ONGOING trigger is used when the UI is visible.

aleksandar-apostolov: In the worst case, we'll end up with a running call and a notification, without the fg service.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
aleksandar-apostolov: No idea why it was there tbh, I removed as it made no sense.
`registerPushDevice` should be called by the integration app. However having made the mistake on Android to have it done automatically we now need to keep this behaviour as default, but add the option to exclude it. 

We will in a later release, move this flag to `false` and make all the integrations call `createDevice` themselves or optin by setting `autoRegisterPushDevice` to `true`

aleksandar-apostolov: btw, this has nothing to do with the fgs, its just reverting a change.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
liviu-timar: Not ideal, but at least we're not crashing anymore in those edge-cases. Sounds good.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
liviu-timar: Checked this. It looks good because `connectAsync` is `await`ed and we don't have that `continuation` inside it anymore (since the new socket implementation). The problem was that `createDevice` sends the token to the backend, but the user needs to be already connected for it to work. There was a race condition between `connect` and `createDevice`. I don't remember the exact details, I can dig up a thread with Kanat, if needed.

liviu-timar (approved):

PR #1242, opened 2024-12-03
[AND 24] Implement manual quality selection

liviu-timar:
### 🎯 Goal

Add functionality so that users can choose the desired resolution for incoming video or to disable it completely.

### 🛠 Implementation details

- `Call.kt`: Added methods `setPreferredIncomingVideoResolution` and `setIncomingVideoEnabled` to manage the incoming video settings.
- `RtcSession.kt`: Introduced `TrackOverridesHandler` to handle video track overrides for participants, including resolution and visibility settings.
- `RtcSession#setVideoSubscriptions`: Applied constraints to tracks to be sent in `UpdateSubscriptionsRequest`.
- `ParticipantState#updateFromParticipantInfo`: Applied constraints to `_videoEnabled` to control the UI.
- Demo app: New sub menu in demo app call menu.

### 🎨 UI Changes

Added new `Incoming video settings` menu in Demo app call screen.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1247,6 +1248,34 @@ public class Call(
...
+        }
+    }
+
+    /**
+     * Enables/disables incoming video feed.
+     *
+     * @param enabled Whether the video feed should be enabled or disabled. Set to `null` to switch back to auto.
+     * @param sessionIds The participant session IDs to enable/disable the video feed for. If `null`, the setting will be applied to all participants.
+     */
+    fun setIncomingVideoEnabled(enabled: Boolean?, sessionIds: List<String>? = null) {
```
aleksandar-apostolov, line 1317: I would preferably call this `setIncomingVideoQualitySettings` and then supply a config object instead of having two APIs.

aleksandar-apostolov: Like QualitySettings.disabled() or QualitySettings.requireResolution(x, y) (just thinking)

liviu-timar: We eventually decided to follow [JS](https://getstream.io/video/docs/javascript/advanced/manual-video-quality-selection/) on this and keep the two APIs.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ParticipantState.kt
```diff
@@ -221,7 +221,13 @@ public data class ParticipantState(
         // _dominantSpeaker.value = participant.is_dominant_speaker. we ignore this and only handle the event
         updateAudioLevel(participant.audio_level)
         _audioEnabled.value = participant.published_tracks.contains(TrackType.TRACK_TYPE_AUDIO)
-        _videoEnabled.value = participant.published_tracks.contains(TrackType.TRACK_TYPE_VIDEO)
+
+        val hasVideoTrack = participant.published_tracks.contains(TrackType.TRACK_TYPE_VIDEO)
+        _videoEnabled.value = call.session?.trackOverridesHandler?.applyOverrides(
```
aleksandar-apostolov: Lets not update the `videoEnabled` for the participant, this is actually related to the remote participant state. We need to rely on the track overrides to hide the video if we disable it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -223,23 +222,29 @@ public class RtcSession internal constructor(
...
-        MutableStateFlow<Map<String, Map<TrackType, TrackDimensions>>>(
-            emptyMap(),
-        )
-    val trackDimensionsDebounced = trackDimensions.debounce(100)
-
     // run all calls on a supervisor job so we can easily cancel them
     private val supervisorJob = SupervisorJob()
     private val coroutineScope = CoroutineScope(clientImpl.scope.coroutineContext + supervisorJob)
 
+    internal val defaultVideoDimension = VideoDimension(1080, 2340)
```
aleksandar-apostolov, line 241: why not full HD?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/TrackOverridesHandler.kt
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    /**
+     * Updates incoming video visibility overrides.
+     *
+     * @param sessionIds List of session IDs to update. If `null`, the override will be applied to all participants.
+     * @param visible Video visibility to set. Set to `null` to switch back to auto.
+     */
+    internal fun updateOverrides(
```
aleksandar-apostolov: do we need this to be internal, the whole class is internal?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/TrackOverridesHandler.kt
```diff
@@ -0,0 +1,183 @@
...
+        if (sessionIds == null) {
+            putOrRemoveVisibilityOverride(
+                sessionId = ALL_PARTICIPANTS,
+                putVisibility = putVisibility,
+                existingDimensions = trackOverrides[ALL_PARTICIPANTS]?.dimensions,
+            )
+
+            getParticipantList().forEach {
+                val override = applyOverrides(it.sessionId, true)
+                it._videoEnabled.value = override
```
aleksandar-apostolov: We should not update the participant state here.
See https://github.com/GetStream/stream-video-android/pull/1242/files#r1869802104

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ParticipantState.kt
liviu-timar: This was one of the points that I raised as a red flag and, after discussing it, I'll check if the video is enabled directly in the UI.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/TrackOverridesHandler.kt
liviu-timar: This was one of the points that I raised as a red flag and, after discussing it, I'll check if the video is enabled directly in the UI.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
liviu-timar: That was the value, I just moved the property, didn't touch the actual value.

PR #1256, opened 2024-12-18
[AND-171] Refactor call service configs

liviu-timar:
### 🎯 Goal

Extend call service configs so we can set `runCallServiceInForeground` and `audioUsage` per call `type`.

👉  Intermediary step to a larger refactoring.

### 🛠 Implementation details

- Added new methods and updated existing ones in the `CallServiceConfig` class to support more flexible configurations, including the ability to update call service settings dynamically.


liviu-timar:
Marked as Draft because of bug:
> The fg service for an accepted incoming call is not starting, even if the settings say to start it. The condition in maybeStartFgService is true, so it reads the config correctly. Tt also calls startForegroundWithServiceType with the correct type. For the outgoing device it works (fg service is started after call is accepted on the other device).

aleksandar-apostolov:
We did another implementation

PR #1266, opened 2025-01-02
AND-116: SDK size reporting

andremion:
#### Show SDK size badges in the README file, e.g.:

stream-video-android-core

#### Compare sizes before and after a PR and show them in a comparison table as a PR comment, e.g.:

| SDK | Before | After | Difference | Status |
|-|-|-|-|-|
| stream-video-android-core | 2.68 MB | 2.68 MB | 0.00 MB | 🟢 |

Max tolerance is `500 KB`
Fine tolerance is `250 KB`

if the difference is less than zero, then the status is 🚀
if the difference is greater or equal to the max tolerance, then the status is 🔴
if the difference is greater or equal to the fine tolerance, then the status is 🟡
otherwise, the status is 🟢

#### Sizes are calculated for the following SDKs:

- stream-video-android-core
- stream-video-android-ui-xml
- stream-video-android-ui-compose

### 🛠 Implementation details

- There are two metrics build flavors for each SDK:
  - One is used as a baseline, and the other includes the SDK

- The size calculation is achieved by comparing a baseline APK with an APK that includes the SDK

- The comparison task is triggered on every push request

- The metrics update is triggered on every merge to develop

### 🎉 GIF

Size matters


.github/workflows/android.yml
```diff
@@ -7,6 +7,10 @@ on:
     branches:
       - '**'
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
```
andremion, line 13: Prevent multiple PR workflows from running at the same time

.github/workflows/android.yml
```diff
@@ -7,6 +7,10 @@ on:
     branches:
       - '**'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
andremion: Prevent multiple PR workflows from running at the same time

.github/actions/gradle-cache/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Gradle Cache'
```
JcMinarro: Are we planning to extract these actions to a external repository to share it across repositories (Chat, Video, Push, ...)

andremion: It is possible, yes.
Once this is merged and tested in the development workflow, we could think about that... 👍🏻

aleksandar-apostolov (approved):

PR #1267, opened 2025-01-03
Use coroutine scope instead of using runBlocking for rejoining

skydoves:
### 🎯 Goal

Use coroutine scope instead of using runBlocking for rejoining

### 🛠 Implementation details

Since `runBlocking` blocks the entire Android main thread and occupies it until the coroutine scope is completed, it can lead to ANRs or prevent the UI from rendering properly.


skydoves:
Closing - The `prepareRejoin()` function will eventually be invoked inside the worker thread from the call site; the consideration will not happen for now.

PR #1269, opened 2025-01-03
Add core desugar library for using Java 8

skydoves:
### 🎯 Goal

We still use Java 8 functions, such as `epochSecond` and `toEpochMilli`, which require a minimum SDK level of 26 for compilation. To ensure compatibility and proper functionality on devices running SDK levels below 26, we need to configure the desugar library in our project.

https://developer.android.com/studio/write/java8-support


skydoves:
Closing this because of some issues that using this on the library module: https://issuetracker.google.com/issues/259041363

PR #1281, opened 2025-01-14
AND-246: SDK size reporting changes

andremion:
### 🎯 Goal

Update the SDK baseline sizes and the README badges in the scope of PRs

### 🛠 Implementation details

A single SDK size workflow that is triggered on every push to PR and does the following:

- Update the size baseline file and the README badges
- Push changes to the PR branch
- Post a comment to the PR with a comparison table

No commits are pushed to develop

I will update the repo settings to have this workflow as a required check after this PR get merged


.github/workflows/sdk-size.yml
```diff
@@ -26,8 +29,67 @@ jobs:
...
+          git add $METRICS_FILE README.md
+          git commit -m "[skip ci] Update SDK sizes" || echo "No changes to commit"
+
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ env.BRANCH_NAME }}
 
       - name: Get current SDK sizes
```
JcMinarro: The "CURRENT_SDK_SIZES" json should be taken from develop branch, because now we are updating it on every commit on the PR.
You can access to the raw content of this file [here](https://raw.githubusercontent.com/GetStream/stream-video-android/refs/heads/develop/metrics/size.json)

andremion: You're right 🤦🏻 
I will fix that
Thanks 🙌🏻

JcMinarro (approved):

PR #1284, opened 2025-01-15
sdk-size-fix-test

andremion:


aleksandar-apostolov:
We can leave the current size reporting as is.

PR #1287, opened 2025-01-20
[AND-260] Use legacy call config when it's passed and the new registry config is not

liviu-timar:
### 🎯 Goal

Use deprecated `callServiceConfig` parameter in `Builder` as fallback.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -211,16 +197,25 @@ public class StreamVideoBuilder @JvmOverloads constructor(
...
-        callServiceConfigRegistry.createConfigRegistry {
-            register(
-                CallType.AnyMarker.name,
-                DefaultCallConfigurations.default
-                    .copy(
-                        runCallServiceInForeground = runForegroundServiceForCalls,
-                        audioUsage = defaultAudioUsage,
-                    ),
-            )
+        // Set call configuration
```
aleksandar-apostolov, line 200: We can easily validate this behaviour with a unit tests

aleksandar-apostolov (approved):

PR #1290, opened 2025-01-22
[AND-209] Update stream-android-push to v1.3.0.

VelikovPetar:
### 🎯 Goal
Update the `stream-android-push` library to `v1.3.0`.

### 🛠 Implementation details
- Update `stream-android-push` to `v1.3.0`
- Change `PushDelegateProvider` declaration to `AndroidPushDelegateProvider` in `AndroidManifest`
- Update `VideoPushDelegate` implementation to match what is expected by `AndroidPushDelegateProvider`
- Update `StreamNotificationManager` implementation to match `PushDelegate` changes
- Update creation of `FirebasePushDeviceGenerator` in sample apps - pass `Context` via constructor
- Update `consumer-proguard-rules.pro` with `AndroidPushDelegateProvider ` instead of `PushDelegateProvider `


settings.gradle.kts
```diff
@@ -16,6 +16,7 @@ dependencyResolutionManagement {
         google()
         mavenCentral()
         maven(url = "https://plugins.gradle.org/m2/")
+        maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
```
JcMinarro: Maybe limit it to only allow Stream artifacts from this snapthots repository? :thinking: 
```suggestion
        maven(url = "https://oss.sonatype.org/content/repositories/snapshots/") {
			content {
				includeGroup("io.getstream")
			}
		}
```

skydoves (approved):
Looks good to me!

settings.gradle.kts
skydoves: We should use snapshots only for our snapshot/alpha version. If not, customers also need to add this repository on their Maven Central configuration.

liviu-timar (approved):

settings.gradle.kts
JcMinarro: Yep, it is only for our internal development when we need a pre-release version to complete a task

VelikovPetar: I will update this, even though it's probably just for a very short time, lets be safe.

aleksandar-apostolov (approved):

PR #1310, opened 2025-03-07
[AND-376] Add QR code to call welcome dialog

liviu-timar:
### 🎯 Goal

At the moment, it’s not possible to easily connect Android and iOS device on a call. You need to get the QR code from the web app, or share the link in some way. We can simplify this by showing a QR code in the `Welcome` dialog when you're alone in the call.

### 🛠 Implementation details

- Used `google.xzing.core` to generate the QR
- Displayed it in the Welcome dialog.

### 🎨 UI Changes

image


stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/base/StreamButton.kt
```diff
@@ -120,8 +122,10 @@ public fun StreamButton(
     }
     val textStyle = style.textStyle.of(state = state)
     Text(
-        style = textStyle.value.platform,
         text = text,
+        style = textStyle.value.platform,
+        maxLines = 1,
```
aleksandar-apostolov: Breaking change, we don't want this behaviour for all the buttons.

liviu-timar: I added them as parameters, with the default value that `Text` has for `overflow` and `1` for `maxLines`.

aleksandar-apostolov (approved):

PR #1317, opened 2025-03-12
[AND-393] Revert builder & ringing call UI component changes

liviu-timar:
### 🎯 Goal

- Revert removed `StreamVideoBuilder` params.
- Revert/add `@Deprecated` versions of UI composables for ringing calls.
- Revert & deprecate `canSkiPermissionRationale` param name typo in `StreamCallActivityConfiguration`.


demo-app/src/main/kotlin/io/getstream/video/android/CallActivity.kt
```diff
@@ -125,6 +127,18 @@ class CallActivity : ComposeStreamCallActivity() {
             }
         }
 
+        @Composable
+        override fun StreamCallActivity.AudioCallContent(call: Call) {
```
liviu-timar, line 131: Should we use the [new](https://github.com/GetStream/stream-video-android/pull/1317/files#diff-cae015ffa631b05c9bcd40c87148a1398e71817b091438605b8e2386fd5ccf07R163) `AudioOnlyContent` by default in `StreamCallActivityComposeDelegate`? It shows the actual participants, instead of all the members.

liviu-timar: Or should implementors customize it, like we do here in the `demo-app`?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/AudioCallContent.kt
```diff
@@ -69,7 +160,7 @@ import io.getstream.video.android.mock.previewCall
  * @param onBackPressed Handler used when the user taps on the back button.
  */
 @Composable
-public fun AudioCallContent(
+public fun AudioOnlyCallContent(
```
liviu-timar, line 163: I created a component with a different name, because the changed parameter has a default value, so some current usages would default to the new component if I keep the same name.

Changed parameter:
`detailsContent: (@Composable ColumnScope.(participants: List<MemberState>, topPadding: Dp) -> Unit)? = null,`
vs
`detailsContent: (@Composable ColumnScope.(remoteParticipants: List<ParticipantState>, topPadding: Dp) -> Unit)?`
to be able to show actual participants that are in the call, not all the members.

demo-app/src/main/kotlin/io/getstream/video/android/CallActivity.kt
aleksandar-apostolov: Implementations should customize it, changing behaviour is breaking.

aleksandar-apostolov (approved):

PR #1322, opened 2025-03-17
Integrate Telecom into the Stream Video SDK

aleksandar-apostolov:
### 🎯 Goal

Introduce support for `Telecom` integration in the Android SDK.

### 🛠 Implementation details

Implements integration as `ConnectionService` with separate `Connection` object managed by the SDK.

### 🧪 Testing

A phone with SIM is required to do testing because apart from regular ringing scenarios also tests needs to be made to verify ringing scenarios while:
1. Device is already in a SIM call
2. SIM call comes in while the user is in a Stream call.
3. Integration with wearables (e.g. watch and routing of audio)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/telecom/TelecomCallService.kt
```diff
@@ -0,0 +1,569 @@
...
+                        stopCallSound()
+                    }
+
+                    else -> {}
+                }
+            }
+        }
+    }
+
+    private fun playCallSound(soundUri: Uri?) {
```
rahul-lohra: Do we have any plan to refactor duplicate function like this one and other one `requestAudioFocus` which are also present `CallService` ?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/telecom/TelecomUtils.kt
```diff
@@ -0,0 +1,96 @@
...
+ *    https://github.com/GetStream/stream-video-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.notifications.internal.service.telecom
```
rahul-lohra, line 17: I think we can make some methods internal if they are not part of public apis

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/telecom/TelecomUtils.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    val phoneAccount = PhoneAccount.builder(
+        phoneAccountHandle,
+        "Stream calls", // Visible to the user in the system settings
+    )
+        // Capabilities define what kind of calls this account supports
+        .setCapabilities(capabilities)
+        // Optionally set an icon (for display in system dialer)
+        .setIcon(Icon.createWithResource(context, R.drawable.stream_video_ic_call))
+        .setShortDescription("My VoIP calls")
```
rahul-lohra, line 102: I’m concerned about hardcoding the label, icon, and description for the PhoneAccount.

If two or more apps using our SDK are installed on the same device, the system settings would display identical names and icons for each app's calling account.

This could confuse users, as they wouldn’t be able to distinguish between the accounts

Wdyt?

aleksandar-apostolov: Good catch!

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/telecom/TelecomCallService.kt
aleksandar-apostolov: Maybe we'll extract a `CallController` of a sort. Do you think its worthwhile to do it now or later? (both "service" classes are internal)

rahul-lohra: I think its better to do it now, otherwise it will become a tech debt

liviu-timar: The sound related functions were extracted in a `CallSoundPlayer` on the `telecom-v2` branch. I'll add it on the list to do the same here.

aleksandar-apostolov:
Close this one handled in #1495

PR #1331, opened 2025-03-21
[AND-414] Improve demo-app ringing use-case

liviu-timar:
### 🎯 Goal

We need to request permissions for incoming/outgoing calls also. Currently, permissions are requested only when you join a normal call.

### 🎨 UI Changes

Will add videos.

| Before | After |
| --- | --- |
| img | img |


stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/permission/LaunchPermissionRequest.kt
```diff
@@ -124,7 +127,12 @@ public interface LaunchPermissionRequestScope {
      * None of the permissions were granted.
      */
     @Composable
-    public fun NoneGranted(content: @Composable (showRationale: Boolean) -> Unit)
+    public fun NoneGranted(
+        content: @Composable (
+            notGranted: List<String>,
```
liviu-timar: This is a breaking change, in theory. I want to know what permissions are not granted, to improve the rationale message. Currently it only shows `microphone`.

I can find a way to avoid it, by creating another interface and inheriting this one or something like that. Or I can just keep it simple and show a generic message and not add a new parameter here.

Wdyt?

aleksandar-apostolov (approved):

aleksandar-apostolov (approved):

PR #1334, opened 2025-03-24
Add screen sharing fallback and avatar default in SDK components

liviu-timar:
### 🎯 Goal

We need to show a fallback content for the screensharing renderer while the track is loading or is not available. Now we show a `track failed to load` error.

### 🛠 Implementation details

- Added a new fallback parameter, starting from `ParticipantsLayout` and down the tree. `ParticipantsLayout` is used in `CallContent` and can be customized in the `videoContent` parameter.
- The default for the fallback content is the user avatar.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screen_recording_20250409_104934 | Screen_recording_20250409_104518 |


aleksandar-apostolov (approved):

aleksandar-apostolov:
Why is this still a draft?

PR #1339, opened 2025-03-28
[AND-429] Read call ID from launch intent extra

liviu-timar:
### 🎯 Goal

- Read call ID from launch intent extra. Useful for testing scenarios.

### 🛠 Implementation details

 - Reading the call ID from the extra with key `io.getstream.video.android.demoapp.CALL_ID`:
```kotlin
internal const val EXTRA_CALL_ID = "io.getstream.video.android.demoapp.CALL_ID"

val intent = Intent(composeTestRule.activity, MainActivity::class.java).apply {
    putExtra(EXTRA_CALL_ID, testCallId)
}
```

Future improvements:
- Sanitize the call ID
- Keep the call ID in the ViewModel
- Preserve the launch intent call ID when ending a call

### 🧪 Testing

- Use the above intent to launch the app
- Use this launch flag in terminal/Android Studio: `-e io.getstream.video.android.demoapp.CALL_ID test-call-1234`


testableapple (commented):
LGTM! 👍

Just a quick question, do we need these dependencies in DemoApp and these sample tests? I have a local branch with the testing infra setup that will conflict with these changes.

testableapple:
@liviu-timar, could we also add an option to pass the user id/name via intent?

liviu-timar:
> LGTM! 👍
> 
> Just a quick question, do we need these dependencies in DemoApp and these sample tests? I have a local branch with the testing infra setup that will conflict with these changes.

@testableapple You don't need the changes made in `build.gradle`, the `androidTest` folder and `library.versions.toml`. Try it out and let me know how it works.

Maybe it's best to move them to a different PR.

aleksandar-apostolov (approved):

testableapple:
Hey @liviu-timar, just tested the new callId argument, works like a charm!

QQ, could we implement only the changes under `demo-app/src/main`? I believe the other changes are excessive and will conflict with my work on https://github.com/GetStream/stream-video-android/pull/1349. WDYT?

testableapple (approved):
LGTM 🙌

liviu-timar:
> Hey @liviu-timar, just tested the new callId argument, works like a charm!
> 
> QQ, could we implement only the changes under `demo-app/src/main`? I believe the other changes are excessive and will conflict with my work on #1349. WDYT?

Done, can you do a quick test again?

PR #1353, opened 2025-04-09
[AND-433] Add default values to the collection properties generated by Open API

rahul-lohra:
### 🎯 Goal

[AND-433] Add default values to the collection properties generated by Open API

OpenAPI generated models have fields which are mandatory, but sometimes don't arrive from the backend.
e.g. custom field in `OwnUserResponse`.

The signature of the field is `val custom: Map<String, Any?>`
We should generate these fields with a default value of empty map to prevent the parsers from failing if the field is missing. We already use relaxed parsing on the Moshi parsers, however they cannot handle everything.

The best approach for us is to generate the models to include default values for collections:
e.g.
`val custom: Map<String, Any?> = emptyMap()`

For optional fields the handling should remain the same.


scripts/coverage.gradle
```diff
@@ -23,6 +23,7 @@ if (!rootProject.ext.sonar.ignoreModules.contains(name)) {
         currentProject {
             instrumentation {
                 excludedClasses.addAll(
+                        "io.getstream.android.video.generated.models.*",
```
testableapple: Won't the next line cover this path as well?

rahul-lohra: Yes you are right, I will remove this line.

PR #1363, opened 2025-04-23
[AND-446] Allow access to `Call` objects directly from `QueriedCalls`

aleksandar-apostolov:
### 🎯 Goal
`queryCalls` now correctly exposes access to watchable `Call` objects.

### 🛠 Implementation details
A new `queriedCalls` map is added which handles calls that are queried with the `watch = true` flag. These calls are then returned via the `QueriedCalls` response under the field `watchedCalls` these calls can then be observed and will receive events from the client and their `CallState` will be updated.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -860,6 +872,45 @@ internal class StreamVideoClient internal constructor(
         }
     }
 
+    private suspend fun rewatchCalls(
+        calls: List<Call>,
+    ): Result<Unit> {
+        // These are internal SDK defaults
+        val filter = Filters.`in`("cid", calls.map { it.cid }).toMap()
```
martinmitrevski, line 895: why do we have hardcoded filters? What if customers had different ones?

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/CallStateTest.kt
```diff
@@ -200,22 +201,99 @@ class CallStateTest : IntegrationTestBase() {
...
-                val call = clientImpl.call(it.call.type, it.call.id)
+                val call = clientImpl.queriedCall(it.call.type, it.call.id)
+                assertThat(call.state.settings.value).isNull()
+            }
+        }
+    }
+
+    @Test
+    fun `Querying calls with watch true should populate the state`() = runTest {
+//        val createResult = client.call("default", randomUUID()).create(custom=mapOf("color" to "green"))
```
martinmitrevski: can we remove this?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -540,6 +551,7 @@ internal class StreamVideoClient internal constructor(
                 it.session?.handleEvent(event)
                 it.handleEvent(event)
             }
+            queriedCalls[selectedCid]?.state?.handleEvent(event)
```
martinmitrevski: this is updating the watched call based on the event, right?

martinmitrevski (commented):
Looks good in general, mostly we should re-check the hardcoded filter. You should probably store the original query in order to rewatch it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
aleksandar-apostolov: Yes, same as with the calls map, here we also need to propagate the events to the watched calls from queryCalls

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
aleksandar-apostolov: This filter is for our `calls` map, to preserve the previous behaviour of handling a coordinator disconnect. The
`queriedCalls` map will be pre-filled based on the filter the integration used. If multiple calls to `queryCalls` were made the `queryCalls` map will contain all the returned calls from multiple `queryCalls` with different filter, we simply then rewatch those calls.

This maybe wrong thou. Because the list can change based on the filter.

PR #1365, opened 2025-04-24
[AND-481] Device selection issues

liviu-timar:
### 🎯 Goal

Select speaker for video calls and earpiece for audio calls (but based on dashboard settings).

### 🛠 Implementation details

- Removed `enforceSetup` for the `select` method, as this method receives a `device` to select, so it means `setup` was already done.
- This solves the issue where there were nested `enforceSetup` calls with passed-in callbacks and the first lambda was replaced by the second and didn't run until the end, causing the device not being selected in some cases.


rahul-lohra (approved):
LGTM

PR #1367, opened 2025-04-24
[AND-482] Add external audio device monitoring and select if available

liviu-timar:
### 🎯 Goal

Select external audio device when available (Bluetooth Headset, Wired Headset etc)

### 🛠 Implementation details

Add a new monitoring method. It will respect what AudioSwitch tells us: select external device whenever available.


rahul-lohra (approved):
lgtm

PR #1369, opened 2025-04-25
[AND-484] Add the possibility to play incoming/outgoing call sounds when device is on silent

liviu-timar:
### 🎯 Goal

Currently, incoming and outgoing call sounds are not played if the device is on silent/muted. The goal is to make it possible via configuration.

### 🛠 Implementation details

- Encapsulated call sounds in `CallSoundPlayer`
- Use specific audio attributes depending if should play sound when muted or not
- Add flags in `RingingConfig` to be used in `StreamVideoBuilder(sounds)`

### 🧪 Testing

- Tested on Samsung phone and Pixel emulator
- Tested the flags for each type of sound
- Tested in all states of phone (silent, vibrate, sound)

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/RingingConfig.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/RingingConfig.kt
@@ -37,6 +37,8 @@ import io.getstream.video.android.core.utils.safeCallWithDefault
 public interface RingingConfig {
     val incomingCallSoundUri: Uri?
     val outgoingCallSoundUri: Uri?
+    val playIncomingSoundIfMuted: Boolean
+    val playOutgoingSoundIfMuted: Boolean
 }
 
 /**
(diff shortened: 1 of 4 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/RingingConfig.kt
```diff
@@ -83,21 +114,25 @@ public fun deviceRingtoneRingingConfig(context: Context): RingingConfig = object
...
+    override val incomingCallSoundUri: Uri? = incomingCallSoundResId.toUriOrNull(context)
+    override val outgoingCallSoundUri: Uri? = outgoingCallSoundResId.toUriOrNull(context)
 }
 
 /**
  * Returns a ringing config that uses custom URIs for incoming and outgoing call sounds.
  *
  * @param incomingCallSoundUri The URI for the incoming call sound.
  * @param outgoingCallSoundUri The URI for the outgoing call sound.
+ * @param playIncomingSoundIfMuted Whether to play the incoming sound even if the device is muted.
```
rahul-lohra: I think these 2 comments `playIncomingSoundIfMuted` `playOutgoingSoundIfMuted` should be removed

rahul-lohra: @liviu-timar

rahul-lohra (approved):
lgtm

PR #1373, opened 2025-04-28
Fix issue when caller cancels a group call

liviu-timar:
### 🎯 Goal

Fix reported issue: if the caller hangs up the call before anyone answers. Other Android users are stuck in the call indefinitely.

### 🛠 Implementation details

Added check for creator reject or cancel reason.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
@@ -1040,7 +1040,9 @@ public class CallState(
         val state: RingingState = if (hasActiveCall) {
             cancelTimeout()
             RingingState.Active
-        } else if (rejectedBy.isNotEmpty() && rejectedBy.size >= outgoingMembersCount) {
+        } else if (rejectedBy.isNotEmpty() && rejectedBy.size >= outgoingMembersCount ||
+            (rejectReason?.alias == RejectReason.Cancel.alias || rejectedBy.contains(createdBy?.id))
+        ) {
             call.leave()
             cancelTimeout()
 


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1040,7 +1040,9 @@ public class CallState(
         val state: RingingState = if (hasActiveCall) {
             cancelTimeout()
             RingingState.Active
-        } else if (rejectedBy.isNotEmpty() && rejectedBy.size >= outgoingMembersCount) {
+        } else if (rejectedBy.isNotEmpty() && rejectedBy.size >= outgoingMembersCount ||
```
liviu-timar: @aleksandar-apostolov @rahul-lohra Do you think it's enough to add this extra condition or can it cause any side-effects?

rahul-lohra: Why did we add this 
`rejectReason?.alias == RejectReason.Cancel.alias`?

rahul-lohra: I will propose to create a class as the logic is a bit complex

rahul-lohra: We must make an api call to `reject` as described [slack thread](https://getstream.slack.com/archives/C05KL0HEWLA/p1745850711283639?thread_ts=1745587764.289039&cid=C05KL0HEWLA)

It is missing in this PR

liviu-timar: yes, that api call to `reject` is a separate issue, from the list that we split between us. We need to check the side effects of adding a `reject` call here.

rahul-lohra (approved):
lgtm

PR #1379, opened 2025-04-30
[AND-489] Improve call rejection flow

rahul-lohra:
### 🎯 Goal

Make a call to reject api, when all the callees have declined the call

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
@@ -45,6 +45,7 @@ import io.getstream.video.android.core.notifications.NotificationHandler.Compani
 import io.getstream.video.android.core.notifications.NotificationHandler.Companion.INTENT_EXTRA_CALL_CID
 import io.getstream.video.android.core.notifications.NotificationHandler.Companion.INTENT_EXTRA_CALL_DISPLAY_NAME
 import io.getstream.video.android.core.notifications.internal.receivers.ToggleCameraBroadcastReceiver
+import io.getstream.video.android.core.socket.common.scope.ClientScope
 import io.getstream.video.android.core.sounds.CallSoundPlayer
 import io.getstream.video.android.core.utils.safeCallWithDefault
 import io.getstream.video.android.core.utils.safeCallWithResult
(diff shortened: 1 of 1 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -502,6 +503,7 @@ internal open class CallService : Service() {
                     }
 
                     is RingingState.RejectedByAll -> {
+                        call.reject(RejectReason.Decline)
```
liviu-timar: Shouldn't we add this to `CallState.updateRingingState`, the actual place where we set the ringing state to `RejectedByAll`. Adding it in the service will need maintaining it when we replace the service, for example, with the Telecom service. Not a big issue, just wydt?

rahul-lohra: Writing in `CallState.updateRingingState` is also a bit problematic, as this function is supposed to only update state and do nothing more

But now we do much more than that

rahul-lohra: I want to follow Single Responsibility Principle but at the same time I understand that CallService might require future refactoring

rahul-lohra: I think it is acceptable as we are observing riningState only at `CallService` in the core-module. Wdyt?

rahul-lohra: At the moment the coroutine scope of `CallState` is cancelled after the call is ended so we cannot run any async operation. I already tried it

liviu-timar: Got it, then it seems more straight-forward this way.

liviu-timar: > At the moment the coroutine scope of CallState is cancelled after the call is ended so we cannot run any async operation

The service also cancels its scope at cleanup, this is a thing that needs to be checked.

liviu-timar (approved):

PR #1380, opened 2025-04-30
Call end/reject improvements

liviu-timar:
### 🎯 Goal

See implementation details.

### 🛠 Implementation details

- Added `leave` at timeout to leave call even if part of receivers accepted.
- Added ringing call check at reject to avoid leaving call UI when initiator leaves call


aleksandar-apostolov (approved):

aleksandar-apostolov (commented):
Why are we making the `CallSoundPlayer` internal?

PR #1385, opened 2025-05-01
Improve speaker toggling

liviu-timar:
### 🎯 Goal

Now, if we have the speaker selected at join, the previous device will be set as speaker, which leads to selecting only speaker when using the ToggleSpeakerphone toggle and setSpeakerPhone method.

Also, the state of the speaker toggle is not aligned with the selected device: if you have the toggle on the UI and also a Select Device menu, when you select Earpiece, the speaker toggle is still `on`

Also, it should respond to BT and wired headsets.

### 🛠 Implementation details

- Changed the logic of how `selectedBeforeSpeaker` is set and read in `setSpeakerPhone`
- Changed the logic of `select` to disable speaker if needed.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
@@ -83,7 +83,7 @@ class SpeakerManager(
     val volume: StateFlow<Int?> = _volume
 
     /** The status of the audio */
-    private val _status = MutableStateFlow<DeviceStatus>(DeviceStatus.NotSelected)
+    internal val _status = MutableStateFlow<DeviceStatus>(DeviceStatus.NotSelected)
     val status: StateFlow<DeviceStatus> = _status
 
     /** Represents whether the speakerphone is enabled */
@@ -436,6 +436,14 @@ class MicrophoneManager(
         logger.i { "selecting device $device" }
         ifAudioHandlerInitialized { it.selectDevice(device?.toAudioDevice()) }
         _selectedDevice.value = device
+
+        if (device !is StreamAudioDevice.Speakerphone && mediaManager.speaker.isEnabled.value) {
+            mediaManager.speaker._status.value = DeviceStatus.Disabled
+        }
+
+        if (device is StreamAudioDevice.Speakerphone) {
+            mediaManager.speaker._status.value = DeviceStatus.Enabled
+        }
 
         if (device !is StreamAudioDevice.BluetoothHeadset && device !is StreamAudioDevice.WiredHeadset) {
             selectedDeviceBeforeHeadset = device
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
@@ -144,7 +144,14 @@ class SpeakerManager(
             if (enable) {
                 val speaker =
                     devices.filterIsInstance<StreamAudioDevice.Speakerphone>().firstOrNull()
-                selectedBeforeSpeaker = selectedDevice.value
+                selectedBeforeSpeaker = selectedDevice.value.takeUnless {
+                    it is StreamAudioDevice.Speakerphone
+                } ?: devices.firstOrNull {
+                    it !is StreamAudioDevice.Speakerphone
+                }
+
+                logger.d { "#deviceDebug; selectedBeforeSpeaker: $selectedBeforeSpeaker" }
+
                 _speakerPhoneEnabled.value = true
                 microphoneManager.select(speaker)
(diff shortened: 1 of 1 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
```diff
@@ -154,9 +161,13 @@ class SpeakerManager(
...
-                    defaultFallbackFromType ?: selectedBeforeSpeaker ?: devices.firstOrNull {
-                        it !is StreamAudioDevice.Speakerphone
-                    }
+                    defaultFallbackFromType
+                        ?: selectedBeforeSpeaker.takeIf {
+                            it !is StreamAudioDevice.Speakerphone && devices.contains(it)
+                        }
+                        ?: devices.firstOrNull {
+                            it !is StreamAudioDevice.Speakerphone
+                        }
```
rahul-lohra: I feel the readability of the code is poor, its too complex to understand a simple code 

Can we improve it? Wdyt?

Maybe this 

```kotlin
val isSpeaker = selectedDevice.value is StreamAudioDevice.Speakerphone
val firstNonSpeaker = devices.firstOrNull { it !is StreamAudioDevice.Speakerphone }

selectedBeforeSpeaker = if (!isSpeaker) {
    selectedDevice.value
} else {
    firstNonSpeaker
}

val fallback: StreamAudioDevice? = when {
    defaultFallbackFromType != null -> defaultFallbackFromType
    selectedBeforeSpeaker != null &&
        selectedBeforeSpeaker !is StreamAudioDevice.Speakerphone &&
        devices.contains(selectedBeforeSpeaker) -> selectedBeforeSpeaker
    else -> firstNonSpeaker
}
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
```diff
@@ -426,6 +437,14 @@ class MicrophoneManager(
         ifAudioHandlerInitialized { it.selectDevice(device?.toAudioDevice()) }
```
rahul-lohra, line 440: I think the code can be simplified to this 

```kotlin
fun select(device: StreamAudioDevice?) {
    logger.i { "selecting device $device" }

    ifAudioHandlerInitialized { it.selectDevice(device?.toAudioDevice()) }
    _selectedDevice.value = device

    when (device) {
        is StreamAudioDevice.Speakerphone -> {
            mediaManager.speaker._status.value = DeviceStatus.Enabled
        }
        else -> {
            if (mediaManager.speaker.isEnabled.value == true) {
                mediaManager.speaker._status.value = DeviceStatus.Disabled
            }
        }
    }

    val isHeadset = device is StreamAudioDevice.BluetoothHeadset || device is StreamAudioDevice.WiredHeadset
    if (!isHeadset) {
        selectedDeviceBeforeHeadset = device
    }
}
```

rahul-lohra (commented):
Code requires simplicity

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
```diff
@@ -426,6 +437,14 @@ class MicrophoneManager(
...
+        if (device !is StreamAudioDevice.Speakerphone && mediaManager.speaker.isEnabled.value == true) {
+            mediaManager.speaker._status.value = DeviceStatus.Disabled
+        }
+
+        if (device is StreamAudioDevice.Speakerphone) {
+            mediaManager.speaker._status.value = DeviceStatus.Enabled
+        }
+
         if (device !is StreamAudioDevice.BluetoothHeadset && device !is StreamAudioDevice.WiredHeadset) {
             selectedDeviceBeforeHeadset = device
```
rahul-lohra: The wording device is wrong and doesn't make sense. When we are setting for something happened in past instead we should use lastSelectedDevice 

```kotlin
selectedDeviceBeforeHeadset = device
```

instead it would be more meaningful if we write 
```kotlin
selectedDeviceBeforeHeadset = lastSelectedDevice
```

rahul-lohra: The logic for `selectedDeviceBeforeHeadset`  would be wrong in the case when the device don't have any headset or wired headset in the first place and then selectedDeviceBeforeHeadset will be initialised witth some value but instead it should not be initialised

This will blindly assign `selectedDeviceBeforeHeadset` even in cases where no headset has ever been or can ever be connected

Please cmimw

rahul-lohra: Maybe you can use `nonHeadsetFallbackDevice` instead of `selectedDeviceBeforeHeadset`

rahul-lohra (approved):
LGTM

PR #1386, opened 2025-05-01
[AND-478] Allow for video renderer to take config in order to not update its visibility

aleksandar-apostolov:
### 🎯 Goal

In certain livestream use-case depending on the UI implementation there can be a race condition in which the video renderer can wrongly announce its visibility. This "announcement" is not needed for livestream so can be safely turned off.

### 🛠 Implementation details

Add an `updateVisibility` configuration option that can be passed to the `VideoRenderer` via a `VideoRenrererConfig` object. When this flag is `false` the `call.updateVisibility` is not called as well as the `cleantTrack` thus preserving the track. See videos for example.

An integration has to supply a `VideoRendererConfig` with `updateVisibility = false` to achieve the correct behaviour.

### BEFORE
https://github.com/user-attachments/assets/df3c2ad9-2bb7-48cd-962f-6aba637b7e1b

### AFTER
https://github.com/user-attachments/assets/3e11adc5-6fe3-48ec-829f-9f4c31f07a48

### DIFF
--- a/stream-video-android-core/src/main/AndroidManifest.xml
+++ b/stream-video-android-core/src/main/AndroidManifest.xml
@@ -77,7 +77,11 @@
                 <action android:name="io.getstream.video.android.action.LEAVE_CALL" />
             </intent-filter>
         </receiver>
-        <activity android:name=".notifications.internal.DismissNotificationActivity" />
+        <activity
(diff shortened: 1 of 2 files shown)


rahul-lohra (approved):
LGTM

PR #1390, opened 2025-05-05
[AND-508] Prevent multiple calls to `leave()` and `cleanup()` which may cause a concurrency in WebRTC leading to native crashes

aleksandar-apostolov:
### 🎯 Goal

The goal of this change is to prevent concurrent calls to `leave()` and `cleanup()` in the `Call` and `RtcSession`. Multiple `leave()` or `cleanup()` calls may cause an operation on already disposed media track in which case a native crash may occur.  

### 🛠 Implementation details

Create an `AtomicUnitCall` which supports the underlying function e.g. 
```kotlin
fun leave() = atomicLeave {
 // The leave code that will be called only once per the lifecycle of the `Call` object.
}
```

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AtomicUnitCall.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AtomicUnitCall.kt
@@ -21,7 +21,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 /**
  * An atomic call that is executed only once, no matter how many consecutive calls are made.
  */
-class AtomicUnitCall {
+internal class AtomicUnitCall {
     private var called = AtomicBoolean(false)
 
     /**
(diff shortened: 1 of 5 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AtomicUnitCall.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package io.getstream.video.android.core.utils
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * An atomic call that is executed only once, no matter how many consecutive calls are made.
+ */
+internal class AtomicUnitCall {
```
liviu-timar, line 24: `Unit` refers to the fact that `block` returns `Unit` or it's used as `unit of work`? If it's `unit of work` then maybe `AtomicCall` or `RunOnce` or something similar is a better name.

aleksandar-apostolov: Unit to the fact that `block: () -> Unit`.

liviu-timar (approved):

PR #1391, opened 2025-05-05
[AND-509] Deprecate the `subscribe` method in favour of `events` flow

aleksandar-apostolov:
### 🎯 Goal

Prevent leaks of the `CallActivity` and `CallService` objects.

### 🛠 Implementation details

1. Deprecate the `call.subscribe`, also added a `call.unsubscribe` which is deprecated from the start since this method is not recommended to be used, however `unsubscribe` must be added regardless.
2. Added a `events` flow on the `Call` object which allows for observing the events as well as is safely discarded when a lifecycle scope is closed and/or the service scope is finished.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -78,7 +78,6 @@ import io.getstream.video.android.core.model.UpdateUserPermissionsData
 import io.getstream.video.android.core.model.VideoTrack
 import io.getstream.video.android.core.model.toIceServer
 import io.getstream.video.android.core.utils.RampValueUpAndDownHelper
-import io.getstream.video.android.core.utils.safeCall
 import io.getstream.video.android.core.utils.safeCallWithDefault
 import io.getstream.video.android.core.utils.toQueriedMembers
 import io.getstream.video.android.model.User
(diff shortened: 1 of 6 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1062,6 +1079,10 @@ public class Call(
                 }
             }
         }
+
+        if (!events.tryEmit(event)) {
+            logger.e { "Failed to emit event to observers: [event: $event]" }
```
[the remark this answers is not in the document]
aleksandar-apostolov: Negative, if the `event.tryEmit()` fails we probably are closing the `Call` object and doing cleanup.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -526,7 +527,7 @@ internal open class CallService : Service() {
     private fun observeCallEvents(callId: StreamCallId, streamVideo: StreamVideoClient) {
         serviceScope.launch {
             val call = streamVideo.call(callId.type, callId.id)
-            call.subscribe { event ->
+            call.events.collectLatest { event ->
```
liviu-timar: Is `collectLatest` appropriate here? I'm just thinking that cancelling the block for the previous value could lead to incomplete handling of previous events (that we're currently handling or that we'll maybe add in the future).

liviu-timar: Other than this, I approved the PR.

liviu-timar (approved):

rahul-lohra (approved):
lgtm

PR #1393, opened 2025-05-05
[AND-510] Remove early return because the resolution might be available later

aleksandar-apostolov:
### 🎯 Goal

Fix an issue where camera toggle is not working due to initial missing resolution.

### 🛠 Implementation details

Remove the early return. If the resolution is empty the track is simply empty as well. But we will for sure catch the camera toggle events.


rahul-lohra (approved):
lgtm

PR #1395, opened 2025-05-05
Add timestamp to all the snapshot versions

aleksandar-apostolov:
### 🎯 Goal

Introduce timestamp to the snapshot versions in order to not override them.


martinmitrevski (approved):

JcMinarro:
Is it really needed??
When a new snapshot version is released, it adds a new snapshot "udpate".
For example, You can see all the updates of `1.6.1-SNAPSHOT` version [here](https://oss.sonatype.org/content/repositories/snapshots/io/getstream/stream-video-android-core/1.6.1-SNAPSHOT/)

If you only use `1.6.1-SNAPSHOT` as version name, Gralde will download the last update of this snapshot, but if you want to use a concrete snapshot, you can use as version the concrete "update" you want, like for example `1.6.1-20250430.163954-1` that will be the first snapthsot that was created for `1.6.1-SNAPSHOT`

PR #1397, opened 2025-05-06
[AND-513] Stop retrying resolution selection once it becomes available

rahul-lohra:
### 🎯 Goal

Stop retrying resolution selection once it becomes available

### 🛠 Implementation details

Used

```kotlin
 if (_resolution.value != null) { return@launch }
```


rahul-lohra:
No need for this

PR #1399, opened 2025-05-06
[AND-518] Support custom notification titles and audio/video call sounds

rahul-lohra:
### 🎯 Goal

Customers have requested the ability to differentiate between audio and video call notifications through distinct titles and sounds. Currently, our SDK uses a shared notification layout and relies on the default SDK ringtone, which limits customization and leads to user confusion.

This feature introduces support for:

- Custom notification titles based on call type
- Distinct incoming/outgoing sounds for audio and video calls
- A fallback mechanism to default SDK ringing tones

### 🛠 Implementation details

This PR introduces support for customizing notification titles and per-call-type ringing sounds (incoming/outgoing) for audio and video calls.

# 🔔 Customizing Notification Titles
To customize the title shown in the incoming call notification, override the following string resources in your app’s res/values/strings.xml:

```xml 
<string name="stream_video_incoming_audio_call_notification_description">Incoming audio call</string>
<string name="stream_video_incoming_video_call_notification_description">Incoming video call</string>
```
# 🎵 Customizing Incoming/Outgoing Sounds
To provide custom sounds for audio and video calls, create a Sounds object and pass it when initializing the StreamVideo client.

## ✅ Create the Sound Object
```kotlin
val audioRingingConfig = object : RingingConfig {
    override val incomingCallSoundUri: Uri
        get() = R.raw.audio_call_ringtone.toUriOrNull(context)

    override val outgoingCallSoundUri: Uri
        get() = R.raw.audio_call_ringtone.toUriOrNull(context)
}

val videoRingingConfig = object : RingingConfig {
    override val incomingCallSoundUri: Uri
        get() = R.raw.video_call_ringtone.toUriOrNull(context)

    override val outgoingCallSoundUri: Uri
        get() = R.raw.video_call_ringtone.toUriOrNull(context)
}

val sounds = Sounds(
    audioCallRingingConfig = audioRingingConfig,
    videoCallRingingConfig = videoRingingConfig,
    ringingConfig = audioRingingConfig // fallback for unknown call types
)
```
## 🚀 Pass to the Stream Client

```kotlin
return StreamVideoBuilder(
    context = context,
    sounds = sounds,
    // other config...
).build()

```


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/RingingConfig.kt
```diff
@@ -164,3 +207,27 @@ private fun Int?.toUriOrNull(context: Context): Uri? =
             null
         }
     }
+
+internal fun getIncomingCallSoundUri(streamVideo: StreamVideoClient, callId: StreamCallId): Uri? {
+    return when (callId.type) {
+        CallType.AudioCall.name ->
```
liviu-timar, line 213: What happens if the name of the call is not `audio_call` or `default`? Should we decide based on other criteria?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/DefaultNotificationHandler.kt
```diff
@@ -136,6 +137,14 @@ public open class DefaultNotificationHandler(
             val rejectCallPendingIntent = intentResolver.searchRejectCallPendingIntent(callId)
 
             if (fullScreenPendingIntent != null && acceptCallPendingIntent != null && rejectCallPendingIntent != null) {
+                /**
+                 * Temporary workaround to associate the callId with its PendingIntent
+                 * without introducing a breaking change in the public API.
+                 *
+                 * Note: This should be refactored or removed in the next major release.
+                 */
+                NotificationsMigrationHelper.incomingCallMap[acceptCallPendingIntent] = callId
```
liviu-timar, line 146: Correct me if I'm wrong, I think you added this helper to not add a `callId` parameter to `getIncomingCallNotification`, right?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/DefaultNotificationHandler.kt
```diff
@@ -219,14 +228,19 @@ public open class DefaultNotificationHandler(
             },
         )
 
+        val callId = NotificationsMigrationHelper.incomingCallMap[acceptCallPendingIntent]
+        val contentTextResId = when (callId?.type) {
+            "audio_call" -> R.string.stream_video_incoming_audio_call_notification_description
```
liviu-timar, line 233: Same as [this](https://github.com/GetStream/stream-video-android/pull/1399#discussion_r2077349282) comment.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/DefaultNotificationHandler.kt
rahul-lohra: Correct

PR #1400, opened 2025-05-06
Fix nested MicrophoneManager.setup calls 

liviu-timar:
### 🎯 Goal

This PR resolves an issue where nested calls to `MicrophoneManager.setup` or `enforceSetup` would not execute their callbacks if the `audioHandler` lateinit var was already initialized but `setupCompleted` was still `false`. In this scenario nested setup calls would be ignored. 

This causes, for example, an issue when trying to apply the dashboard mic on/off settings in the incoming call screen (by overriding `IncomingCallContent`), because `setEnabled` and `enable` are called, which means nested `enforceSetup` calls.

The fix ensures that pending callbacks are properly handled and executed in this edge case, making nested setup/enforceSetup calls reliable regardless of the internal initialization state.

### 🛠 Implementation details

The issue by maintaining a `pendingSetupCallbacks` list and invoking all callbacks once the async setup completes, regardless of the internal state. This guarantees that nested setup callbacks are not lost and always run as expected.

> [!NOTE] 
> Another approach that I tried was to use `suspendCoroutine` to wait for the device callback from AudioSwitch to execute. In order to not make the setup method `suspend`, I'd have to use `runBlocking`, but this would cause a deadlock: runBlocking blocks the main thread, but AudioSwitchHandler posts its initialization work to the same main thread using Handler.post. Since the main thread is blocked, the posted work never runs, so the coroutine inside runBlocking never resumes.

### 🎨 UI Changes

Dashboard settings: camera `off` by default, microphone `on` by default

| Before: mic is off, despite dashboard settings | After: mic is on as per dashboard settings |
| --- | --- |
| Screen_recording_20250506_174227 | Screen_recording_20250506_172853 |

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
@@ -356,9 +356,13 @@ class MicrophoneManager(
     private val logger by taggedLogger("Media:MicrophoneManager")
 
     private lateinit var audioHandler: AudioHandler
-    private var setupCompleted: Boolean = false
     internal var audioManager: AudioManager? = null
     internal var priorStatus: DeviceStatus? = null
+
+    @Volatile
+    private var isSetupInProgress = false
+    private var setupCompleted: Boolean = false
+    private val pendingSetupCallbacks = mutableListOf<() -> Unit>()
 
     // Exposed state
     private val _status = MutableStateFlow<DeviceStatus>(DeviceStatus.NotSelected)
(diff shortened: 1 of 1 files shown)


rahul-lohra:
## About `capturedOnAudioDevicesUpdate` not being invoked:

We know that the issue with the current approach is that capturedOnAudioDevicesUpdate doesn’t get invoked unless setupCompleted is already true.
Wouldn't it make sense to assign `setupCompleted = true` before invoking `capturedOnAudioDevicesUpdate?.invoke()`?

## Current approach

```kotlin
 audioDeviceChangeListener = { devices, selected ->
   ....
    capturedOnAudioDevicesUpdate?.invoke()
   ....
    setupCompleted = true
  },
````

## Proposed approach

```kotlin
 audioDeviceChangeListener = { devices, selected ->
   ....
    setupCompleted = true // Moved this up
    capturedOnAudioDevicesUpdate?.invoke()
   ....
  },
````

Wdyt @liviu-timar ?

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt [hunk 1]
```diff
@@ -356,10 +356,14 @@ class MicrophoneManager(
...
 
     private lateinit var audioHandler: AudioHandler
-    private var setupCompleted: Boolean = false
     internal var audioManager: AudioManager? = null
     internal var priorStatus: DeviceStatus? = null
 
+    @Volatile
+    private var isSetupInProgress = false
+    private var setupCompleted: Boolean = false
+    private val pendingSetupCallbacks = mutableListOf<() -> Unit>()
```
rahul-lohra, line 365: If we decide to continue with this approach, I have a suggestion regarding the code style.

1. To make the setup state management more expressive and less error-prone, we could replace the current two-flag system with a single sealed class, 
2. Restrict access to pendingSetupCallbacks only when the setup state is pending (e.g. Loading) 
like this:

```kotlin
sealed class SetupState {
    object Initial : SetupState()
    data class Loading(
        val pendingCallbacks: MutableList<() -> Unit> = mutableListOf()
    ) : SetupState()
    object Finished : SetupState()
    data class Failed(val error: Throwable) : SetupState()
}

````

Then replace the two-flag system with a single property:
```kotlin
private var setupState: SetupState = SetupState.Initial
```
## Example use in logic
```kotlin
when (val state = setupState) {
    is SetupState.Initial -> {
        setupState = SetupState.Loading()
        // start async setup
    }

    is SetupState.Loading -> {
        state.pendingCallbacks.add { callback() }
    }

    is SetupState.Finished -> {
        callback() // call immediately
    }

    is SetupState.Failed -> {
        // maybe retry
    }
}
```

## When setup completes:

```kotlin
val callbacks = (setupState as? SetupState.Loading)?.pendingCallbacks ?: emptyList()
callbacks.forEach { it() }
setupState = SetupState.Finished
```

You can adjust the code based on the requirement

liviu-timar:
Implemented by #1402

PR #1402, opened 2025-05-07
Fix nested MicrophoneManager.setup calls - simpler solution

liviu-timar:
### 🎯 Goal

Implement a more straight-forward fix for the nested `setup` calls issue described in #1400.

### 🛠 Implementation details

- Move the `setupCompleted` assignment, basically
- Synchronize the `setup` method

> [!NOTE] 
> We'll keep #1400 in mind and come back to it if needed.


rahul-lohra (approved):
We have to improve this code later. Its okay for now

PR #1404, opened 2025-05-07
[AND-449] Fix: call controls not working on incoming and outgoing screens

liviu-timar:
### 🎯 Goal

- Apply mic & cam dashboard settings on incoming and outgoing call screens
- Add mic toggle to incoming call screen

### 🛠 Implementation details

- Applying dashboard settings in `StreamCallActivity#onCreate`.
- Incoming call screen has new microphone toggle (with default null value to not be a breaking change).

### 🎨 UI Changes

- Camera and microphone toggles will be on or off, based on dashboard settings.
- For incoming: a new microphone toggle is visible. Also added a preview with 300dp width (small phone) for the call controls to make sure the toggles don't overlap.

| Before | After |
| --- | --- |
| Screenshot_20250508_165533 | Screenshot_20250508_165934 |

### 🧪 Testing

Change cam and mic dashboard settings and check if they're applied in the incoming, outgoing and in-call screens.


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -792,6 +795,25 @@ public abstract class StreamCallActivity : ComponentActivity() {
         }
     }
 
+    /**
+     * Used to apply dashboard settings to the call.
+     * By default, it enables or disables the microphone and camera based on the settings.
+     *
+     * @param call the call
+     */
+    public open fun applyDashboardSettings(call: Call) {
```
aleksandar-apostolov, line 804: Why is this public?

liviu-timar: I made it `public open` in case integrators want to override it's behaviour.

aleksandar-apostolov (approved):

PR #1416, opened 2025-05-13
[AND-545] Re-order state update sequence to prevent ending up in the wrong state

aleksandar-apostolov:
### 🎯 Goal

Proper state updates of the members that we don't end-up in a wrong state.

### 🛠 Implementation details

Reorder the `updateFromResponse` calls in `CallState.kt` in order to avoid having empty members and ending up in the wrong state thus `leave()`-ing the call.

### DIFF
--- a/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
+++ b/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
@@ -40,6 +40,7 @@ import io.getstream.result.flatMap
 import io.getstream.result.onErrorSuspend
 import io.getstream.result.onSuccessSuspend
 import io.getstream.video.android.core.Call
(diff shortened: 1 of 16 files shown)


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -434,6 +437,13 @@ public abstract class StreamCallActivity : ComponentActivity() {
         leave(call, onSuccessFinish, onErrorFinish)
     }
 
+    override fun finish() {
```
rahul-lohra: Its hard to understand. Why are we doing this?

aleksandar-apostolov: Should not be in this PR

rahul-lohra (approved):
lgtm

PR #1422, opened 2025-05-15
Do not fully reset the `Call` if the integration decides to keep it to allow for another `join()` later

aleksandar-apostolov:
### 🎯 Goal

Do not fully reset the `Call` object to allow for a `join()` call later on and to keep the data flows going in case they are monitored after the call is left.

### 🛠 Implementation details

Implement soft vs hard leave where soft doesn't reset all the `Call` data, just returns it in a `PreJoin` state.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -759,25 +759,47 @@ public class Call(
         session?.leaveWithReason(disconnectionReason?.message ?: "user")
         session?.cleanup()
         leaveTimeoutAfterDisconnect?.cancel()
+        // Always detach from the SFU and stop ongoing jobs that *must* die.
+        session?.leaveWithReason(disconnectionReason?.message ?: "user")
+        session?.cleanup()
+        session = null
+
+        leaveTimeoutAfterDisconnect?.cancel()
         network.unsubscribe(listener)
         sfuListener?.cancel()
         sfuEvents?.cancel()
-        state._connection.value = RealtimeConnection.Disconnected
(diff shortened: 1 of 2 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -758,26 +758,48 @@ public class Call(
     private fun leave(disconnectionReason: Throwable?) = atomicLeave {
         session?.leaveWithReason(disconnectionReason?.message ?: "user")
         session?.cleanup()
+        leaveTimeoutAfterDisconnect?.cancel()
+        // Always detach from the SFU and stop ongoing jobs that *must* die.
+        session?.leaveWithReason(disconnectionReason?.message ?: "user")
+        session?.cleanup()
```
rahul-lohra: Duplicate code found, it is also present just 2 line above

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1466,6 +1466,48 @@ public class CallState(
...
+    internal fun resetAfterLeave() {
+        _connection.value = RealtimeConnection.PreJoin
+        _session.value = null // will be repopulated by the next join()
+        _ringingState.value = RingingState.Idle
+        _acceptedBy.value = emptySet()
+        _rejectedBy.value = emptySet()
+
+        participantsVisibilityMonitor?.cancel()
+        participantsVisibilityMonitor = null
+        ringingTimerJob?.cancel()
```
rahul-lohra, line 1487: I believe this approach is risky, as it might cancel API calls triggered during the ringing timeout, especially under poor network conditions. We should double-check to be sure.

aleksandar-apostolov:
Too risky, closing.

PR #1423, opened 2025-05-16
[AND-554] Keep the needed calls separately to preserve updates.

aleksandar-apostolov:
### 🎯 Goal

When a call is left it is destroyed and removed from the internal state management of the SDK client. To prevent this if a configuration requires it we are going to keep these calls

### 🛠 Implementation details

Add a `destroyedCalls` list that holds the old call objects in case subscriptions are still valid for those objects.


rahul-lohra (approved):
lgtm

PR #1445, opened 2025-06-11
Allow for the SDK to collect RTC stats if configured

aleksandar-apostolov:
### 🎯 Goal

Add RTC stats for the Peer connections and several important events based around the SFU socket and its API.

### 🛠 Implementation details

Add the RTC stats implementation as specified.

TracerFactory
1. Manages and creates the tracers by name / id. Can be enabled or disabled via the `StreamVideoBuilder`

Traces
1. Allow for generic tracing and bookkeeping of tagged data.
2. Create a trace factory that creates and manages all the tracers which can then be aggregated.
3. RtcSession holds most of the traces.
4. For the SFU we trace every API call via a Proxy on the Interface.
5. StreamPeerConnection traces almost every callback
6. Sfu traces several events

Stats report
1. Create a stats report found in the `StreamPeerConnection` capable of calculating delta on stats and extract performance stats from the RtcStatsReport

Send stats
Wire both the tracers and the delta and performance stats into the `setndStats` on the `RtcSession`.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1129,6 +1182,14 @@ public class RtcSession internal constructor(
         return subscriber?.getStats()
     }
 
+    fun List<Array<Any?>>.toJson(): String {
```
rahul-lohra: Maybe we can mark this as `private` ?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1129,6 +1182,14 @@ public class RtcSession internal constructor(
...
+        val outer = JSONArray()
+        for (inner in this) {
+            outer.put(JSONArray(inner)) // wraps each Array into its own JSONArray
+        }
+        return outer.toString() // compact JSON → use .toString(2) for pretty
+    }
+    private val parser: VideoParser = MoshiVideoParser()
```
rahul-lohra: Should we declare this variable at the top? Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/trace/TraceRecord.kt
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * A single trace item captured by [Tracer].
+ *
+ * @property tag A short identifier that categorises the trace entry.
+ * @property id  Optional identifier propagated from the owning [Tracer].
+ * @property data Arbitrary payload supplied by the caller.
+ * @property timestamp Epoch-millis when the entry was recorded.
+ */
+data class TraceRecord(
```
rahul-lohra, line 31: Do you think that we should mark classes and methods as `internal` instead of leaving as public?

rahul-lohra (commented):
Please review the comments

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/trace/TracerFactory.kt
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package io.getstream.video.android.core.trace
+
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Factory that provides [Tracer] instances.
+ */
+class TracerFactory(private var enabled: Boolean = true) {
```
rahul-lohra: Maybe we can mark some of the methods as internal, what do you think?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/stats/StatsTracer.kt
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * Collects and processes WebRTC stats for publishers and subscribers.
+ *
+ * @param pc                 current [PeerConnection]
+ * @param peerType           whether this side is PUBLISHER or SUBSCRIBER
+ * @param trackIdToTrackType mapping from WebRTC track IDs to [TrackType] (audio / video / screen-share…)
+ *
+ * @internal
+ */
+class StatsTracer(
```
rahul-lohra, line 46: Should we mark this class as `internal`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/stats/StatsTracer.kt
```diff
@@ -0,0 +1,282 @@
...
+    return result
+}
+
+/**
+ * Wrapper returned by [StatsTracer.get] that contains:
+ *  * the full WebRTC stats report for the current poll,
+ *  * a delta-compressed diff against the previous poll,
+ *  * high-level performance metrics (encode / decode).
+ */
+data class ComputedStats(
```
rahul-lohra, line 276: Should we use `@InternalStreamVideoApi` for `ComputedStats` ?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/stats/StatsTracer.kt
aleksandar-apostolov: Yes, we should.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/trace/InterfaceTracer.kt
```diff
@@ -0,0 +1,48 @@
...
+): T {
+    val clazz = target!!::class.java
+    val handler = InterfaceMethodInvocationCounter(tracer, target)
+    return Proxy.newProxyInstance(clazz.classLoader, arrayOf(T::class.java), handler) as T
+}
+
+/**
+ * Count the invocations of the methods of the target object.
+ *
+ * @param scope The scope in which the counter should run.
```
rahul-lohra, line 35: I think we forgot to remove `@param scope` from comments

rahul-lohra (approved):
lgtm

PR #1458, opened 2025-06-23
Allow for stereo to be enabled if offered in the remote SDP while also improving tracing on the `Subscriber` negotiation process

aleksandar-apostolov:
### 🎯 Goal

Allow to enable stereo if offered in the SDP.

### Implementation details

Modify the answer SDP to enable `stereo=1` if it is offered in the remote SDP offer.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -762,7 +762,7 @@ public class RtcSession internal constructor(
         supervisorJob.cancel()
     }
 
-    internal val muteState = MutableStateFlow(
+    internal var muteState = MutableStateFlow(
```
rahul-lohra: I think we can keep it `val`. Please check

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnection.kt
```diff
@@ -75,6 +75,7 @@ open class StreamPeerConnection(
     private val onNegotiationNeeded: ((StreamPeerConnection, StreamPeerType) -> Unit)?,
     private val onIceCandidate: ((IceCandidate, StreamPeerType) -> Unit)?,
     private val maxBitRate: Int,
+    private val traceCreateAnswer: Boolean = true,
```
rahul-lohra, line 81: Should we update the KDoc for the constructor of this class as we have added a new parameter?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
aleksandar-apostolov: Good catch, this is a remnant of a bugfix attempt, we'll do separate PR.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnection.kt
aleksandar-apostolov: `StreamPeerConnection` is internal class (even thou public). But nevertheless added KDoc for 3 other missing fields :)

PR #1464, opened 2025-06-26
Exclude the `sendStats` RPC call from tracing

aleksandar-apostolov:
### 🎯 Goal

Exclude the `sendStats` RPC call from tracing. It is not needed.

### 🛠 Implementation details

Implement a new decorator which explicitly traces errors and method calls instead of the generic java proxy implementation so we can control which methods are traced.


rahul-lohra (approved):
lgtm

PR #1468, opened 2025-06-27
Do not update the video subscriptions if unchanged. Throttle dimension based on subscriptions size

aleksandar-apostolov:
### 🎯 Goal
[AND-637]
Do not update the video subscriptions if unchanged. Throttle dimension based on subscriptions size

### 🛠 Implementation details

Subscriber - now compares previous subscriptions before sending the request. Also video size is throttled if there are too many videos subscribed, because default video is FHD and if the viewport does not update its size we must scale down manually otherwise we show 6 FHD videos.

### DIFF
--- a/stream-video-android-core/api/stream-video-android-core.api
+++ b/stream-video-android-core/api/stream-video-android-core.api
@@ -1,6 +1,7 @@
 public abstract interface class io/getstream/android/video/generated/apis/ProductvideoApi {
 	public abstract fun acceptCall (Ljava/lang/String;Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public abstract fun blockUser (Ljava/lang/String;Ljava/lang/String;Lio/getstream/android/video/generated/models/BlockUserRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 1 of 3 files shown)


stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/video/VideoRenderer.kt
```diff
@@ -101,6 +102,7 @@ public fun VideoRenderer(
                     Math.random().toString()
                 }
             }
+            var size by remember { mutableStateOf(IntSize.Zero) }
```
rahul-lohra: Is this variable in use? Can we verify that?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
```diff
@@ -110,14 +112,19 @@ internal class Subscriber(
          * Default video dimension.
          */
         val defaultVideoDimension = VideoDimension(720, 1280)
+        val throthledVideoDimension = VideoDimension(180, 320)
```
rahul-lohra: I think the spelling is `throthledVideoDimension` is wrong

PR #1476, opened 2025-07-09
Reconnect logic improvements, based on the connection state and/or backend API errors

aleksandar-apostolov:
### 🎯 Goal

Improve reconnect logic as per specs.

### 🛠 Implementation details

SfuSocket 
- HealthCheck and liveness set to 5s and 15s

PeerConnection 
- ICE connection failure will trigger a restartIce in 2-3 seconds if connection does not recover itsef. 
- Connection change to FAILED will trigger a rejoin()

SignalingApi 
- new error code (`PARTICIPANT_SIGNAL_LOST`) will trigger a fast reconnect


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1332,6 +1351,7 @@ public class RtcSession internal constructor(
     }
 
     private suspend fun parseError(e: Throwable): Failure {
+        Error.NetworkError
```
rahul-lohra: I think there is no need for  `Error.NetworkError`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/SignalLostSignalingServiceDecorator.kt
```diff
@@ -0,0 +1,120 @@
...
+import stream.video.sfu.signal.StartNoiseCancellationRequest
+import stream.video.sfu.signal.StartNoiseCancellationResponse
+import stream.video.sfu.signal.StopNoiseCancellationRequest
+import stream.video.sfu.signal.StopNoiseCancellationResponse
+import stream.video.sfu.signal.UpdateMuteStatesRequest
+import stream.video.sfu.signal.UpdateMuteStatesResponse
+import stream.video.sfu.signal.UpdateSubscriptionsRequest
+import stream.video.sfu.signal.UpdateSubscriptionsResponse
+
+class SignalLostSignalingServiceDecorator(private val decorated: SignalServerService, private val onPropagateError: (Error) -> Unit) : SignalServerService {
```
rahul-lohra: Can we make it internal?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/job/RestartIceJobDelegate.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package io.getstream.video.android.core.call.connection.job
+
+import io.getstream.video.android.core.utils.safeCall
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+
+open class RestartIceJobDelegate(private val scope: CoroutineScope, private val scheduledJobTimeoutMs: Long = 3000) {
```
rahul-lohra: I am curious why its `open`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/SignalLostSignalingServiceDecorator.kt
aleksandar-apostolov: Yes!

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/job/RestartIceJobDelegate.kt
aleksandar-apostolov: Because we may want to provide different ICE implementations, but this one should be internal.

PR #1477, opened 2025-07-18
Handle “accept call” action in onNewIntent and update intent reference

rahul-lohra:
### 🎯 Goal

This PR ensures correct intent handling for call actions by:

1. Calling `setIntent(intent)` inside `onNewIntent()` so that lifecycle methods like `onResume()` always receive the latest intent.
2. Handling the **"accept call"** action directly within `onNewIntent()` to allow faster and more reliable call acceptance when the activity is already running.

### 🛠 Implementation details

_Describe the implementation_

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| https://github.com/user-attachments/assets/6ebf8282-9d78-4229-b3b2-c665a604907b | https://github.com/user-attachments/assets/33907a3b-5f80-41b0-9e5e-a2ebda9af76c |

### DIFF
--- a/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
+++ b/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
@@ -256,7 +256,9 @@ public abstract class StreamCallActivity : ComponentActivity() {
                     intent,
                     onSuccess = { _, _, call, action ->
                         logger.d { "Calling [onNewIntent(intent)], because call is initialized $call, action=$action" }
-                        onIntentAction(call, action, onError = onErrorFinish) { successCall -> }
+                        onIntentAction(call, action, onError = onErrorFinish) { successCall ->
+                            applyDashboardSettings(successCall)
+                        }
                     },
                     onError = {
                         // We are not calling onErrorFinish here on purpose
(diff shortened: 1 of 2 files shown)


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -230,6 +236,42 @@ public abstract class StreamCallActivity : ComponentActivity() {
...
+    override fun onNewIntent(intent: Intent) {
+        super.onNewIntent(intent)
+        /**
+         * Necessary because the intent is read during the activity's lifecycle methods.
+         */
+        setIntent(intent)
+        when (intent.action) {
+            NotificationHandler.ACTION_ACCEPT_CALL -> {
+                // Exit case
+                // TODO Later
```
aleksandar-apostolov, line 248: Exit the current call and join the new one. If they are different. If we are accepting the same call, do nothing.

aleksandar-apostolov: This should also be customisable behaviour so an open function as hook should be added.

rahul-lohra: Sure

rahul-lohra:
Closed. It is included as part of https://github.com/GetStream/stream-video-android/pull/1478

PR #1479, opened 2025-07-25
Pause incoming video if network conditions are bad

aleksandar-apostolov:
### 🎯 Goal

The goal of this PR is to pause incoming video in bad network conditions in order to save bandwith for the audio.

### 🛠 Implementation details

1. React to new SFU events that tells the SDK which tracks are paused.
2. Keep and update paused state per participant track.
3. Add `call` api to enable/disable this feature.
4. VideoRenderer has a new badNetworkContent which is shown in case the `video.pasued == true`
5. Label is extened with a new icon to signal a bad network has caused the video to pause.
6. Demo app now enables this feature by default and shows a banner in case there are disabled videos.

### 🎨 UI Changes
image

### 🧪 Testing

This can be tested on an emulator by changing the network options.


demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -23,6 +23,7 @@ import android.content.ClipboardManager
 import android.content.Context
 import android.content.res.Configuration
 import android.media.projection.MediaProjectionManager
+import android.util.Log
```
rahul-lohra: Can we use taggedLogger?

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -174,6 +180,20 @@ fun CallScreen(
...
+    LaunchedEffect(key1 = call) {
+        while (true) {
+            delay(2000)
+            val pausedVideos = call.state.participants.value.filter { it.videoPaused.value == true }
+            Log.d("CallScreen", "Paused videos: $pausedVideos")
+            if (pausedVideos.isNotEmpty()) {
+                anyPausedVideos = true
+            } else {
+                anyPausedVideos = false
+            }
```
rahul-lohra: I think we can optimize it. This code seems like it is polling from `participants` after every 2 seconds interval irrespective of any changes 

How about we make it reactive by using `sample`

Like this -

```kotlin
LaunchedEffect(call) {
    call.state.participants
        .sample(2000) 
        .collect { participants ->
            val pausedVideos = participants.filter { it.videoPaused.value }
            Log.d("CallScreen", "Paused videos: $pausedVideos")
            anyPausedVideos = pausedVideos.isNotEmpty()
        }
}
```

The core logic here is - this collection will happen only when there is a `change `in  `call.state.participants` in 2 seconds

Or how how about suppressing the emission until `pausedVideos` itself is changed 

Like this

```kotin
call.state.participants
    .sample(2000L)
    .map { participants ->
        // Compute only paused videos
        participants.filter { it.videoPaused }
    }
    .distinctUntilChanged() // Only emit if the filtered list actually changes
    .collect { pausedVideos ->
        Log.d("CallScreen", "Paused videos: $pausedVideos")
        anyPausedVideos = pausedVideos.isNotEmpty()
    }

```

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -174,6 +180,20 @@ fun CallScreen(
 
     val connection by call.state.connection.collectAsStateWithLifecycle()
     val me by call.state.me.collectAsStateWithLifecycle()
+    var anyPausedVideos by remember { mutableStateOf(false) }
+
+    LaunchedEffect(key1 = call) {
+        while (true) {
+            delay(2000)
```
rahul-lohra, line 190: I was thinking if we can pass this time as an argument?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/SerialProcessor.kt
```diff
@@ -0,0 +1,81 @@
...
+            onFailure = { Result.failure(it) },
+        )
+    }
+
+    /**
+     * Stop the processor.
+     */
+    fun stop() {
+        workerJob?.cancel()
+        workerJob = null
```
rahul-lohra, line 85: Do we have any reason to not close the channel, once we are done?

rahul-lohra (approved):
lgtm

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
aleksandar-apostolov: No need, this is only in `demo-app` call screen. Not part of any component.

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
aleksandar-apostolov: This is a great suggestion. This code lives only in `demo-app` anyway, but is much better example.

PR #1482, opened 2025-08-01
Improve call rejection from Notification Button

rahul-lohra:
### 🎯 Goal

Improve call rejection from Notification Button

### 🛠 Implementation details

The UI will observe `Call.state.rejectedBy` if it is self then the code will try to close the activity if there is no active call

```kotlin
/**
           * Call can be rejected by [RejectCallBroadcastReceiver] so the activity
           * needs to observe [Call.state.rejectedBy]
           */
          val rejectedBy by call.state.rejectedBy.collectAsStateWithLifecycle()
          LaunchedEffect(rejectedBy) {
              val currentUserId = StreamVideo.instanceOrNull()?.userId
              if (rejectedBy.contains(currentUserId)) {
                  // check if there is no ongoing call then safely finish it else do nothing
                  val noActiveCall = (StreamVideo.instanceOrNull()?.state?.activeCall?.value == null)
                  if (noActiveCall) {
                      onEnded(call)
                      val configuration = configurationMap[call.id]
                      if (configuration?.closeScreenOnCallEnded == true) {
                          safeFinish()
                      } else {
                          logger.d { "Don't close activity as some other call is active" }
                      }
                  }
              }
          }
```

### DIFF
--- a/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
+++ b/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
@@ -129,6 +129,35 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
                 logger.d { "Created [${clazz.simpleName}] intent. -> $this" }
             }
         }
+
+        public fun callIntentBundle(
+            cid: StreamCallId,
+            members: List<String> = defaultExtraMembers,
+            leaveWhenLastInCall: Boolean = DEFAULT_LEAVE_WHEN_LAST,
+            configuration: StreamCallActivityConfiguration = StreamCallActivityConfiguration(),
+            extraData: Bundle? = null,
+        ): Bundle {
+            return Bundle().apply {
+                // Setup the outgoing call action
+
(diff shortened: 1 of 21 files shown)


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -177,7 +177,7 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
      * You can get Call id via
      * `intent?.streamCallId(NotificationHandler.INTENT_EXTRA_CALL_CID)?.id`
      */
-    private val configurationMap: HashMap<String, StreamCallActivityConfiguration> =
+    public val configurationMap: HashMap<String, StreamCallActivityConfiguration> =
```
aleksandar-apostolov: Is there a reason to expose this as public?

rahul-lohra: It is accessed in `StreamCallActivityComposeDelegate`

aleksandar-apostolov (approved):

PR #1484, opened 2025-08-04
Throw exception with message when we have non-ascii app name in headers

rahul-lohra:
### 🎯 Goal

_Describe why we are making this change_

### 🛠 Implementation details

_Describe the implementation_


aleksandar-apostolov:
I think we need to Sanitise the ASCII instead, because this is internal to our backend.

aleksandar-apostolov:
We'll handle this in `core` where we sanitise the input.

PR #1485, opened 2025-08-05
Prevent WebSocket from overwriting PN-filled notification titles

rahul-lohra:
### 🎯 Goal

1. Prevent WebSocket from overwriting PN-filled notification titles 
2. Correct the logic of picking caller-name from WS flow

This change ensures that if a push notification has already filled in the notification title, it won't be overwritten by the internal WebSocket update logic. This prioritizes more accurate or user-visible information received from push.

### 🛠 Implementation details

* Introduced a `notification` property in the `Call` object to persist the latest notification state associated with that call.
* Added a `NotificationDispatcher` class responsible for:

  * Wrapping the `NotificationManager.notify` call.
  * Saving the dispatched `Notification` instance into the associated `Call` object.
* Modified the WebSocket notification logic to check if a notification already exists for the call and skip updating the title if it was already set by the push notification flow.
* Ensures a clean separation of concerns: push notifications set rich content, and WebSocket updates fill in only when needed.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/NotificationHandler.kt
```diff
@@ -73,3 +75,8 @@ public interface NotificationHandler :
         const val INCOMING_CALL_NOTIFICATION_ID = 24756
     }
 }
+
+// TODO Rahul: Review from Alex
```
aleksandar-apostolov: Remove the TODO's

PR #1490, opened 2025-08-06
Use `call_display_name` from the PN as first option when choosing `Notification` title

dongyi-chun:
### 🎯 Goal

Change to use a call_display_name as a first option

### 🛠 Implementation details

Add a consolidated private fun to get a `callDisplayName` for all call types

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
@@ -16,6 +16,7 @@
 
 package io.getstream.video.android.core.notifications.internal
 
+import androidx.annotation.VisibleForTesting
 import io.getstream.android.push.PushDevice
 import io.getstream.android.push.delegate.AndroidPushDelegateProvider
 import io.getstream.android.push.delegate.PushDelegate
@@ -34,7 +35,6 @@ import kotlinx.coroutines.launch
  */
 internal class VideoPushDelegate : PushDelegate() {
     private val logger by taggedLogger("Call:PushDelegate")
-    private val DEFAULT_CALL_TEXT = "Unknown caller"
 
     /**
      * Handle a push message.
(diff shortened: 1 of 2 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
```diff
@@ -61,25 +61,31 @@ internal class VideoPushDelegate : PushDelegate() {
...
     }
 
     private fun handleLiveStartedType(callId: StreamCallId, payload: Map<String, Any?>) {
-        val callDisplayName = (payload[KEY_CREATED_BY_DISPLAY_NAME] as String).ifEmpty { DEFAULT_CALL_TEXT }
+        val callDisplayName = getCallDisplayName(payload)
         getStreamVideo("live-started-notification")?.onLiveCall(callId, callDisplayName, payload)
     }
 
+    private fun getCallDisplayName(payload: Map<String, Any?>): String {
+        return (payload[KEY_CALL_DISPLAY_NAME] as String).ifEmpty {
```
dongyi-chun: Not proactivley verified for the case if there are no such a key-value exist. That means it may crash in that case, need more condition if so.

aleksandar-apostolov: This is correct, `payload[key]` may indeed return a null in which case `as String` will fail. Another case is if for some reason the payload is not a `String`.

The correct approach should be something like this

```kotlin
private fun getCallDisplayName(payload: Map<String, Any?>): String {
    val callDisplayName = (payload[KEY_CALL_DISPLAY_NAME] as? String)
        ?.trim()
        ?.takeIf { it.isNotEmpty() }
    if (callDisplayName != null) {
        return callDisplayName
    }

    val createdByDisplayName = (payload[KEY_CREATED_BY_DISPLAY_NAME] as? String)
        ?.trim()
        ?.takeIf { it.isNotEmpty() }
    return createdByDisplayName ?: DEFAULT_CALL_TEXT
}

```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt
```diff
@@ -184,16 +205,17 @@ internal class VideoPushDelegate : PushDelegate() {
     private fun Map<String, Any?>.isValidIncomingCall(): Boolean =
         !(this[KEY_TYPE] as? String).isNullOrBlank() && !(this[KEY_CALL_CID] as? String).isNullOrBlank()
 
-    private companion object {
+    internal companion object {
+        internal val DEFAULT_CALL_TEXT = "Unknown caller"
```
dongyi-chun, line 209: Need to move to the string resources for i18n but I'm just leaving it to your team's future work.

aleksandar-apostolov:
no CLA

PR #1497, opened 2025-08-18
Improve call state management and activity stability

rahul-lohra:
### Goals
1. Prevent crash in `StreamCallActivity` by replacing throwing Exception with `Result.Error`
2. Fix `onLastParticipant` trigger logic to depend on state instead of events
3. Prevent CallService from being stopped by expired calls

## Fix `onLastParticipant` trigger logic to depend on state instead of events

Existing logic was giving unreliable result. Basically with only 2 participants on the call, I was getting count as 4. I have seen this issue many times
```
15:04:28.067 DefaultCallActivity      D  (DefaultDispatcher-worker-7:187) Participant [0]=Kanat
15:04:28.067 DefaultCallActivity      D  (DefaultDispatcher-worker-7:187) Participant [1]=Rahul Lohra
15:04:38.922 DefaultCallActivity      D  (main:2) Participant left, remaining: 2
15:04:38.923 DefaultCallActivity      D  (DefaultDispatcher-worker-10:239) Participant [0]=Kanat
15:04:38.923 DefaultCallActivity      D  (DefaultDispatcher-worker-10:239) Participant [1]=Rahul Lohra
15:04:39.240 DefaultCallActivity      D  (main:2) Participant left, remaining: 2
15:04:39.246 DefaultCallActivity      D  (DefaultDispatcher-worker-4:150) Participant [0]=Kanat
15:04:39.247 DefaultCallActivity      D  (main:2) Participant left, remaining: 4
```

The existing implementation has serious issue which is replaced by this logic
The issue is that `call.state.participants` is updated asynchronously it is not guaranteed that we will always get updated
`call.state.participants` inside right existing ParticipantLeftEvent

Further more, this event is emitted first then `CallState` will perform several other updates on `call.state.participants` which made this logic less reliable

### Existing implementation

```kotlin
 is ParticipantLeftEvent, is CallSessionParticipantLeftEvent -> {
                val total = call.state.participants.value.size
                logger.d { "Participant left, remaining: $total" }
                lifecycleScope.launch(Dispatchers.Default) {
                    call.state.participants.value.forEachIndexed { i, v ->
                        logger.d { "Participant [$i]=${v.name.value}" }
                    }
                }

                if (total <= 1) {
                    onLastParticipant(call)
                }
            }
```

### New implementation:
 The fix is to rely on state `cachedCall.state.participants` instead of events

```kotlin
    @CallSuper
    public open fun onCallEvent(call: Call, event: VideoEvent) {
        when (event) {
           ....

            is ParticipantLeftEvent, is CallSessionParticipantLeftEvent -> {
                participantCountJob?.cancel()
                participantCountJob = lifecycleScope.launch(supervisorJob) {
                    cachedCall.state.participants.collect {
                        logger.d { "Participant left, remaining: ${it.size}" }
                        lifecycleScope.launch(Dispatchers.Default) {
                            it.forEachIndexed { i, v ->
                                logger.d { "Participant [$i]=${v.name.value}" }
                            }
                        }
                        if (it.size <= 1) {
                            onLastParticipant(call)
                        }
                    }
                }
            }
        }
    }
```
## Prevent CallService from being stopped by expired calls

There can be a scenario where User A receives a call and rejects/cancels it, but the CallRejectedEvent has not yet been received. If User A then receives a new incoming call before the previous rejection event arrives, the late CallRejectedEvent from the expired call may incorrectly stop the CallService. This results in the active call’s notification being removed and the service being stopped prematurely.

### Existing Implementation

```
private fun observeRingingState(callId: StreamCallId, streamVideo: StreamVideoClient) {
        serviceScope.launch {
            val call = streamVideo.call(callId.type, callId.id)
            call.state.ringingState.collect  {... stopService() ...}
}
```
       
### New Implementation 

We replaced `serviceScope` with `call.scope`. The call.scope will be automatically destoyed once the call is cleaned up
```kotlin
 private fun observeRingingState(callId: StreamCallId, streamVideo: StreamVideoClient) {
            val call = streamVideo.call(callId.type, callId.id)
        call.scope.launch {
            call.state.ringingState.collect  {... stopService() ...}
}
```


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -931,17 +992,20 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
             }
 
             is ParticipantLeftEvent, is CallSessionParticipantLeftEvent -> {
-                val total = call.state.participants.value.size
-                logger.d { "Participant left, remaining: $total" }
-                lifecycleScope.launch(Dispatchers.Default) {
-                    call.state.participants.value.forEachIndexed { i, v ->
-                        logger.d { "Participant [$i]=${v.name.value}" }
+                participantCountJob?.cancel()
+                participantCountJob = lifecycleScope.launch(supervisorJob) {
```
aleksandar-apostolov: Why do we re-launch a coroutine on every event?

aleksandar-apostolov (approved):

PR #1502, opened 2025-08-28
Open UserDevice APIs and add more logs

dongyi-chun:
### 🎯 Goal

Make available to access `UserDevice` functions by API

### 🛠 Implementation details

- Adding `getDevice()` and `updateDevice()` API
- Adding more log in `createDevice()`, `updateDevice()` and `removeStoredDevice()`


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/StreamNotificationManager.kt
```diff
@@ -45,7 +45,6 @@ import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.launch
 
 internal class StreamNotificationManager private constructor(
-    private val context: Context,
```
rahul-lohra: This is breaking change @dongyi-chun

PR #1505, opened 2025-09-01
Introduce PR quality checks from common repo

aleksandar-apostolov:
### Goal

Introduce the quality check for PRs

### Implementation
Use the common workflow

### Testing
Will show up in PR.


.github/workflows/pr-quality.yml
```diff
@@ -0,0 +1,16 @@
+# .github/workflows/pr-quality-check.yml (caller)
+name: PR checklist
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, labeled, unlabeled, reopened]
```
testableapple, line 6: Do we need to run this check on edited/labeled/unlabeled?

aleksandar-apostolov: yes, because if you remove the label it needs to tell you that you have to add a pr:* label which is then used in the release notes i.e. PRs which don't have the label are not included in the release notes.

PR #1508, opened 2025-09-02
Enhance SDK stability and memory efficiency in large rooms

PratimMallick:
### Goal  
Improve the stability and performance of the SDK, especially when joining large rooms.  

### Implementation  
This PR introduces the following changes:  

1. **Decoupled `ParticipantState` from `Call`**  
   - `ParticipantState` no longer holds a reference to the `Call` instance, reducing coupling and improving maintainability.  

2. **Lazy initialization of `MutableStateFlows` in `ParticipantState`**  
   - Previously, ~20+ `MutableStateFlows` were eagerly initialized.  
   - Now, they are lazily created only when needed, reducing the memory footprint.  

3. **Use `ConcurrentHashMap` in `CallState`**  
   - Replaced `SortedMap` with `ConcurrentHashMap` for storing participants.  
   - Benefits:  
     - Improved performance when creating copies.  
     - Reduced risk of data corruption due to race conditions.  

4. **Serial processing of WebRTC events in `RtcSession`**  
   - Introduced a `SerialProcessor` backed by a single thread to handle WebRTC-related jobs.  
   - Ensures events are processed sequentially and in order.  
   - Prevents native crashes caused by simultaneous joins/leaves when handling large rooms.  

5. RTCSession gets the rtcScope from a Scope provider that reuses the single thread executor across sessions. Every Call gets its own Executor which is shutdown with the call cleanup
  
5. **Added logging to `SerialProcessor` so that we can view the oder in which jobs are submitted and executed in the queue

### 🎨 UI Changes  
None.  

### Testing  
Since this PR affects the SDK’s core WebRTC handling, thorough regression testing is required across all flows involving:  
- Track publishing  
- Track subscribing  
- Participant join/leave in large rooms


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ParticipantState.kt
```diff
@@ -62,24 +65,24 @@ public data class ParticipantState(
...
 
     val isLocal by lazy {
-        sessionId == call.sessionId
+        callActions.isLocalParticipant(sessionId)
     }
 
     /** video track */
-    internal val _videoTrack = MutableStateFlow<VideoTrack?>(null)
+    internal val _videoTrack by lazy { MutableStateFlow<VideoTrack?>(null) }
     val videoTrack: StateFlow<VideoTrack?> = _videoTrack
```
gpunto, line 75: I wonder what's the memory difference for a case like this. 🤔 How big is a state flow in memory right after init without lazy? Keep in mind that the lazy delegate also creates an object, and that object will live forever in memory, whereas before we only had the flow. If it's only a few bytes, the tradeoff might not be worth it

We're also accessing the property immediately at construction time when we assign it to the public property, so this isn't lazy in the end

aleksandar-apostolov: Agreed, 
Also I think the problem may not be in `_videoTrack` being lazy or not but in the other derived flows tbh.

e.g.
```
val _flow1 : Flow
val _flow2: Flow

val flow3 = combine(flow1, flow2) {

}
```

Wouldn't this create a coroutine to immediately collect the two flows. We have this on several places in this class.
`videoEnabled: Flow<Boolean>` is such an example.

PratimMallick: Good point about the property being accessed imemediately! I dont think then lazy would have any effect. 

But from memory point of view, I had assumed that when hundreds of participants join together it would be swell up the memory

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1281,13 +1310,13 @@ public class CallState(
         userId: String,
         updateFlow: Boolean = false,
     ): ParticipantState {
-        val participantMap = _participants.value.toSortedMap()
-        val participantState = if (participantMap.contains(sessionId)) {
-            participantMap[sessionId]!!
+        val participantState = if (internalParticipants.containsKey(sessionId)) {
+            internalParticipants[sessionId]!!
```
aleksandar-apostolov, line 1336: Why not `internalParticipants[sessionId] ?: ParticipantState(...)`

aleksandar-apostolov: we can avoid the `if/else` and the `!!` operator.

PratimMallick: Modified

PratimMallick: Making this change is leading weirdly to failure of an E2E test - reconnectionTest .. but I am unable to find its logs file in allure test ops

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -212,6 +218,18 @@ public class RtcSession internal constructor(
     private val supervisorJob: CompletableJob = SupervisorJob(),
     private val coroutineScope: CoroutineScope =
         CoroutineScope(clientImpl.scope.coroutineContext + supervisorJob),
+    private val rtcSessionExecutor: Executor = Executors.newSingleThreadExecutor { runnable ->
+        Thread(runnable, "rtc-session-coroutine").apply {
+            isDaemon = true
+        }
+    },
+    private val rtcSessionScope: CoroutineScope = CoroutineScope(
+        clientImpl.scope.coroutineContext +
```
rahul-lohra: Any reason to use `clientImpl.scope.coroutineContext`? 

Ideally we should use call.state.scope because when we clear the call scope then all its tasks are cleared automatically 

So I can recommend something like this

```kotlin
private val rtcSessionScope: CoroutineScope =
        call.scope +
                supervisorJob +
                CoroutineName("rtc-session-coroutine") +
                rtcSessionExecutor.asCoroutineDispatcher(),
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -524,7 +542,7 @@ public class RtcSession internal constructor(
      */
     suspend fun reconnect(forceRestart: Boolean) {
         // ice restart
-        val subscriberAsync = coroutineScope.async {
+        val subscriberAsync = rtcSessionScope.async {
```
rahul-lohra, line 535: Is `async` truly effective within rtcSessionScope since it's backed by a single thread? If not then maybe we make the code more simple

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -767,8 +785,11 @@ public class RtcSession internal constructor(
         logger.i { "[cleanup] #sfu; #track; no args" }
 
         coroutineScope.launch {
-            sfuConnectionModule.socketConnection.disconnect()
-            sfuConnectionMigrationModule?.socketConnection?.disconnect()
+            serialProcessor.submit("cleanupSfuConnections") {
+                sfuConnectionModule.socketConnection.disconnect()
+                sfuConnectionMigrationModule?.socketConnection?.disconnect()
+                Unit
```
rahul-lohra, line 782: I think we can skip writing Unit at the end. Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -767,8 +785,11 @@ public class RtcSession internal constructor(
         logger.i { "[cleanup] #sfu; #track; no args" }
 
         coroutineScope.launch {
-            sfuConnectionModule.socketConnection.disconnect()
-            sfuConnectionMigrationModule?.socketConnection?.disconnect()
+            serialProcessor.submit("cleanupSfuConnections") {
```
rahul-lohra, line 779: Tbh this code looks ugly to me _(The nesting of serialProcessor.submit() inside `coroutineScope.launch` )_ Maybe we can improve the code style. Wdyt?

```
coroutineScope.launch {
    serialProcessor.submit("..") {}
}
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -212,6 +218,18 @@ public class RtcSession internal constructor(
     private val supervisorJob: CompletableJob = SupervisorJob(),
     private val coroutineScope: CoroutineScope =
         CoroutineScope(clientImpl.scope.coroutineContext + supervisorJob),
+    private val rtcSessionExecutor: Executor = Executors.newSingleThreadExecutor { runnable ->
```
rahul-lohra: Why don't we use this api name `limitedParallelism` from coroutines. The api seems easy to use 

```
val mySingleThreadDispatcher = Dispatchers.Default.limitedParallelism(1, "incrementDispatcher")
```

link: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/limited-parallelism.html

rahul-lohra (commented):
Please check my comments @PratimMallick

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Although limited parallelism ensures that there will be a single thread to execute always but it does not ensure that it will be the same thread being used every time, which is the requirement for webrtc. Hence the use of Executor which ensures this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: yes style can be improved. Any suggestions?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: It was not compiling without this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: The reason was keeping it at the same scope as the coroutineScope which was being used earlier

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: We need `async` since we have to wait for the result of subscriberAsync and publisherAsync before we return from the  `reconnect` method

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Maybe can have an extension function

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Although this scope is backed by a single thread but still it can run two routines co-operatedly. If we use serial processor then first subscriber ice restart will finish and then only publish ice restart will work. where as we need them to run on a co-operate manner

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ParticipantState.kt
PratimMallick: Removed all the lazy loading as it wasnt being beneficial

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/scope/ScopeProviderImpl.kt
```diff
@@ -0,0 +1,79 @@
...
+                    isDaemon = true
+                }
+            }
+        }
+
+        return CoroutineScope(
+            coroutineScope.coroutineContext +
+                supervisorJob +
+                CoroutineName("rtc-session-coroutine") +
+                executor!!.asCoroutineDispatcher(),
```
rahul-lohra: Try to avoid force-non `executor!!` null. Even if your code is correct. The code quality will flag it as dangerous and will cause unnecessary noise 

We could have just return a coroutine with io dispatcher in `null` case and you can write comment that the io one will one execute in real scenario

PratimMallick: Updated to a fallback mechanism where if the executor is not ready, the provider will send a regular coroutine scope

rahul-lohra (approved):
lgtm

PR #1512, opened 2025-09-08
Enhance Call Lifecycle Management, Demo-App PN Flows, and Demo-App SDK Initialization

rahul-lohra:
### Goal

Core-SDK Goals

1. Improve handling of delayed events like _(CallRejectedEvent)_ caused by network disruptions.
2. Ensure active calls & ringing calls are not cleared during expired-calls cleanup.
3. Prevent premature service shutdown during expired-calls cleanup

### Demo-App Goals

1. Improve PN device registration & token renewal flow 

Added support for initializing the Video SDK via a content provider.

## 1.  Improved handling of delayed CallRejectedEvent caused by network disruptions.

We introduced a new abstraction, **`SlowEvent`**, to represent events that are detected or reconstructed belatedly (e.g., due to push notification timing or network delays).

* Added a new marker interface `SlowEvent`.
* Implemented `CallRejectedSlowEvent` to mirror the behavior of `CallRejectedEvent` when it arrives late.
* Introduced `NoSlowEvent` as the default placeholder representing the absence of any slow event.
* Added state flow support (`slowEvent: StateFlow<SlowEvent>`) to allow SDK components to observe and react to slow events consistently.

This ensures that when a `CallRejectedEvent` is delayed, the SDK can still execute the appropriate logic — either mirroring the original event’s instructions or applying modified handling where needed.

## 2. Ensure active calls & ringing calls are not cleared during expired-calls cleanup.

we introduced 2 new internal apis to resolve this

| Visibility | Method signature                  |
|------------|-----------------------------------|
| internal   | `fun removeActiveCall(call: Call)` |
| internal   | `fun removeRingingCall(call: Call)` |

#### Background
Previously, when a call was cleaned up, the SDK would directly invoke `context.stopService()`.  
This caused an issue where the service could be shut down even if there were still active calls running.

#### Improvement
We now delegate shutdown control to the service itself.  
The service inspects incoming intents to determine whether the stop request originates from an **active call** or an **expired call**.  
Only when no active calls remain will the service shut down.

#### Changes introduced

- Added a new bundle key to tag stop-service requests:

| Type | Bundle Key |
|------|------------|
| Bool | `val EXTRA_STOP_SERVICE = "io.getstream.video.android.core.stop_service"` |

- Introduced two private helper methods:

| Visibility | Method signature |
|------------|------------------|
| private | `fun shouldStopServiceFromIntent(intent: Intent?): Boolean` |
| private | `fun isIntentForExpiredCall(intent: Intent?): Boolean` |

#### Behavior

- If the intent indicates a request from an expired call, the service ignores it by returning `START_NOT_STICKY`.  
- If the intent is from an active call and no other calls remain, the service proceeds with shutdown.

##### Strategy 1
1. User A will call User B
2. User B will toggle internet-off  
3.  User A will cancel the call to User B
5. User B will toggle internet-on  
4. User A will call again to User B

##### Strategy 2
Forcefully add delay in the processing of _(CallRejectedEvent)_  
Repeat everything from ##### Strategy 1 except internet toggling

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,8 @@
-### 🎯 Goal
+### Goal
 
 _Describe why we are making this change_
 
-### 🛠 Implementation details
+### Implementation
 
 _Describe the implementation_
 
@@ -35,7 +35,7 @@ _Add relevant videos_
 </tbody>
 </table>
 
-### 🧪 Testing
+### Testing
 
 _Explain how this change can be tested (or why it can't be tested)_
 
--- a/.github/workflows/pr-clean-old-drafts.yaml
+++ b/.github/workflows/pr-clean-old-drafts.yaml
@@ -1,0 +1,18 @@
+# .github/workflows/pr-quality-check.yml (caller)
+name: PR checklist
+
+on:
+  schedule:
+    - cron: "0 7 * * 1"  # every Monday 09:00 CEST (UTC+2)
+  workflow_dispatch:
+
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
+jobs:
+  pr-checklist:
+    uses: GetStream/android-ci-actions/.github/workflows/pr-clean-stale.yaml@main
+    secrets: inherit
--- a/.github/workflows/pr-clean-old-drafts.yaml
+++ b/.github/workflows/pr-clean-old-drafts.yaml
@@ -1,5 +1,5 @@
 # .github/workflows/pr-quality-check.yml (caller)
-name: PR checklist
+name: PR cleanup
 
 on:
   schedule:
--- a/.github/workflows/pr-quality.yml
+++ b/.github/workflows/pr-quality.yml
@@ -1,0 +1,16 @@
+# .github/workflows/pr-quality-check.yml (caller)
+name: PR checklist
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, labeled, unlabeled, reopened]
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
+jobs:
+  pr-checklist:
+    uses: GetStream/android-ci-actions/.github/workflows/pr-quality.yml@main
+    secrets: inherit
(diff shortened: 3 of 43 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/slowevent/SlowEventContext.kt
```diff
@@ -0,0 +1,32 @@
...
+ *    https://github.com/GetStream/stream-video-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.slowevent
```
aleksandar-apostolov: Shouldn't we delete this as its not needed anymore?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/slowevent/SlowEventDetector.kt
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.slowevent
+
+import io.getstream.video.android.core.Call
+
+internal interface SlowEventDetector {
```
aleksandar-apostolov: Same as `SlowEventContext`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/slowevent/SlowEventContext.kt
rahul-lohra: Deleted

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/slowevent/SlowEventDetector.kt
rahul-lohra: Deleted

PR #1515, opened 2025-09-11
Always play ringtone despite audio focus

rahul-lohra:
### Goal

Currently the incoming call ringtone was not playing when the app was on background
### Implementation

Always play call ringtone despite audio focus

### Testing

Smoke test incoming call when app is on foreground, background


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/sounds/CallSoundPlayer.kt
```diff
@@ -86,7 +86,7 @@ internal class CallSoundPlayer(private val context: Context) {
         }
 
         logger.d { "[requestAudioFocus] Audio focus " + if (isGranted) "granted" else "not granted" }
-        if (isGranted) onGranted()
+        if (isGranted) onGranted() else onGranted() // Forcing as playing ringtone do not ideally requires audio focus
```
aleksandar-apostolov: Why not `onGranted()` without the check?

aleksandar-apostolov: Also add safety net around this.

rahul-lohra: Double check it, its protected by `try-catch` already

PR #1518, opened 2025-09-15
Enhance retry handling for failed call joins.

rahul-lohra:
### Goal

Enhance retry handling for failed call joins.

### Implementation

1. Treat no internet available as a non-permanent error in `isPermanentError` and allow retries.
2. Observe `RealtimeConnection.Failed` to clear the ringing call notification.

### Testing

1. Add a breakpoint on where we invoke `coordinatorConnectionModule.api.joinCall` in StreamVideoClient class
2. Toggle off internet
3. Resume the breakpoint

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -445,10 +445,18 @@ public class Call(
             }
             delay(retryCount - 1 * 1000L)
         }
-        return Failure(value = Error.GenericError("Join failed after 3 retries"))
+        session = null
+        val errorMessage = "Join failed after 3 retries"
+        state._connection.value = RealtimeConnection.Failed(errorMessage)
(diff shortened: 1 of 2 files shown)


aleksandar-apostolov (approved):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -445,10 +445,18 @@ public class Call(
...
-        return Failure(value = Error.GenericError("Join failed after 3 retries"))
+        session = null
+        val errorMessage = "Join failed after 3 retries"
+        state._connection.value = RealtimeConnection.Failed(errorMessage)
+        return Failure(value = Error.GenericError(errorMessage))
     }
 
     internal fun isPermanentError(error: Any): Boolean {
+        if (error is Error.ThrowableError) {
+            if (error.message.contains("Unable to resolve host")) {
```
dongyi-chun, line 456: This will make minimum retry, any reason for filtering out with error message?

PR #1521, opened 2025-09-17
Improved reconnection flow when network is lost for a brief period of time

aleksandar-apostolov:
### Goal

When network is lost and recovered the SDK would do a full rejoin instead of reconnecting to the existing session. This causes a `ParticipantLeftEvent` on the other party where sometimes it can be an issue.

### Implementation

- Call `fastReconnect()` inbstead of `rejoin()`
- Do not automatically `fastReconnect()` on socket events, let the `Call` object decide when network is recovered to decide what to do.
- Improve network connectivity detection
- Serial processing for the `publishStream` of the `Publisher` to avoid negotiation with the wrong tracks
- Create the publisher ONLY IF it is not created in `RtcSession` (this is what caused the inability to correctly reconnect)
- Add some tracing and debuging on the `fastReconnect()` path and peer connection factory.

### Testing

1. In an established call Turn off the Wifi and after 5 seconds turn it back on. 
2. You should do only `fastReconnect()` in this case. 
3. The absence of any `[rejoin]` tags in the logs will confirm this.
4. Verify that video and audio is restored

### DIFF
--- a/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/rtc/RtcSessionTest2.kt
+++ b/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/rtc/RtcSessionTest2.kt
@@ -284,7 +284,7 @@ class RtcSessionTest2 {
             "sdpMLineIndex": 0,
             "candidate": "candidate-data",
             "usernameFragment": "fake-username-frag"}
-                """.trimIndent(),
+            """.trimIndent(),
             peerType = PeerType.PEER_TYPE_PUBLISHER_UNSPECIFIED,
         )
 


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -553,7 +564,7 @@ public class Call(
             session?.publisher?.iceState?.collect {
                 when (it) {
                     PeerConnection.IceConnectionState.FAILED, PeerConnection.IceConnectionState.DISCONNECTED -> {
-                        session?.publisher?.restartIce()
+                        // session?.publisher?.restartIce("PeerConnection.IceConnectionState.FAILED or PeerConnection.IceConnectionState.DISCONNECTED: $it")
```
martinmitrevski: is this on purpose?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/SfuSocketStateService.kt
```diff
@@ -290,12 +302,12 @@ internal class SfuSocketStateService(initialState: SfuSocketState = SfuSocketSta
...
                 }
                 onEvent<SfuSocketStateEvent.Stop> { SfuSocketState.Disconnected.Stopped }
-                onEvent<SfuSocketStateEvent.NetworkAvailable> {
-                    SfuSocketState.RestartConnection(
-                        RestartReason.NETWORK_AVAILABLE,
-                        WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN,
-                    )
-                }
+//                onEvent<SfuSocketStateEvent.NetworkAvailable> {
+//                    SfuSocketState.RestartConnection(
```
martinmitrevski: same here

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
aleksandar-apostolov: Yes, it is. We are deleting this code.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/SfuSocketStateService.kt
aleksandar-apostolov: Same as above

aleksandar-apostolov:
> check the commented code

Removed that as it is not needed.

PR #1528, opened 2025-09-24
Update `ownCapabilities` for local user only

aleksandar-apostolov:
### Goal

`ownCapabilities` should be done for local user only.

### Implementation

When `UpdateCallPermissionsEvent` is received check the `event.user.id ==. client.userId` before processing the event.

### Testing

Change capabilities on a call for different user and verify that the current user has not changed its own capabilities.


rahul-lohra (approved):
Lgtm

PR #1543, opened 2025-10-09
Upgrade to webrtc m137 and remove stream-webrtc-android-ui dep

PratimMallick:
### Goal

To upgrade to m137 version of webrtc and remove stream-webrtc-android-ui dependancy

### Implementation

Compiled from the m137 head of webrtc along with other patches 
Removed depedancy on webrtc-android-ui
Moved `VideoTextureViewRenderer` to core

### Testing

Install and test all features related to webrtc

### DIFF
--- a/stream-video-android-core/api/stream-video-android-core.api
+++ b/stream-video-android-core/api/stream-video-android-core.api
@@ -7212,6 +7212,8 @@ public final class io/getstream/video/android/core/Call {
 	public final fun isVideoEnabled ()Z
 	public final fun join (ZLio/getstream/video/android/core/CreateCallOptions;ZZLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun join$default (Lio/getstream/video/android/core/Call;ZLio/getstream/video/android/core/CreateCallOptions;ZZLkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
+	public final fun kickUser (Ljava/lang/String;ZLkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static synthetic fun kickUser$default (Lio/getstream/video/android/core/Call;Ljava/lang/String;ZLkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
 	public final fun leave ()V
 	public final fun listRecordings (Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun listRecordings$default (Lio/getstream/video/android/core/Call;Ljava/lang/String;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
--- a/buildSrc/src/main/kotlin/io/getstream/video/android/Configuration.kt
+++ b/buildSrc/src/main/kotlin/io/getstream/video/android/Configuration.kt
@@ -5,7 +5,7 @@ object Configuration {
     const val targetSdk = 35
     const val minSdk = 24
     const val majorVersion = 1
-    const val minorVersion = 12
+    const val minorVersion = 13
     const val patchVersion = 0
     const val versionName = "$majorVersion.$minorVersion.$patchVersion"
     const val versionCode = 57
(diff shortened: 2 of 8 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/renderer/VideoTextureViewRenderer.kt
```diff
@@ -0,0 +1,252 @@
...
+ *    https://github.com/GetStream/stream-video-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.renderer
```
rahul-lohra: I’m a bit concerned about the breaking change introduced by this package rename.
The stream-webrtc-android-ui dependency previously contained this class under a different package, and now moving it into the SDK with a new package name makes it a public API breaking change.

Ideally, clients shouldn’t be impacted by such changes unless it’s part of a major SDK revamp.
How about reverting to the original package name for now and correcting it in the next planned major release?

I can prpopose a solution: 

Just replace its package name with existing one: `package io.getstream.webrtc.android.ui`

You may add a deprecation notice on top of this class as 

```kotlin
@Deprecated(
    message = "This class will be moved to io.getstream.video.android.core.renderer in the next major release.",
    replaceWith = ReplaceWith(
        "io.getstream.video.android.core.renderer.VideoTextureViewRenderer"
    )
)```

File: stream-video-android-core/api/stream-video-android-core.api [hunk 2]
```diff
@@ -7203,8 +7203,8 @@ public final class io/getstream/video/android/core/Call {
 	public final fun grantPermissions (Ljava/lang/String;Ljava/util/List;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun handleEvent (Lio/getstream/android/video/generated/models/VideoEvent;)V
 	public final fun hasCapability ([Lio/getstream/android/video/generated/models/OwnCapability;)Z
-	public final fun initRenderer (Lio/getstream/webrtc/android/ui/VideoTextureViewRenderer;Ljava/lang/String;Lstream/video/sfu/models/TrackType;Lkotlin/jvm/functions/Function1;Ljava/lang/String;)V
-	public static synthetic fun initRenderer$default (Lio/getstream/video/android/core/Call;Lio/getstream/webrtc/android/ui/VideoTextureViewRenderer;Ljava/lang/String;Lstream/video/sfu/models/TrackType;Lkotlin/jvm/functions/Function1;Ljava/lang/String;ILjava/lang/Object;)V
+	public final fun initRenderer (Lio/getstream/video/android/core/renderer/VideoTextureViewRenderer;Ljava/lang/String;Lstream/video/sfu/models/TrackType;Lkotlin/jvm/functions/Function1;Ljava/lang/String;)V
```
aleksandar-apostolov: Breaking change, we should revert it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/renderer/VideoTextureViewRenderer.kt
PratimMallick: Okay! This makes sense

stream-video-android-core/api/stream-video-android-core.api
PratimMallick: Moving `VideoTextureViewRenderer` to same named package in core so that it doesnt break the public api

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/renderer/VideoTextureViewRenderer.kt
PratimMallick: Updated

stream-video-android-core/api/stream-video-android-core.api
PratimMallick: Updated

PR #1545, opened 2025-10-10
Ensure microphone stays active during backgrounded audio/video calls

rahul-lohra:
### Goal

Ensure the microphone remain active when the app goes to the background by preventing the call-related services from being marked as short-lived.

### Implementation

1. Correctly use service type as `ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL or ServiceInfo.FOREGROUND_SERVICE_TYPE_CAMERA or ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE` for video/audio cal types

### Testing

Smoke test audio/video call, put the application on background or turn off the screen

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/permission/android/DefaultStreamPermissionCheck.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/permission/android/DefaultStreamPermissionCheck.kt
@@ -24,13 +24,13 @@ import io.getstream.video.android.core.Call
  */
 internal class DefaultStreamPermissionCheck : StreamPermissionCheck {
 
-    @Deprecated("Use checkAndroidPermissionsV2 instead")
+    @Deprecated("Use checkAndroidPermissionsGroup instead")
     override fun checkAndroidPermissions(
         context: Context,
         call: Call,
-    ) = checkAndroidPermissionsV2(context, call).first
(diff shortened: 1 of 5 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/permission/android/DefaultStreamPermissionCheck.kt
```diff
@@ -23,8 +23,17 @@ import io.getstream.video.android.core.Call
...
 internal class DefaultStreamPermissionCheck : StreamPermissionCheck {
+
+    @Deprecated("Use checkAndroidPermissionsV2 instead")
     override fun checkAndroidPermissions(
         context: Context,
         call: Call,
-    ): Boolean = checkPermissionsExpectations(context, call)
+    ) = checkAndroidPermissionsV2(context, call).first
+
+    override fun checkAndroidPermissionsV2(
```
aleksandar-apostolov: Why not rename this to
`checkAndroidPermissionsGroup` or something in that direction. This way `V2` makes no semantic sense.

rahul-lohra: Yeah sure, its better

PR #1550, opened 2025-10-20
Add the possibility to vibrate during an incoming call.

aleksandar-apostolov:
### Goal

[AND-861](https://linear.app/stream/issue/AND-861/yoobic-lack-of-vibration-on-android-call-sdk-for-incoming-calls)

Add the possibility to vibrate during an incomming call.

### Implementation

`CallSoundPlayer` - > Renamed to `CallSoundAndVibrationPlayer`.
Added `vibrate()` and `stopVibration()` methods.
Using internal `android.os.Vibrator` to do a repeatable vibration pattern.

Added a config parameter on the `StreamVideoBuilder`. By default the configuration allows setting a `pattern` and on/off flag as in `enabled:Boolean`.
By default the vibrations are disabled as this is the current behaviour.

Added three utility methods for ready to use configurations.

### Testing

Vibrations are enabled in demo-app, run the demo app and make a ringing call.


rahul-lohra (approved):
lgtm

PR #1560, opened 2025-10-30
Add support for stereo playback

PratimMallick:
### Goal
To enable stereo playout and provide ability to switch between stereo and mono playout mid call

### Implementation

Provided `setAudioUsage` method on `SpeakerManager` that can be used after joining a call to change the current audio mode from `VOICE_COMMUNICATION` to `MEDIA` that will enable stereo playout. 
A StateFlow `audioUsage` is added as well which apps can observe to know the current audio usage

### Testing
A debug option is added in the debug menu option to toggle between mono and stereo playout. 
To test stereo output, join a call from android and play a screenshare track from youtube on web containing stereo sound. Toggle to stereo mode from android and observe the stereo effect.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -53,7 +53,7 @@ tink = "1.9.0"
 turbine = "0.13.0"
 itu = "1.7.3"
 
-streamWebRTC = "137.1.0"
+streamWebRTC = "137.1.1"
 streamNoiseCancellation = "1.0.4"
 streamResult = "1.3.0"
 streamChat = "6.10.0"


aleksandar-apostolov (approved):

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/MenuDefinitions.kt
```diff
@@ -288,6 +292,8 @@ fun debugSubmenu(
     isIncomingVideoEnabled: Boolean,
     onToggleIncomingVideoEnabled: (Boolean) -> Unit,
     loadTranscriptions: suspend () -> List<MenuItem>,
+    audioUsageUiState: AudioUsageUiState,
+    onToggleAudioUsage: suspend () -> Unit,
```
rahul-lohra: No need for `suspend` as `call.speaker.setAudioUsage(newAudioUsage)` is not a suspending function

Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -164,97 +166,106 @@ public class StreamPeerConnectionFactory(
                 .createInitializationOptions(),
         )
 
+        adm = initAudioDeviceModule()
```
rahul-lohra, line 169: Can we make it `internal` ?

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/MenuDefinitions.kt
PratimMallick: yes thats a good catch.. not needed to be suspend

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
PratimMallick: hmm, actually we can make it private and jsut expose a public method that will update it usage, 
Other classes do not need the adm instance

PratimMallick: Will do in the next PR of hifi audio

PR #1564, opened 2025-11-04
Added support for High fidelity (hi-fi) audio capture

PratimMallick:
### Goal

This PR adds support for HiFi (high-fidelity) audio in the Stream Video Android SDK, enabling users to choose a certain type of audio profile before joining the call.
 This enhancement allows applications to provide better audio quality for music streaming, podcast calls, or any scenario requiring higher audio fidelity

### Implementation

1. **Audio Bitrate Profile Support** (`MediaManager.kt`):
   - Added `audioBitrateProfile` StateFlow in `MicrophoneManager` to track current audio profile
   - Implemented `setAudioBitrateProfile()` method that validates:
     - HiFi audio must be enabled in dashboard settings to use `MUSIC_HIGH_QUALITY`
     - Profile can only be changed before joining a call
   - Added `AudioBitrateProfile` enum support with `VOICE_STANDARD_UNSPECIFIED` and `MUSIC_HIGH_QUALITY` options

2. **Peer Connection Factory Refactoring** (`StreamPeerConnectionFactory.kt`):
   - Refactored factory creation to accept `audioBitrateProfileProvider` function, which enables it to turn off hardware AEC, NS 
   - Factory is recreated when audio profile changes (before joining)
   
 3. **Call State Management** (`Call.kt`):
   - Added `ensureFactoryMatchesAudioProfile()` to validate factory matches current profile
   - Implemented `recreateFactoryAndAudioTracks()` to rebuild factory and tracks. This is needed if app has to provide a toggle in Lobby screen, where intial factory is already created with the default audioProfile
   - Made video/audio tracks and sources nullable and lazy-initialized for proper recreation
   - Shared EGL base context between factory and media manager to break circular dependency

4. **UI Components** (`ToggleHifiAudioAction.kt`):
   - New composable action button for toggling HiFi audio
   - Integrates with `CallAction` system via `ToggleHifiAudio` action
   - Provides visual feedback with music note icons (MusicNote/MusicOff)

5. **Demo App Integration**:
   - Added HiFi audio toggle in `CallLobbyScreen` and `CallScreen`

### 🎨 UI Changes
Screenshot_20251105_122805

### Testing

1. **HiFi Audio Toggle**:
   - Start a call from the demo app
   - Use the music note icon button in call controls to toggle HiFi audio
   - Verify that the publishing bitrate of audio is 128 kbps, 
   - Verify that remote audio is looped back and sent again creating an echo which denotes that AEC is turned off 
   - If you can, then play an instrument of your choice and observe the quality on the other side. The high notes should not be cut off denoting , both NS and AGC are turned off

### DIFF
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -1,6 +1,6 @@
 {
   "debug": {
-    "stream-video-android-core": 13796,
+    "stream-video-android-core": 13780,
     "stream-video-android-ui-xml": 7376,
     "stream-video-android-ui-compose": 9764
   },


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -103,24 +113,25 @@ public class StreamPeerConnectionFactory(
...
-        EglBase.create()
+        sharedEglBaseProvider()
     }
 
     /**
      * Default video decoder factory used to unpack video from the remote tracks.
      */
     private val videoDecoderFactory by lazy {
-        SelectiveVideoDecoderFactory(eglBase.eglBaseContext)
+        SelectiveVideoDecoderFactory(sharedEglBaseProvider().eglBaseContext)
```
rahul-lohra, line 126: Do you think calling sharedEglBaseProvider() creates a new instance each time?
If so, should we avoid invoking it repeatedly?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -510,4 +546,29 @@ public class StreamPeerConnectionFactory(
...
     }
+
+    /**
+     * Updates the audio track usage for the audio device module.
+     * This allows toggling between different audio usage types (e.g., USAGE_MEDIA vs USAGE_VOICE_COMMUNICATION).
+     *
+     * @param audioUsage The audio usage value to set (e.g., AudioAttributes.USAGE_MEDIA or AudioAttributes.USAGE_VOICE_COMMUNICATION)
+     * @return true if the update was successful, false if the ADM is not available or the update failed
+     */
+    public fun updateAudioTrackUsage(audioUsage: Int): Boolean {
```
rahul-lohra: Do we really need to make it `public`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -510,4 +546,29 @@ public class StreamPeerConnectionFactory(
...
+        return adm?.updateAudioTrackUsage(audioUsage) ?: false
+    }
+
+    /**
+     * Disposes the factory and releases resources.
+     * This should only be called when no active peer connections are using it.
+     */
+    public fun dispose() {
```
rahul-lohra: Do we really need to make it `public`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -232,13 +233,97 @@ public class Call(
...
+     * If the factory hasn't been created yet, it will be created with the current profile
+     * when first accessed, so no recreation is needed.
+     */
+    internal fun ensureFactoryMatchesAudioProfile() {
+        val factory = _peerConnectionFactory
+
+        // If factory hasn't been created yet, it will be created with current profile automatically
+        if (factory == null) {
+            return
+        }
```
rahul-lohra, line 278: I am not sure if this case will ever be true with current implementation. Because the getter of _peerConnectionFactory will always return a non-null value. Wdyt?

rahul-lohra (commented):
please check my comments

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
PratimMallick: It can be internal function

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
PratimMallick: This can be internal as well

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: In our code,

```   
 private var _peerConnectionFactory: StreamPeerConnectionFactory? = null

internal var peerConnectionFactory: StreamPeerConnectionFactory
        get() {
            if (_peerConnectionFactory == null) {
                _peerConnectionFactory = StreamPeerConnectionFactory(
                    context = clientImpl.context,
                    audioProcessing = clientImpl.audioProcessing,
                    audioUsage = clientImpl.callServiceConfigRegistry.get(type).audioUsage,
                    audioUsageProvider = { clientImpl.callServiceConfigRegistry.get(type).audioUsage },
                    audioBitrateProfileProvider = { mediaManager.microphone.audioBitrateProfile.value },
                    sharedEglBaseProvider = { eglBase },
                )
            }
            return _peerConnectionFactory!!
        }
        set(value) {
            _peerConnectionFactory = value
        }
```
        
We have the creation of factory instance in the getter of `peerConnectionFactory` and `peerConnectionFactory` can be null when  `peerConnectionFactory` hasn't been accessed till now 
        
The reason for the check 

```
if (factory == null) {
            return
        }
```
is that `join` maybe called without factory initialization (when calling from apps that do not have preview). And in such a case further checks are not needed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
PratimMallick: no it does not create a new instance on calling `sharedEglBaseProvider()` as it 
Calls the lambda directly. It uses the instance which Call class provided.

rahul-lohra (approved):
lgtm

PR #1570, opened 2025-11-12
Add possibility to share screen audio when screensharing is enabled

PratimMallick:
### Goal

This PR adds support for capturing and transmitting screen audio during screen sharing on Android. Previously, when users shared their screen, only the video was transmitted. This change enables capturing system audio (e.g., from media playback, other apps) and mixing it with microphone audio before transmission. 

Please note that the screen audio is not send if user mutes their microphone since the local audio track is unpublished when muting.

### Implementation

**Screen Audio Capture:**
- Added screen audio capture using `AudioRecord` with `AudioPlaybackCaptureConfiguration` (Android Q+)
- Implemented on-demand audio reading directly from `AudioRecord` using `READ_BLOCKING` mode
- Screen audio is read synchronously when requested by the WebRTC audio callback

**Audio Mixing:**
- Created `addAndConvertBuffers()` utility function to mix microphone and screen audio buffers
- Integrated mixing logic in `StreamPeerConnectionFactory.setAudioBufferCallback()`:
  - Audio filters are applied to microphone audio first
  - Screen audio is then mixed with filtered microphone audio
  - When microphone is muted, screen audio is muted as well
  - When microphone is enabled, both are mixed together

**Audio Track Management:**
- Added `includeAudio` parameter to `Call.startScreenSharing()` (defaults to `false`)
- Added `audioEnabled` StateFlow to `ScreenShareManager` to track screen share audio state separately from video

**Key Changes:**
- `MediaManager.ScreenShareManager`: Added screen audio capture infrastructure and `audioEnabled` StateFlow
- `StreamPeerConnectionFactory`: Added audio mixing logic with proper filter application order
- Demo app: Updated to pass `includeAudio=true` when starting screen sharing

### 🎨 UI Changes

None

### Testing

On the demo app : 
1. Start screen sharing with `includeAudio=true` and verify system audio is captured and transmitted
2. Mute microphone while screen sharing with audio enabled - verify screen audio also stops on the other side
3. Unmute microphone while screen sharing - verify both microphone and screen audio are mixed and transmitted
5. Join call with microphone muted, then start screen sharing with audio - verify audio track is not heard on the other side but once mic is unmuted it is transmitted


demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -928,6 +928,7 @@ private fun BadNetworkLabel(
 
 @Composable
 fun isTablet(): Boolean {
+    return true
```
PratimMallick: This was done only for testing

PratimMallick: Should we remove this check al together in demo app @rahul-lohra @aleksandar-apostolov ?

aleksandar-apostolov: I would love to, but the menu is defined by product and for the public demo screenshare is allowed on tablet.
Revert the isTablet check as it was.
We need to allow screenshare for phones as well on our development app, but not on the public.

aleksandar-apostolov (approved):

PR #1585, opened 2025-12-16
Fix an issue that causes missing video due to track arriving before the participant info has arrived

aleksandar-apostolov:
### Goal

Fix a race condition where video tracks are permanently lost when they arrive before participant information, resulting in users seeing infinite loading spinners instead of video.

### Implementation

Implemented an orphaned tracks mechanism that stores tracks arriving before participants exist, then automatically reconciles them when participant information arrives via any event (JoinResponse, ParticipantJoined, TrackPublished/Unpublished with optional participant field).

### Testing
 Created comprehensive integration tests covering all race condition scenarios including track-before-participant ordering.

## Release Notes

* **New Features**
  * Enhanced large call performance by tracking participant information alongside media stream events.
  * Improved handling of media tracks arriving before participant data is available.

* **Bug Fixes**
  * Fixed stream removal detection to properly clean up audio and video tracks.
  * Enhanced orphaned track cleanup to prevent memory leaks.

* **Tests**
  * Added comprehensive tests for media stream reconciliation and removal workflows.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
@@ -1302,7 +1302,6 @@ public class RtcSession internal constructor(
                         }
 
                         is TrackUnpublishedEvent -> {
-
                             updatePublishState(
                                 userId = event.userId,
                                 sessionId = event.sessionId,


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -307,6 +335,58 @@ public class RtcSession internal constructor(
...
+            }
+        }
+    }
+
+    /**
+     * Retrieves and removes all orphaned tracks for a specific sessionId.
+     * Returns a list of track type to track pairs that can be attached to the participant.
+     */
+    private fun takeOrphanedTracks(sessionId: String): List<Pair<TrackType, MediaTrack>> {
+        return synchronized(orphanedTracks) {
```
PratimMallick: Do we need to syncronize this on orphanedTracks? 

this method is only called from `reconcileOrphanedTracks`  and its getting called from the `handleEvent` method which runs in serial processor

aleksandar-apostolov: Most likely not.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -415,6 +513,19 @@ public class RtcSession internal constructor(
                 setTrack(sessionId, trackType, track)
             }
         }
+
+        // Listen for removed streams to cleanup orphaned tracks
+        coroutineScope.launch {
```
PratimMallick, line 540: This may create problem as we are mutating the orphantracks list from two places -one while collecting added streams and another while collecting removedStreams unless they are synchronized and not running on differnt threads

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -815,6 +926,17 @@ public class RtcSession internal constructor(
...
         publisher = null
 
+        // cleanup orphaned tracks to prevent memory leaks
+        synchronized(orphanedTracks) {
+            if (orphanedTracks.isNotEmpty()) {
+                logger.w {
+                    "[cleanup] #orphaned-track; Clearing ${orphanedTracks.size} orphaned tracks " +
+                        "that were never reconciled"
+                }
+                orphanedTracks.clear()
```
PratimMallick: another way could be having a mutex lock and use that before add(), remove() and clear() of orphanedTracks. That will fix the other issue also

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1135,6 +1281,20 @@ public class RtcSession internal constructor(
...
 
                         is TrackUnpublishedEvent -> {
+                            // In large calls, TrackUnpublishedEvent may include participant info
+                            // instead of sending separate events
+                            if (event.participant != null) {
+                                logger.i {
+                                    "[TrackUnpublishedEvent] #orphaned-track; Participant info included, " +
+                                        "updating participant for sessionId=${event.sessionId}"
+                                }
+                                val participantState = call.state.getOrCreateParticipant(
```
PratimMallick: on calling `getOrCreateParticipant` we also end up calling `upsertParticipants` which adds the participant to `internalParticipants` Map and ultimately updates the `_participants` flow - which updatest he public `participants` flow

I think this is not needed as it may un-necessarily send updates to the UI about a participant getting added whereas its just a trackUnpublishedEvent 

We should send an update event only when we have ParticipantAdded/Removed event

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
```diff
@@ -584,6 +604,69 @@ internal class Subscriber(
         }
     }
 
+    override fun onRemoveStream(stream: MediaStream?) {
```
PratimMallick, line 607: Not sure but I think we may not get this callback from webrtc atall .. have to check 

but why do we need this anyway? We already know of the tracks that were removed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
aleksandar-apostolov: Thats why we had the synchronized blocks :)

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
aleksandar-apostolov: To remove orphaned tracks that never got a participant to go with.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
aleksandar-apostolov: We dont always gey participant update events. When there are lots of participants the updates stop coming and the way to update latest data for the participant is via track publish/unpublish events.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
aleksandar-apostolov: We'll do it with a `ConcurrentHashMap`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
aleksandar-apostolov: Agreed to remove this.

PR #1588, opened 2025-12-23
Improve CallService lifecycle, notification handling, and ringing-to-active transition

rahul-lohra:
### Goal

1. Fix FG Service Notification dismiss from Notification Tray
2. Fix Ringing State update process
3. Reuse same FG Service when transitioning from Ringing to Active Call instead of killing & respawn FG Service
4. Split Call Service into smaller classes

## Core classes
1. ClientState
2. CallState

| Class | Change description |
| --- | --- |
| ClientState | Instead of stopping  restarting service we are now transitioning incoming/outgoing call to active on the same service |
| CallState | - Prevent race-condition in event consumption & processing in Service Class by firing new local events when their original events are consumed <br> -Since we will use same service to transition from ringing to active so we need to store a reference of original notification Id <br> - Process `LocalCallMissedEvent` here so that it will work for calls which are not active |

## Service classes refactoring

| Class | Change description |
| --- | --- |
| ServiceIntentBuilder | Creates and configures the various Intent objects used to communicate with the CallService. |
| ServiceLauncher | Provides a clean and centralized API to start, stop, and interact with the CallService. |
| CallServiceLifecycleManager | Manages the core business logic of starting, ending, and handling the lifecycle of a call. |
| CallServiceNotificationManager | Responsible for creating, displaying, and updating all notifications shown by the service. |
| ServiceState | The single source of truth that holds the current state of the service, such as active and ringing call IDs. |
| CallServiceEventObserver | Listens to VideoEvent changes from the Call and updates the ServiceState accordingly. |
| CallServiceNotificationUpdateObserver | Observes state changes and triggers updates to the foreground notification. |
| CallServiceRingingStateObserver | Manages the ringing state and handles transitions between incoming, active, and idle calls. |

## Util Classes
| Class | Change description |
| --- | --- |
| Debouncer |  To debounce the stopping of FG Service to ensure enough time has passed for the system to process the notification removal. |
| Throttler | Throttling the service by [CallService.SERVICE_DESTROY_THROTTLE_TIME_MS] such that the stop service is invoked once (at least less frequently) |

####  Fix FG Service Notification dismiss from Notification Tray
1. Open the notification drawer in callee phone
2. Make a call and end the call within 1-2 second.
3. The notification should properly update on the callee's phone

#### Service lifecycle
1. When a call is ended, service should get destroyed (check from the logs)

</details>

## Summary by CodeRabbit

* **New Features**
  * Missed-call notifications now include caller/creator info and use stable notification IDs.
  * Notifications and foreground service starts are permission-aware and choose appropriate service types.
  * Debouncing/throttling reduces rapid stop/start service actions for smoother behavior.

* **Bug Fixes**
  * More consistent ringing → accepted transitions across devices.
  * Improved notification lifecycle: better updates, cancellations, and timing when multiple calls occur.

### DIFF
--- a/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceRingingStateObserverTest.kt
+++ b/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceRingingStateObserverTest.kt
@@ -183,6 +183,7 @@ class CallServiceRingingStateObserverTest {
 
     @Test
     fun `rejected by all rejects call and stops service`() = runTest {
+        every { streamVideo.scope } returns testScope
         observer.observe { onStopServiceInvoked.add(Unit) }
         advanceUntilIdle()
 


stream-video-android-core/api/stream-video-android-core.api
```diff
@@ -3558,13 +3558,15 @@ public final class io/getstream/android/video/generated/models/ListTranscription
...
 	public final fun component1 ()Ljava/lang/String;
-	public final fun copy (Ljava/lang/String;)Lio/getstream/android/video/generated/models/LocalCallMissedEvent;
-	public static synthetic fun copy$default (Lio/getstream/android/video/generated/models/LocalCallMissedEvent;Ljava/lang/String;ILjava/lang/Object;)Lio/getstream/android/video/generated/models/LocalCallMissedEvent;
+	public final fun component2 ()Ljava/lang/String;
+	public final fun copy (Ljava/lang/String;Ljava/lang/String;)Lio/getstream/android/video/generated/models/LocalCallMissedEvent;
+	public static synthetic fun copy$default (Lio/getstream/android/video/generated/models/LocalCallMissedEvent;Ljava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Lio/getstream/android/video/generated/models/LocalCallMissedEvent;
 	public fun equals (Ljava/lang/Object;)Z
 	public fun getCallCID ()Ljava/lang/String;
 	public final fun getCallCid ()Ljava/lang/String;
+	public final fun getCreatedById ()Ljava/lang/String;
```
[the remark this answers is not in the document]
rahul-lohra: Ignore this as only we will create new instance of `LocalCallMissedEvent`, our clients will only read the instance

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceRingingStateObserver.kt
```diff
@@ -0,0 +1,149 @@
...
+    private fun handleRejectedByAllState(onStopService: () -> Unit) {
+        ClientScope().launch {
+            call.reject(
+                source = "RingingState.RejectedByAll",
+                reason = RejectReason.Decline,
+            )
+        }
+        soundPlayer?.stopCallSound()
+        onStopService()
+    }
```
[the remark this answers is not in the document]
rahul-lohra: call.reject() has bigger scope than service scope. So this is acceptable

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -192,16 +101,34 @@ internal open class CallService : Service() {
         const val TRIGGER_OUTGOING_CALL = "outgoing_call"
         const val TRIGGER_ONGOING_CALL = "ongoing_call"
         const val EXTRA_STOP_SERVICE = "io.getstream.video.android.core.stop_service"
+
+        const val SERVICE_DESTROY_THRESHOLD_TIME_MS = 2_000L
+        const val SERVICE_DESTROY_THROTTLE_TIME_MS = 1_000L
```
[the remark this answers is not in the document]
rahul-lohra: It is used in `io.getstream.video.android.core.notifications.internal.service.ServiceLaucher.stopService()`

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/managers/CallServiceNotificationManagerTest.kt
```diff
@@ -0,0 +1,195 @@
...
+        every {
+            ContextCompat.checkSelfPermission(any(), any())
+        } returns PackageManager.PERMISSION_GRANTED
+
+        every { StreamVideo.instanceOrNull() } returns null
+
+        sut.justNotify(service, callId, 1001, notification)
+
+        // Should not crash
+    }
```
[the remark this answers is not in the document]
rahul-lohra: Please ignore this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -420,518 +462,159 @@ internal open class CallService : Service() {
...
+            logger.d { "[stopServiceGracefully] differenceInSeconds: $differenceInSeconds" }
+            if (differenceInSeconds >= debouncerThresholdTimeInSeconds) {
+                internalStopServiceGracefully()
+            } else {
+                debouncer.submit(debouncerThresholdTimeInSeconds) {
+                    internalStopServiceGracefully()
+                }
+            }
         }
+    }
```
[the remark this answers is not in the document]
rahul-lohra: Actually, the intent is to prevent the service from being stopped if it was started only 1–2 seconds earlier. This is controlled by `SERVICE_DESTROY_THRESHOLD_TIME_MS`, which is currently hardcoded to a very small value (around 1–2 seconds).

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -420,518 +462,159 @@ internal open class CallService : Service() {
...
-        // Call sounds
         /**
          * Temp Fix!! The observeRingingState scope was getting cancelled and as a result,
          * ringing state was not properly updated
          */
-        callSoundAndVibrationPlayer?.stopCallSound()
-
-        // Stop any jobs
+        serviceState.soundPlayer?.stopCallSound() // TODO should check which call owns the sound
         serviceScope.cancel()
```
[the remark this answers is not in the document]
rahul-lohra: This is legacy logic with an older comment, so I’ve left it unchanged.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/AndroidUtils.kt
```diff
@@ -257,3 +257,9 @@ internal fun isAppInForeground(): Boolean {
         false // fallback if lifecycle isn't initialized yet
     }
 }
+
+internal fun debugPrintLastStackFrames(tag: String, count: Int = 10) {
+    val stack = Thread.currentThread().stackTrace
+    val message = stack.takeLast(count).joinToString("\n")
+    Log.d(tag, message)
+}
```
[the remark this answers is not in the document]
rahul-lohra: It will be used in debugging only so I left it untouched

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/Throttler.kt
```diff
@@ -0,0 +1,84 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.notifications.internal
+
+import io.getstream.log.taggedLogger
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.getValue
+
+internal object Throttler {
```
aleksandar-apostolov: This is a singleton with a map that grows indefinitely. Keys are never removed except manually via reset(). If different call IDs or dynamic keys are used, this will leak memory over time. 
Is it necessary for this to be `object`? Can we create instance of this?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/events/LocalEvent.kt
```diff
@@ -35,7 +35,7 @@ abstract class LocalEvent : WSCallEvent, VideoEvent()
  * allowing the SDK to handle it as if a [io.getstream.android.video.generated.models.CallRejectedEvent] had
  * been received in real time (or to apply adjusted logic if needed).
  */
-data class LocalCallMissedEvent(val callCid: String) : LocalEvent() {
+data class LocalCallMissedEvent(val createdById: String, val callCid: String) : LocalEvent() {
```
aleksandar-apostolov: Breaking change, since its a public class we need to address it, add sensible default.
Q: Why is this public?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -420,518 +462,159 @@ internal open class CallService : Service() {
...
         logger.d { "[showIncomingCall] notificationId: $notificationId" }
         val hasActiveCall = StreamVideo.instanceOrNull()?.state?.activeCall?.value != null
-        logger.d { "[showIncomingCall] hasActiveCall: $hasActiveCall" }
-
-        if (!hasActiveCall) { // If there isn't another call in progress
-            // The service was started with startForegroundService() (from companion object), so we need to call startForeground().
-            logger.d { "[showIncomingCall] Starting foreground service with notification" }
 
+        if (!hasActiveCall) {
             StreamVideo.instanceOrNull()?.call(callId.type, callId.id)
```
aleksandar-apostolov: StreamVideo.instanceOrNull() is called twice without caching the result. The instance could become null (e.g., app cleanup).

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceEventObserver.kt
```diff
@@ -0,0 +1,155 @@
...
+    fun observe(onServiceStop: () -> Unit, onRemoveIncoming: () -> Unit) {
+        observeCallEvents(onServiceStop, onRemoveIncoming)
+        observeConnectionState(onServiceStop)
+    }
+
+    /**
+     * Observes call events (accepted, rejected, ended, missed).
+     */
+    private fun observeCallEvents(onServiceStop: () -> Unit, onRemoveIncoming: () -> Unit) {
+        scope.launch {
```
aleksandar-apostolov: Different scope used, is this intentional? (see line 137)

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    private fun calculateServiceType(context: Context): Int {
+        return if (hasAllPermissions(context)) {
+            allPermissionsServiceType()
+        } else {
+            noPermissionServiceType()
+        }
+    }
```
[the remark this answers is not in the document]
aleksandar-apostolov: +1 on this.
 On devices running Android Q (API 29) or lower, calling hasAllPermissions() violates the @RequiresApi(R)

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/models/ServiceState.kt
```diff
@@ -0,0 +1,66 @@
...
+import android.app.Service
+import android.content.Intent
+import android.content.IntentFilter
+import io.getstream.video.android.core.notifications.internal.receivers.ToggleCameraBroadcastReceiver
+import io.getstream.video.android.core.sounds.CallSoundAndVibrationPlayer
+import io.getstream.video.android.model.StreamCallId
+import kotlinx.coroutines.CoroutineScope
+import org.threeten.bp.OffsetDateTime
+
+internal class ServiceState {
```
aleksandar-apostolov: ServiceState properties are accessed from multiple threads (main thread, IO dispatcher, observer coroutines, Handler) without synchronization, risking race conditions and visibility issues.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/Debouncer.kt
```diff
@@ -0,0 +1,40 @@
...
+        runnable?.let { handler.removeCallbacks(it) }
+
+        runnable = Runnable { action() }
+        runnable?.let {
+            handler.postDelayed(it, delayMs)
+        }
+    }
+
+    fun cancel() {
+        runnable?.let { handler.removeCallbacks(it) }
```
aleksandar-apostolov, line 37: removeCallbacks() only removes if not already executing. If the callback is mid-execution when cancel is called, it continues. Is there a way to cancel the handler?

aleksandar-apostolov: Or maybe we should handle this in the callsite.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/managers/CallServiceNotificationManager.kt
```diff
@@ -0,0 +1,78 @@
...
+        if (ActivityCompat.checkSelfPermission(
+                service, Manifest.permission.POST_NOTIFICATIONS,
+            ) == PackageManager.PERMISSION_GRANTED
+        ) {
+            StreamVideo.Companion.instanceOrNull()?.getStreamNotificationDispatcher()
+                ?.notify(callId, notificationId, notification)
+        }
+    }
+
+    fun cancelNotifications(service: Service, callId: StreamCallId) {
```
aleksandar-apostolov, line 65: Only cancels 2 notification IDs:
  1. `callId.hashCode()`
  2. `call.state.notificationId` (if available)

  But notifications can be posted with:
  - `callId.getNotificationId(NotificationType.Incoming)`
  - `callId.getNotificationId(NotificationType.Outgoing)`
  - `callId.getNotificationId(NotificationType.Ongoing)`
  
  Should we handle it here?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/managers/CallServiceLifecycleManager.kt
```diff
@@ -0,0 +1,107 @@
...
+            val call = streamVideo.call(callId.type, callId.id)
+            val update = call.get()
+
+            if (update.isFailure) {
+                onError()
+                return@launch
+            }
+        }
+
+        scope.launch {
```
aleksandar-apostolov, line 45: why is this run in separate coroutine? what if connect fails here, should we not call onError?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceRingingStateObserver.kt
```diff
@@ -0,0 +1,149 @@
...
+     */
+    private fun handleActiveState() {
+        soundPlayer?.stopCallSound()
+    }
+
+    /**
+     * Handles rejected by all state - rejects call and stops service.
+     */
+    private fun handleRejectedByAllState(onStopService: () -> Unit) {
+        ClientScope().launch {
```
aleksandar-apostolov: Why do we create new scope here but we already have `scope` in the constructor of this object?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
```diff
@@ -247,155 +173,271 @@ internal open class CallService : Service() {
...
+        maybeHandleMediaIntent(intent, params.callId)
 
-        logger.i {
-            "[onStartCommand]. callId: ${intentCallId?.id}, trigger: $trigger, Callservice hashcode: ${hashCode()}"
-        }
-        logger.d {
-            "[onStartCommand] streamVideo: ${streamVideo != null}, intentCallId: ${intentCallId != null}, trigger: $trigger"
+        val call = params.streamVideo.call(params.callId.type, params.callId.id)
+
+        if (!verifyPermissions(params.streamVideo, call, params.callId, params.trigger)) {
```
aleksandar-apostolov, line 213: Returning START_REDELIVER_INTENT on permission failure causes infinite retry loop; should we return START_NOT_STICKY instead?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/Throttler.kt
rahul-lohra: Thanks for pointing this out 👍. It will be replaced with `class`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/events/LocalEvent.kt
rahul-lohra: Yes we should add `default`

> Why is this public?

It is used in `StreamCallActivity` to detect when to perform `call.leave()` based on Video Events

```kotlin
class StreamCallActivity { ...
    public open fun onCallEvent(call: Call, event: VideoEvent) {
        when (event) {
            is CallEndedEvent, is CallEndedSfuEvent, is CallSessionEndedEvent, is LocalCallMissedEvent -> {
                leave(call, onSuccess = onSuccessFinish, onError = onErrorFinish)
            }
....
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/models/ServiceState.kt
rahul-lohra: Yes it needs to be thread safe, thanks for pointing this critical issue.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/Debouncer.kt
rahul-lohra: Unfortunately we cannot cancel a mid-execution runnable. Android framework do not allow it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
rahul-lohra: It's not a problem in this function at this moment but sure we can cache the result.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/managers/CallServiceNotificationManager.kt
rahul-lohra: Cancelling the two notification IDs is sufficient in this case.

* `callId.hashCode()` is used for **livestream-related notifications**
* `call.state.notificationId` is used for **ringing notifications**

Regarding the additional IDs:

```kotlin
callId.getNotificationId(NotificationType.Incoming)
callId.getNotificationId(NotificationType.Outgoing)
callId.getNotificationId(NotificationType.Ongoing)
```

Once a ringing notification (incoming / outgoing / ongoing) is dispatched, the SDK stores the corresponding `notificationId` in `call.state.notificationId`. As a result, explicitly cancelling these derived IDs here would be redundant, since they ultimately resolve to the same identifier already handled via `call.state.notificationId`.

Including them would therefore lead to **duplicate cancellation attempts**, which I am intentionally avoiding to keep the logic minimal and deterministic.

---

> If this assumption changes in the future (i.e., notification IDs are no longer persisted in `call.state`), we can revisit this and expand the cancellation logic accordingly.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceRingingStateObserver.kt
rahul-lohra: Actually, `call.reject()` should outlive both the call scope and the service scope and therefore needs to be executed in the **client-level scope**. The call scope or service scope may be cancelled or finished while the network request triggered by `call.reject()` is still in flight, and tying it to those shorter-lived scopes could cause the operation to be prematurely cancelled.

In future I am thinking to write this logic inside `call.reject()` function

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt
rahul-lohra: Yes `START_REDELIVER_INTENT`  does not make any sense here. We can change it to `START_NOT_STICKY`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceRingingStateObserver.kt
aleksandar-apostolov: But in this case should we not get the scope from the `StreamVideoImpl` ?

rahul-lohra: Yeah we can use that. I will change the code

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/models/ServiceState.kt
rahul-lohra: Made the service state _(including its getter & setters)_ thread-safe by encapsulating all service-related variables within `ServiceStateSnapshot` and exposing them via a `StateFlow`.

```kotlin
internal data class ServiceStateSnapshot(
    val currentCallId: StreamCallId? = null,
    val notificationId: Int? = null,
    val soundPlayer: CallSoundAndVibrationPlayer? = null,
    val toggleCameraBroadcastReceiver: ToggleCameraBroadcastReceiver? = null,
    val isReceiverRegistered: Boolean = false,
    val startTime: OffsetDateTime? = null,
)

internal class ServiceStateController {
    private val _state = MutableStateFlow(ServiceStateSnapshot())
    val state: StateFlow<ServiceStateSnapshot> = _state
    
    // The getters & setters
     val currentCallId: StreamCallId?
        get() = state.value.currentCallId
        
     fun setCurrentCallId(callId: StreamCallId) {
       _state.update { it.copy(currentCallId = callId) }
    }
}
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceEventObserver.kt
rahul-lohra: Corrected it with same scope. Thank you

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/managers/CallServiceLifecycleManager.kt
rahul-lohra: This function is legacy code [Ref](https://github.com/GetStream/stream-video-android/blob/487e7f3b9c823b6bc23d88e7ce3c1f3c6638dbb2/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/CallService.kt#L529). So I think it is safe for this PR. 

But we need to revisit this logic and properly collect errors and take action. At the moment `connectIfNotAlreadyConnected` returns `Unit` instead of `Result`. So the error is swallowed inside

PR #1591, opened 2025-12-24
Replace Twilio's AudioSwitch library with a custom implementation

PratimMallick:
### Goal

This PR replaces Twilio's AudioSwitch library with a custom implementation built on native Android AudioManager APIs. The new implementation maintains full API compatibility while providing better control over audio device management and reducing external dependencies.

### Core Implementation
- **New `StreamAudioSwitch` class**: Custom audio switch implementation using Android's `AudioManager` APIs
- **Audio device managers**:
    - `ModernAudioDeviceManager` for Android API 23+ (uses `AudioManager.setCommunicationDevice()`)
    - `LegacyAudioDeviceManager` for older Android versions (uses `AudioManager.setMode()` and routing APIs)
- **Audio handler system**: Factory pattern with `AudioHandlerFactory` to create appropriate handlers based on Android version
- **Device enumeration**: Native Android device discovery using `AudioManager.getDevices()`

### API Compatibility
- Maintained backward compatibility by keeping the deprecated `audio` property in `StreamAudioDevice`
- Added `audioDeviceInfo` property for native Android device information
- Updated dependency visibility: Changed Twilio AudioSwitch from `implementation` to `api` to support consumers still using the deprecated property

## Migration Notes

- No breaking changes - existing APIs remain functional
- The `audio` property on `StreamAudioDevice` is deprecated in favor of `audioDeviceInfo`
- Consumers can continue using existing code, but should migrate to `audioDeviceInfo` for future compatibility

### 🎨 UI Changes
None

### Testing

Use BT devices, wired headsets to test android on devices android 12+ and pre android 12(Legacy)

## Summary by CodeRabbit

* **New Features**
  * Added configurable built-in audio switch functionality for enhanced audio device management.
  * Improved audio device selection mechanism with better Bluetooth headset and wired device detection.

* **Improvements**
  * Optimized device enumeration across Android API levels (API 24–31+) for more reliable audio routing.
  * Updated Bluetooth permissions to support modern connectivity requirements.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitch.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitch.kt
@@ -46,6 +46,7 @@ internal typealias StreamAudioDeviceChangeListener = (
 internal class StreamAudioSwitch(
     private val context: Context,
     preferredDeviceList: List<Class<out StreamAudioDevice>>? = null,
+    private val onDeviceSelected: ((StreamAudioDevice) -> Unit)? = null,
 ) {
     private val logger by taggedLogger(TAG)
 
@@ -235,6 +236,7 @@ internal class StreamAudioSwitch(
                     if (success) {
                         _selectedDeviceState.value = deviceToSelect
                         logger.d { "[selectNativeDevice] Native device selected: $deviceToSelect" }
+                        traceDeviceSelected(deviceToSelect)
                     } else {
                         logger.w { "[selectNativeDevice] Failed to select native device: $deviceToSelect" }
                         // Fallback to automatic selection if the requested device is not available
@@ -326,7 +328,12 @@ internal class StreamAudioSwitch(
             // Update from manager
             val managerSelected = manager?.getSelectedDevice()
             if (managerSelected != null) {
+                val hasChanged = currentSelected == null ||
+                    !isSameAudioDevice(managerSelected, currentSelected, nativeDevices)
                 _selectedDeviceState.value = managerSelected
+                if (hasChanged) {
+                    traceDeviceSelected(managerSelected)
+                }
             }
         }
 
@@ -347,6 +354,14 @@ internal class StreamAudioSwitch(
         }
 
         return true
+    }
+
+    private fun traceDeviceSelected(device: StreamAudioDevice) {
+        try {
+            onDeviceSelected?.invoke(device)
+        } catch (e: Exception) {
+            logger.e(e) { "[traceDeviceSelected] Failed to emit trace: ${e.message}" }
+        }
     }
 
     /**
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitchHandler.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitchHandler.kt
@@ -30,6 +30,7 @@ internal class StreamAudioSwitchHandler(
     private val context: Context,
     private val preferredDeviceList: List<Class<out StreamAudioDevice>>,
     private var audioDeviceChangeListener: StreamAudioDeviceChangeListener,
+    private val onDeviceSelected: ((StreamAudioDevice) -> Unit)? = null,
 ) : AudioHandler {
 
     private val logger by taggedLogger(TAG)
@@ -51,6 +52,7 @@ internal class StreamAudioSwitchHandler(
                     val switch = StreamAudioSwitch(
                         context = context,
                         preferredDeviceList = preferredDeviceList,
+                        onDeviceSelected = onDeviceSelected,
                     )
                     streamAudioSwitch = switch
                     switch.start(audioDeviceChangeListener)
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
@@ -766,6 +766,16 @@ class MicrophoneManager(
                         capturedOnAudioDevicesUpdate?.invoke()
                         capturedOnAudioDevicesUpdate = null
                     },
+                    onDeviceSelected = { device ->
+                        mediaManager.call.session?.sfuTracer?.trace(
+                            "audio-device-selected",
+                            mapOf(
+                                "device" to device.name,
+                                "type" to device::class.simpleName,
+                                "id" to device.audioDeviceInfo?.id,
+                            ),
+                        )
+                    },
                 )
 
                 logger.d { "[setup] Calling start on instance $audioHandler" }


stream-video-android-core/api/stream-video-android-core.api
```diff
@@ -8503,85 +8503,79 @@ public abstract interface class io/getstream/video/android/core/api/SignalServer
...
-}
-
-public final class io/getstream/video/android/core/audio/AudioSwitchHandler$Companion {
-}
-
 public abstract class io/getstream/video/android/core/audio/StreamAudioDevice {
 	public static final field Companion Lio/getstream/video/android/core/audio/StreamAudioDevice$Companion;
-	public static final fun fromAudio (Lcom/twilio/audioswitch/AudioDevice;)Lio/getstream/video/android/core/audio/StreamAudioDevice;
-	public abstract fun getAudio ()Lcom/twilio/audioswitch/AudioDevice;
+	public static final fun fromAudioDeviceInfo (Landroid/media/AudioDeviceInfo;)Lio/getstream/video/android/core/audio/StreamAudioDevice;
```
rahul-lohra, line 8513: I can see several breaking changes. If we have decided to move forward with this approach, I suggest we draft a migration document to help developers transition to the newer APIs.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitch.kt
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    private fun unregisterDeviceCallback() {
+        audioDeviceCallback?.let { callback ->
+            StreamAudioManager.unregisterAudioDeviceCallback(audioManager, callback)
+            audioDeviceCallback = null
+        }
+    }
+
+    private fun enumerateDevices() {
```
aleksandar-apostolov, line 311: It seems we treat the selected device as removed when its StreamAudioDevice instance is missing from _availableDevices, but the equality includes the AudioDeviceInfo reference. On API 31+ the framework can return new AudioDeviceInfo objects for the same physical route, so this block clears the selection and calls clearDevice() even though nothing changed. Could we match by a stable key (for example device type plus AudioDeviceInfo.id) so the selection persists when the OS hands back a different object instance?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioManager.kt
```diff
@@ -0,0 +1,136 @@
...
+            null
+        }
+    }
+
+    /**
+     * Sets the communication device
+     * For API 31+: Uses setCommunicationDevice()
+     * For API < 31: Returns false (use setDeviceLegacy in StreamAudioSwitch instead)
+     */
+    @RequiresApi(Build.VERSION_CODES.S)
```
aleksandar-apostolov, line 106: setCommunicationDevice can return false when the OS rejects the route change, but we always return true so callers think it succeeded and update state anyway. Could we propagate the actual result (and maybe log failures) so ModernAudioDeviceManager knows when routing didn’t happen?

aleksandar-apostolov: UPDATE: I see `coderabbitai` also noticed this

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/SettingsMenu.kt
```diff
@@ -216,11 +216,45 @@ internal fun SettingsMenu(
             is StreamAudioDevice.Speakerphone -> Icons.Default.SpeakerPhone
             is StreamAudioDevice.WiredHeadset -> Icons.Default.HeadsetMic
         }
+        // Compare devices by type and audioDeviceInfo ID (if available) since audio can be null when using custom audio switch
```
aleksandar-apostolov, line 219: We now compute isSelected with type/ID fallbacks, but the value passed into AudioDeviceUiState still compares audioDeviceInfo?.id directly. For devices coming from the legacy manager the ID is null, so the highlight never turns on even though isSelected is true. Could we pass the computed flag down so the UI reflects the actual selection?

stream-video-android-core/api/stream-video-android-core.api
aleksandar-apostolov: +1

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitch.kt
```diff
@@ -0,0 +1,447 @@
...
+                logger.d { "[start] Using LegacyAudioDeviceManager (API 24-30)" }
+                LegacyAudioDeviceManager(
+                    context = context,
+                    audioManager = audioManager,
+                    onDeviceChange = { enumerateDevices() },
+                    onBluetoothConnectionFailure = { handleBluetoothConnectionFailure() },
+                )
+            }
+
+            // Request audio focus for voice communication
```
aleksandar-apostolov, line 127: Previously AudioSwitchHandler called AudioSwitch.activate(), which set audioManager.mode = MODE_IN_COMMUNICATION and restored it on stop. In this replacement we never touch audioManager.mode (there are no references anywhere in the module), so we leave the stack in MODE_NORMAL while trying to switch between earpiece/speaker. On many OEMs that means the earpiece route is ignored or uses the wrong audio processing. Could we set MODE_IN_COMMUNICATION in start() and reset it in stop() to match the old behaviour?

stream-video-android-core/src/main/AndroidManifest.xml
```diff
@@ -26,7 +26,9 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
+    <!-- Bluetooth permissions: BLUETOOTH for API < 31, BLUETOOTH_CONNECT for API 31+ -->
```
aleksandar-apostolov: We now comment out BLUETOOTH_CONNECT entirely, but the Compose permission helpers still call rememberCallPermissionsState with Manifest.permission.BLUETOOTH_CONNECT (stream-video-android-ui-compose/.../SinglePermission.kt). Requesting a permission that isn't declared throws IllegalArgumentException on Android 12+, so any app using that composable will crash unless they re-declare the permission themselves. Can we keep the manifest entry (or drop the runtime request) so the helper continues to work out of the box?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioDevice.kt
```diff
@@ -16,53 +16,161 @@
...
-    abstract val audio: AudioDevice
+    /**
+     * The Android AudioDeviceInfo instance.
+     * This provides device identification and capabilities when using native Android audio management.
+     * @see android.media.AudioDeviceInfo
+     */
+    abstract val audioDeviceInfo: AudioDeviceInfo?
 
-    /** An [StreamAudioDevice] representing a Bluetooth Headset.*/
+    /** A [StreamAudioDevice] representing a Bluetooth Headset.*/
```
aleksandar-apostolov, line 41: When we convert an AudioDeviceInfo to a StreamAudioDevice we keep the default labels ("Bluetooth", "Wired Headset", etc.). On API 31+ the OS gives us the real hardware name via AudioDeviceInfo.productName, but we drop it here, so every Bluetooth device shows up with the same generic label. Twilio’s AudioSwitch surfaced the actual device name so the user could pick the right headset—can we populate name from productName to preserve that UX?

stream-video-android-core/api/stream-video-android-core.api
PratimMallick: Yup we have breaking changes, and thats why we are making this part of the major version. Will add a migration doc for this. s

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioDevice.kt
PratimMallick: AudioDeviceInfo has the human readable name . It can be accessed by calling `getProductName()` on the `audioDeviceInfo`

PratimMallick: Earlier this was present inside the `AudioDevice` object which was sent by twilio

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitchHandler.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    override fun stop() {
+        logger.d { "[stop] no args" }
+        mainThreadHandler.removeCallbacksAndMessages(null)
+        mainThreadHandler.post {
+            streamAudioSwitch?.stop()
+            streamAudioSwitch = null
+        }
+    }
```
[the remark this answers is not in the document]
PratimMallick: Good catch. fixining it

stream-video-android-core/src/main/AndroidManifest.xml
PratimMallick: The problem with keeping BLUETOOTH_CONNECT in the manifest is that this permission shows up in the system settings of the app when someone checks Settings → App → Permissions 

Can we not modify the Compose permission helpers accordingly?

PratimMallick: Good you pointed it out. 

Removed all BLUETOOTH_CONNECT references across code, tests, and manifests, and removed the Compose rememberBluetoothPermissionState

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/SettingsMenu.kt
PratimMallick: Yeah good point. Need to use `isSelected`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitch.kt
PratimMallick: wired MODE_IN_COMMUNICATION into StreamAudioSwitch.start() and restore it in stop() so the global audio mode is only active while the switch is running.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/StreamAudioSwitch.kt
PratimMallick: Updated the matcher to do ID comparison only for Bluetooth, and type-only matching for all other devices.

PratimMallick: yes fixed this

PR #1595, opened 2026-01-05
Introduce `connectOnInit` flag to control initial WebSocket connection

rahul-lohra:
### Goal

To prevent the SDK from hitting API rate limits by changing the default connection behavior. Currently, the SDK automatically establishes a WebSocket connection upon initialization. This can lead to excessive, unnecessary connections for clients who initialize the SDK at startup but do not immediately use video features, causing them to hit rate limits.

### Implementation

To fix this, this PR introduces a `connectOnInit` flag in the `StreamVideoBuilder`. By setting `connectOnInit = false`, developers can now defer the connection until it's explicitly required. The SDK will still automatically connect internally when a core action is performed (like starting a call), ensuring functionality remains seamless while giving developers the control to reduce unnecessary startup connections.

### Example Usage

```kotlin
 StreamVideoBuilder(connectOnInit = false).build()
 
 //To manually connect to websocket
 StreamVideo.instanceOrNull()?.connect()
```

### 🎨 UI Changes

None

### Testing

Just perform audio/video call and check if you see cooridnator socket events

## Summary by CodeRabbit

* **New Features**
  * Added `connectOnInit` parameter to control automatic WebSocket connection behavior during app startup (enabled by default for backward compatibility)
  * Introduced a new suspend function for establishing connections, providing a synchronous-style alternative to existing asynchronous methods with structured error handling

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
@@ -22,6 +22,7 @@ import io.getstream.video.android.core.BuildConfig
 import io.getstream.video.android.core.StreamVideo
 import io.getstream.video.android.core.StreamVideoClient
 import io.getstream.video.android.core.internal.InternalStreamVideoApi
+import java.text.Normalizer
 
 @InternalStreamVideoApi
 class HeadersUtil {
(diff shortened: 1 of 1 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -146,6 +148,7 @@ public class StreamVideoBuilder @JvmOverloads constructor(
     @InternalStreamVideoApi
     private val enableStereoForSubscriber: Boolean = true,
     private val telecomConfig: TelecomConfig? = null,
+    private val connectOnInit: Boolean = true,
```
PratimMallick, line 155: Can we wait till the major release, so that we can make it permanently connectOnInit = false

PR #1605, opened 2026-01-26
Make Call object reusable after leave()

aleksandar-apostolov:
### Goal

Make `Call` objects reusable after `leave()` so integrations that cache Call objects can rejoin without creating new instances. Currently, calling `leave()` cancels coroutine scopes making the Call unusable.

### Implementation

- Add `resetScopes()` method called after cleanup to reset state for rejoin:
  - Generate new session IDs for fresh RTC connections
  - Clear participants to remove stale video/audio tracks
  - Reset device statuses to allow media re-initialization
  - Reset `ScopeProvider` to allow executor recreation
- Track `cleanupJob` to prevent race conditions between leave cleanup and rapid rejoin
- Keep Call in client's `calls` map after leave (remove `destroyedCalls` cache)
- Make `MediaManager.scope` a dynamic getter from `call.scope`
- Maintain API compatibility with deprecated `scope` parameter in `MediaManagerImpl`

### Testing

- [ ] Join a call, leave, then rejoin using the same Call object - verify audio/video works
- [ ] Verify CallState (participants, duration, etc.) continues to update after rejoin
- [ ] Verify no race conditions between leave cleanup and rapid rejoin
- [ ] Run `./gradlew apiCheck` to confirm no breaking API changes
- [ ] Test with integrations that cache Call objects


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1504,21 +1507,47 @@ public class Call(
...
-        scope.cancel()
+    /**
+     * Resets state to allow the Call to be reusable after leave().
+     * Generates new session IDs, resets the scopeProvider, clears participants, and resets device statuses.
+     *
+     * IMPORTANT: We do NOT recreate [scope] or [supervisorJob] because [CallState] and its
+     * StateFlows depend on the original scope. The scope lives for the entire lifetime of
+     * the Call object.
+     */
+    private fun resetScopes() {
```
rahul-lohra, line 1574: I think we should use `mutex` to re-init `scopes` and update `isDestroyed` to prevent race-condition

aleksandar-apostolov: Yes, we should

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1512,21 +1556,50 @@ public class Call(
...
-        UserScope(ClientScope()).launch {
-            supervisorJob.children.forEach { it.join() }
-            supervisorJob.cancel()
+    /**
+     * Resets state to allow the Call to be reusable after leave().
+     * Generates new session IDs, resets the scopeProvider, clears participants, and resets device statuses.
+     *
+     * IMPORTANT: We do NOT recreate [scope] or [supervisorJob] because [CallState] and its
+     * StateFlows depend on the original scope. The scope lives for the entire lifetime of
+     * the Call object.
```
rahul-lohra, line 1572: Risk Analysis: Not cancelling call.scope in call.leave()
The problem
We want to make the Call class reusable after leave(). But scope.cancel() in shutDownJobsGracefully() cancels all Flow.stateIn() operators — they cannot be restarted.

If we remove scope.cancel(), we introduce risks in already stable code.

Risk: Zombie coroutines
call.scope is not just used in Call.kt. It is passed down to multiple components that internally launch their own coroutines:


If scope is not cancelled, we need to micromanage every usage of scope running after leave() — Coordinator Event processing/SFU event processing, SFU event listeners, ICE monitoring, stats reporting, audio processing, and more. I am aware we are handling them but introducing micromanag does not not appear scalable
These are zombie operations on a dead session.

Approaches to solve this
Approach 1: Manually cancel each scope.launch
Go through every scope.launch{} usage and cancel them individually in leave().

Downside: Fragile. Easy to miss one. Every new feature that adds a scope.launch needs to remember to add cancellation. Not scalable.

Approach 2: Create a child scope for session work, keep scope for stateIn
Create a new sessionScope (child of scope). Pass it to all components. Cancel only sessionScope in leave(). stateIn flows stay on scope and survive.

Downside: We need to audit every usage of scope and decide — does this go on scope or sessionScope? High risk of miscategorizing. Changes the cancellation behavior of existing stable code.

Approach 3: Create a new scope just for stateIn flows ✅
Keep scope exactly as-is. Create a new stateScope only for the stateIn calls that need to survive.

scope — untouched. Cancelled in leave(). All existing scope.launch code works exactly as today.
stateScope — new. Only the ~6 stateIn calls move here.
Benefit: Existing code path has zero behavior change. If we miss moving a stateIn, it stays on scope and gets cancelled like today — safe default.

Trade-off: stateScope stays alive after leave(). But it only holds StateFlows with their last emitted value — no active work, no upstream collection, no side effects. For non-reusable calls, we can cancel it in cleanup().

Approach 3b: Replace stateIn with a restartable operator ✅
Instead of managing scope lifetimes, replace stateIn with RestartableStateFlow + RestartableScope:

RestartableScope — wraps a coroutine scope. Can be cancelled and restarted.
RestartableStateFlow — registers itself with RestartableScope. On restart, automatically re-subscribes to upstream. Preserves last emitted value across cancel/restart.
Benefit:

shutDownJobsGracefully() still cancels everything (including RestartableScope) — existing behavior preserved
On join(), calling restartableScope.restart() brings all stateIn flows back to life automatically
Change surface is small — only the ~6 stateIn call sites change
No new scope that lives forever
Trade-off: Introduces a new abstraction (RestartableStateFlow). Needs to be well-tested.

Recommendation
I'd lean towards Approach 3b (restartable operator). It keeps the existing cancel-everything behavior intact, isolates the change to a small surface area, and solves the reuse problem cleanly.

Happy to discuss implementation details in a separate thread.

aleksandar-apostolov: Approach 3b cannot work because the state flows need to be observable even if the call is joined/left multiple times even if you are not in a session you still need to be able to get updates via `call.get()` or `client.queryCalls`.

PR #1617, opened 2026-02-13
Improve ringing state transition from Outgoing to Active in joinAndRing flow

rahul-lohra:
### Goal
Improve ringing state transition from Outgoing to Active in `joinAndRing` flow

When initiating an outgoing call via `joinAndRing`, the call gets stuck in `RingingState.Outgoing` and never transitions to `RingingState.Active` after the callee accepts the call. This was traced to two independent race conditions and an unrelated readability issue with a poorly named flag.

We have 2 major issues
**Issue 1: _acceptedBy is cleared by CallSessionStartedEvent**

`_acceptedBy` gets overwritten during `CallSessionStartedEvent` handling via `updateFromResponse`, even though a prior `CallAcceptedEvent` had already populated it.

```kotlin
fun updateFromResponse(response: CallResponse) {        
        _acceptedBy.value = response.session?.acceptedBy?.keys?.toSet() ?: emptySet()
        ...
    }
```    
**Issue 2: Failure to transition to RingingState.Active**

The caller fails to transition from `RingingState.Outgoing` to `RingingState.Active` because `ringingCall` is being set to null prematurely.

The following logic is responsible for transitioning to RingingState.Active during joinAndRing:

```kotlin
else if (hasRingingCall && createdBy?.id == client.userId) {
            if (acceptedBy.isEmpty()) {
                ...
            } else if (!userIsParticipant) {
                ...
            } else {
                isJoinAndRingInProgress.set(false) // This is where we to the transition to `RingingState.Active`
                cancelTimeout()
                RingingState.Active
            }
        } 
```
However, `hasRingingCall` becomes `false` because `ringingCall` is cleared when the `join()` operation completes:

```kotlin
fun setActiveCall(call: Call) {
        when (ringingState) {
          ...
            is RingingState.Outgoing -> {
                    transitionToAcceptCall(call) // transitionToAcceptCall() clears client.state.ringing,
            // which should not happen during joinAndRing
                call.scope.launch {
                    delay(serviceTransitionDelayMs)
                    maybeStartForegroundService(call, CallService.TRIGGER_ONGOING_CALL)
                }
            }
            ...      
```
    
### Implementation

1. Accumulate `_acceptedBy` from video events instead of overwriting it with session snapshots.
2. Prevent `transitionToAcceptCall` from clearing `_ringingCall` during `joinAndRing`, by guarding on `isJoinAndRingInProgress`.
3. Rename `ringingStateUpdatesStopped` to `isJoinAndRingInProgress` to better reflect its purpose and reduce ambiguity.

### 🎨 UI Changes

None

### Testing

Make 1-to-1 calls and verify that the caller transitions correctly to RingingState.Active.
This issue was caused by a race condition, which makes it difficult to reproduce reliably.

## Summary by CodeRabbit

* **New Features**
  * Enhanced checkbox styling in the direct call screen with improved visual customization.

* **Improvements**
  * Updated default behavior: caller now joins first by default.
  * Improved reliability of incoming call state handling and transitions.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1239,6 +1239,7 @@ public class CallState(
     }
 
     private fun updateRingingState(rejectReason: RejectReason? = null) {
+        val ringingStateLogger by taggedLogger("RingingState")
```
aleksandar-apostolov: `ringingStateLogger` is created inside `updateRingingState()` via delegate — this allocates a new logger on every call. This method fires frequently during call state changes. Should be a class-level property like the existing `logger`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1268,19 +1270,19 @@ public class CallState(
...
+        ringingStateLogger.d { "call_id: ${call.cid}, Flags: $ringingStateLogs" }
 
         // no members - call is empty, we can join
-        val state: RingingState = if (hasActiveCall && !ringingStateUpdatesStopped) {
+        val state: RingingState = if (hasActiveCall && !ringingStateUpdatesStopped.get()) {
             cancelTimeout()
             RingingState.Active
         } else if (isRejectedByMe) {
             call.leave("updateRingingState-rejected-self")
             cancelTimeout()
```
aleksandar-apostolov, line 1308: The `// for joinAndRing` comment is unclear — doesn't explain why this branch exists alongside the first `hasActiveCall` check above.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
rahul-lohra: Correct

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
rahul-lohra: I have removed this code as it was not required. Kindly re-visit the PR details I have added full context for new-changes

aleksandar-apostolov (approved):

PR #1624, opened 2026-02-24
[WIP] Improve Foreground Service orchestration when transitioning fromn outgoing call to active call

rahul-lohra:
This is WIP, please do not review
### Goal

Previously, during the transition from an outgoing call to an active call, the Foreground Service (FGS) was started twice.

When initiating an outgoing call, we started the FGS with a minimal foreground service permission type. Upon transitioning to an active call, we then restarted the FGS with the full foreground service permission types (camera/microphone).

However, the transition from an outgoing call to an active call can occur while the app is in the background, which caused an exception because restarting a Foreground Service is not allowed when the app is in the background.

Solution - 
To resolve this, we now start the Foreground Service only once, at the time of initiating the outgoing call, after all required permissions are granted.

By starting the FGS upfront with all necessary foreground service permission types, we eliminate the need to restart the service during the state transition, thereby avoiding the background execution exception.

image

### Implementation

Key changes
1. Since we want to create an outgoing-call only when we have all the required permission so we introduced a permission checker in code-flow of making an outgoing-call in `StreamCallActivity` 

```kotlin
class StreamCallActivity {

@StreamCallActivityDelicateApi
    override fun create( call, ring, members, onSuccess, onError) {
       //1. Prepare call-capabilities ONLY for outgoing call
       //2. If has call related permission
       //3. If yes, create call
       //4. Do nothing
    }
```
2. Our Permission rely on Call capabilities which is fetched from coordinator endpoing which is `call.get()`
- So we created a local capabilities only for outgoing-call 

3. To create outgoing call capabilites and use it in `StreamCallActivity` - we need a way to know if we are making a video outgoing call or audio outgoing call
- So we must inform `StreamCallActivity` explicitly about this

So we created a param to pass through intent
```kotlin
public fun <T : StreamCallActivity> callIntent(
             ...
            outgoingCallType: OutgoingCallType = OutgoingCallType.Unknown, // new type
        ) {
        ...
        putExtra(EXTRA_OUTGOING_CALL_TYPE, outgoingCallType.name)
        ...
        }
        
private const val EXTRA_OUTGOING_CALL_TYPE: String = "io.getstream.video.extra_outgoing_call_type"

public sealed class OutgoingCallType private constructor(public val name: String) {
            public companion object {
                public fun toOutgoingCallType(name: String): OutgoingCallType {
                    return when (name) {
                        Video.name -> Video
                        Audio.name -> Audio
                        else -> Unknown
                    }
                }
            }
            override fun toString(): String {
                return name
            }
            public object Video : OutgoingCallType("video")
            public object Audio : OutgoingCallType("audio")
            public object Unknown : OutgoingCallType("unknown")
        }
 ```

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

### Testing

_Explain how this change can be tested (or why it can't be tested)_

_Provide a patch below if it is necessary for testing_

<details>

<summary>Provide the patch summary here</summary>

## Summary by CodeRabbit

* **New Features**
  * Added support for audio and video outgoing call type differentiation
  * Enhanced permission management for outgoing calls based on call type

* **Refactor**
  * Improved internal trigger handling system with type-safe enums replacing string-based constants


rahul-lohra:
Going with another implementation https://github.com/GetStream/stream-video-android/pull/1627

PR #1627, opened 2026-03-09
Improve Foreground Service orchestration when transitioning from outgoing call to active call

rahul-lohra:
### Goal

1. Transition from Outgoing call to active call should not crash when launching FG Service
2. Render no permission when permissions are missing in making outgoing call 

### Implementation

1. Transition from Outgoing call to active call should not crash when launching FG Service

Root Cause - Launching FG Service even when the application is in background and ringing state transitions from outgoing to active

Proposed Solution
1. Launch only FG Serive once with trigger type as outgoing-call

Because of this we need to correctly handle the following
2. Correctly update the notification transition from outgoing to active

### 🎨 UI Changes

Just a no permission ui in outgoing-call flow

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

### Testing

Scenario: Outgoing call → background → call accepted
	1.	Make an outgoing call to another user.
	2.	While the call is still in the outgoing state, press the Home button to send the app to the background.
	3.	Accept the call on the other device.

Expected behavior
	•	The caller should transition smoothly from outgoing → active state.
	•	PiP should not be rendered, since it was never shown before the app went to the background.
	•	The Call Service should continue running without interruption.
	•	The notification should update from outgoing → active.

End condition
	•	If either the caller or callee declines/ends the call, the other participant should also leave the call and the session should terminate correctly.

## Summary by CodeRabbit

* **New Features**
  * Added a permission request UI for foreground service permissions during call setup.
  * Enhanced capability management to properly reflect available permissions for outgoing calls.

* **Bug Fixes**
  * Improved foreground service handling to prevent unnecessary service starts and log errors when required services are unavailable.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
@@ -190,7 +190,7 @@ class ClientState(private val client: StreamVideo) {
                 val callServiceConfig = callConfigRegistry.get(call.type)
                 val serviceClass = callServiceConfig.serviceClass
                 val isServiceRunning = ServiceIntentBuilder()
-                    .isServiceRunning(this.client.context, serviceClass::class.java)
+                    .isServiceRunning(this.client.context, serviceClass)
                 if (callServiceConfig.runCallServiceInForeground) {
                     if (!isServiceRunning) {
                         logger.e { "Outgoing call service should already be running" }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
@@ -25,6 +25,7 @@ import io.getstream.log.taggedLogger
 import io.getstream.result.Error
 import io.getstream.video.android.core.internal.InternalStreamVideoApi
 import io.getstream.video.android.core.notifications.internal.service.CallService
+import io.getstream.video.android.core.notifications.internal.service.ServiceIntentBuilder
 import io.getstream.video.android.core.notifications.internal.service.ServiceLauncher
 import io.getstream.video.android.core.notifications.internal.telecom.TelecomIntegrationType
 import io.getstream.video.android.core.socket.coordinator.state.VideoSocketState
@@ -184,9 +185,16 @@ class ClientState(private val client: StreamVideo) {
                 if (!call.state.isJoinAndRingInProgress.get()) {
                     transitionToAcceptCall(call)
                 }
-                call.scope.launch {
-                    delay(serviceTransitionDelayMs)
-                    maybeStartForegroundService(call, CallService.TRIGGER_ONGOING_CALL)
+                // Intentionally skipping maybeStartForegroundService because service should already be started
+                // when initiating outgoing-call
+                val callServiceConfig = callConfigRegistry.get(call.type)
+                val serviceClass = callServiceConfig.serviceClass
+                val isServiceRunning = ServiceIntentBuilder()
+                    .isServiceRunning(this.client.context, serviceClass::class.java)
+                if (callServiceConfig.runCallServiceInForeground) {
+                    if (!isServiceRunning) {
+                        logger.e { "Outgoing call service should already be running" }
+                    }
                 }
             }
             else -> {
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/StreamForegroundPermissionUtil.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/StreamForegroundPermissionUtil.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-video-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.notifications.internal.service
+
+import io.getstream.video.android.core.StreamVideo
+import io.getstream.video.android.core.StreamVideoClient
+import io.getstream.video.android.core.internal.InternalStreamVideoApi
+import io.getstream.video.android.core.notifications.internal.service.permissions.AudioCallPermissionManager
+import io.getstream.video.android.core.notifications.internal.service.permissions.ForegroundServicePermissionManager
+import kotlin.jvm.java
+
+@InternalStreamVideoApi
+public object StreamForegroundPermissionUtil {
+
+    public fun getForegroundPermissionsForCallType(callType: String): Set<Int> {
+        val client = StreamVideo.instanceOrNull() as? StreamVideoClient
+        if (client != null) {
+            val config = client.callServiceConfigRegistry.get(callType)
+            return when (config.serviceClass) {
+                CallService::class.java -> ForegroundServicePermissionManager().requiredForegroundTypes
+                AudioCallService::class.java -> AudioCallPermissionManager().requiredForegroundTypes
+                else -> setOf(-1)
+            }
+        }
+        return setOf(-1)
+    }
+}
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/StreamForegroundPermissionUtil.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/StreamForegroundPermissionUtil.kt
@@ -33,9 +33,9 @@ public object StreamForegroundPermissionUtil {
             return when (config.serviceClass) {
                 CallService::class.java -> ForegroundServicePermissionManager().requiredForegroundTypes
                 AudioCallService::class.java -> AudioCallPermissionManager().requiredForegroundTypes
-                else -> setOf(-1)
+                else -> emptySet()
             }
         }
-        return setOf(-1)
+        return emptySet()
     }
 }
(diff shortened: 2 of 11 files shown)


stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -840,16 +857,50 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
...
+                        OwnCapability.SendVideo,
+                    )
+                    FOREGROUND_SERVICE_TYPE_MICROPHONE -> outgoingCallCapabilities.add(
+                        OwnCapability.SendAudio,
+                    )
+                }
+            }
         }
+        call.state.setOwnCapabilities(outgoingCallCapabilities)
     }
```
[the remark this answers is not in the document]
rahul-lohra: This is intentional as incoming-calls or active calls will already have filled `ownCapabilities`

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -840,16 +857,50 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
...
+                    // If other users will get push notification.
+                    ring = ring,
+                )
+                result.onOutcome(call, onSuccess, onError)
+            } else {
+                _renderPermissionUi.value = true
+                uiDelegate.setContent(activity = this@StreamCallActivity, call)
+                renderPermissionUi.first { !it }
+                create(call, ring, members, onSuccess, onError)
+            }
```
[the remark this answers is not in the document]
rahul-lohra: Correct

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ClientState.kt
```diff
@@ -184,9 +185,16 @@ class ClientState(private val client: StreamVideo) {
...
+                // when initiating outgoing-call
+                val callServiceConfig = callConfigRegistry.get(call.type)
+                val serviceClass = callServiceConfig.serviceClass
+                val isServiceRunning = ServiceIntentBuilder()
+                    .isServiceRunning(this.client.context, serviceClass::class.java)
+                if (callServiceConfig.runCallServiceInForeground) {
+                    if (!isServiceRunning) {
+                        logger.e { "Outgoing call service should already be running" }
+                    }
                 }
```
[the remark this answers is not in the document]
rahul-lohra: Correct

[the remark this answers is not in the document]
rahul-lohra: Its updated in new commit

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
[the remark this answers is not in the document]
rahul-lohra: Its updated in new commit. No need for infinite recursion guard at the moment, I will add ti later if needed

stream-video-android-ui-core/api/stream-video-android-ui-core.api
```diff
@@ -134,6 +135,8 @@ public abstract class io/getstream/video/android/ui/common/StreamCallActivity :
 	public fun reject (Lio/getstream/video/android/core/Call;Lio/getstream/video/android/core/model/RejectReason;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;)V
 	public final fun safeFinish ()V
 	protected fun setCallHandlerDelegate (Lio/getstream/video/android/ui/common/IncomingCallHandlerDelegate;)V
+	public final fun setRenderPermissionUi (Lkotlinx/coroutines/flow/StateFlow;)V
+	public final fun updateRenderPermissionUi (Z)V
```
rahul-lohra: This shouldn't be here. I already marked these variables with `@InternalStreamVideoApi`  @aleksandar-apostolov @gpunto

gpunto: If they are public, they are still part of the public api, the annotation just forces consumers to opt in

rahul-lohra: Yeah, I understand that if something is public it is technically part of the public API, and `@InternalStreamVideoApi` only forces consumers to opt-in.

What I’m observing is that the API validator behaves differently across modules.

For example, when a public API is annotated with @InternalStreamVideoApi:
	•	`In stream-video-android-core`, the API does not appear in the .api file, which is the behavior I expected.
	•	In `stream-video-android-ui-compose`, the API **does get added to the .api file`.

For reference, this public API added in the core module in the same PR did not update the .api file:
https://github.com/GetStream/stream-video-android/pull/1627/changes#diff-f229deea2aae6b806a751cc451836b9720165dd57e4de173cc40c2fd43169965R27-R29

My guess is that the apiValidation configuration is not applied consistently across modules.

It is currently configured in stream-video-android-core:
[stream-video-android-core/build.gradle](https://github.com/GetStream/stream-video-android/blob/94feaeaf4ae3676663bd2a59564e5cd060c4f54f/stream-video-android-core/build.gradle.kts#L52)

But it is missing in stream-video-android-ui-compose:
[stream-video-android-ui-compose/build.gradle](https://github.com/GetStream/stream-video-android/blob/develop/stream-video-android-ui-compose/build.gradle.kts)

To make the behavior consistent, we could define the marker once in a shared configuration so all stream-* modules inherit the same rule:

```kotlin
apiValidation {
    nonPublicMarkers += listOf(
        "io.getstream.video.android.core.internal.InternalStreamVideoApi"
    )
}
```
This way any public API annotated with `@InternalStreamVideoApi` would be treated as non-public by the validator across all modules and we avoid these inconsistencies.

For this PR, I’ll add the apiValidation configuration to stream-video-android-ui-compose to make the behavior consistent.

gpunto: > For example, when a public API is annotated with @InternalStreamVideoApi:
• In stream-video-android-core, the API does not appear in the .api file, which is the behavior I expected.
• In stream-video-android-ui-compose, the API **does get added to the .api file`.

Good point, I didn't know that!

> My guess is that the apiValidation configuration is not applied consistently across modules.

That definitely seems to be the case so it makes sense to also add the same config to the module 👍

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -840,16 +857,52 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
...
+                    // If other users will get push notification.
+                    ring = ring,
+                )
+                result.onOutcome(call, onSuccess, onError)
+            } else {
+                _renderPermissionUi.value = true
+                withContext(Dispatchers.Main) {
+                    uiDelegate.setContent(activity = this@StreamCallActivity, call)
+                }
+                renderPermissionUi.first { !it }
```
aleksandar-apostolov, line 880: If the user denies permissions, `NoPermissionUi` calls `InternalPermissionContent` for `SomeGranted`/`NoneGranted` but never calls `onAllPermissionGranted()` — wouldn't `_renderPermissionUi` never flip to `false`, causing `.first { \!it }` to suspend forever? Dead coroutine on IO, no error callback, no way out for the user. Should there be a cancel/deny path that sets `_renderPermissionUi = false` and calls `onError`?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/StreamCallActivityComposeDelegate.kt
```diff
@@ -124,13 +124,42 @@ public open class StreamCallActivityComposeDelegate : StreamCallActivityComposeU
                             .systemBarsPadding(),
                     ) {
                         logger.d { "[setContent] with RootContent" }
-                        activity.RootContent(call = call)
+
+                        val renderPermissionUi by activity.renderPermissionUi.collectAsStateWithLifecycle()
+                        if (renderPermissionUi) {
```
aleksandar-apostolov, line 129: Related to the above — `NoPermissionUi` only provides `onAllPermissionGranted` as a dismiss path. How do `SomeGranted` and `NoneGranted` signal failure back to the `create()` coroutine?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/StreamForegroundPermissionUtil.kt
```diff
@@ -0,0 +1,41 @@
...
+    public fun getForegroundPermissionsForCallType(callType: String): Set<Int> {
+        val client = StreamVideo.instanceOrNull() as? StreamVideoClient
+        if (client != null) {
+            val config = client.callServiceConfigRegistry.get(callType)
+            return when (config.serviceClass) {
+                CallService::class.java -> ForegroundServicePermissionManager().requiredForegroundTypes
+                AudioCallService::class.java -> AudioCallPermissionManager().requiredForegroundTypes
+                else -> setOf(-1)
+            }
+        }
```
aleksandar-apostolov, line 38: `setOf(-1)` as a sentinel for "no permissions" is fragile — if anything downstream iterates expecting valid `ServiceInfo.FOREGROUND_SERVICE_TYPE_*` constants, `-1` could cause unexpected behavior. `emptySet()` would be clearer.

rahul-lohra: Right

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/StreamCallActivityComposeDelegate.kt
rahul-lohra: This logic is intentional.

I have reused the logic of `SomeGranted` and `NoneGranted` which is present is this
```kotlin
override fun StreamCallActivity.RootContent(call: Call) { 
    LaunchPermissionRequest(getRequiredPermissions(call)) {
                        AllPermissionsGranted {
                            RingingCallContent()
                        }

                        SomeGranted { granted, notGranted, showRationale ->
                            InternalPermissionContent(showRationale, call, granted, notGranted)
                        }

                        NoneGranted {
                            InternalPermissionContent(it, call, emptyList(), emptyList())
                        }
                    }
}
```

So the `SomeGranted` and `NoneGranted` will invoke the same `StreamActivity.InternalPermissionContent()`

aleksandar-apostolov (approved):
All feedback addressed. LGTM.

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
rahul-lohra: Correct

PR #1631, opened 2026-03-16
SDK will now notify the coordinator which SFUs it already failed to join during a migration or reconnect

PratimMallick:
### Goal

When the SDK receives an SFU_FULL error (or any recoverable connection failure) from an SFU, it can enter a REJOIN loop — repeatedly requesting credentials from the coordinator and being assigned the same full SFU. This happens because the coordinator doesn't know which SFUs the client has already failed to connect to. Additionally, if an SFU becomes unreachable for any reason (network issues, timeouts), the client retries the same SFU indefinitely via the HealthMonitor without ever requesting a new one.
This change ensures the SDK:
Tracks SFUs it failed to connect to and tells the coordinator to exclude them when requesting new credentials.
Stops retrying an unreachable SFU after 2 attempts and triggers a migration to a new SFU.
Uses SFU IDs (edge_name) instead of URLs for migrating_from, migrating_from_list, and from_sfu_id in ReconnectDetails, aligning with the server-side expectations.

### Implementation

Call.kt:
1. Maintains a thread-safe failedSfuIds list that accumulates SFU edge names the client couldn't connect to.
addFailedSfuId() / getFailedSfuIdsSnapshot() / clearFailedSfuIds() manage this list.
2.migrate() adds the current SFU's edgeName to the failed list before requesting new credentials, and passes it as migratingFrom.
3.joinRequest() automatically populates migratingFromList from the failed SFU snapshot if no explicit list is provided.
onSfuConnectionEstablished() clears the failed list once a connection succeeds (called from RtcSession).
4.All references to SFU URLs in migratingFrom, migratingFromList, and ReconnectDetails.from_sfu_id are replaced with SFU edge names.

RtcSession.kt:
1.Added sfuName (edge name) constructor parameter for identifying the current SFU.
2.Added sfuConnectionRetryCount that increments on DisconnectedTemporarily and WebSocketEventLost states.
3.After MAX_SFU_CONNECTION_RETRIES (2) consecutive failures, disconnects the current socket and calls call.migrate() to request a new SFU from the coordinator.
4.Resets the counter on Connected state.

StreamVideoClient.kt 
1.  Extended joinCall() to accept and forward the migratingFromList parameter to the coordinator API.

### Testing

SFU_FULL scenario: 
Simulate or trigger an SFU_FULL error from the SFU. Verify the client adds the full SFU's edge name to the exclusion list, requests new credentials with migrating_from_list populated, and connects to a different SFU.

Unreachable SFU:
 Disconnect or block the SFU endpoint. Verify the client retries twice, then triggers a migration instead of retrying indefinitely.
 
Successful connection clears state: 
After migration to a new SFU succeeds, verify failedSfuIds is cleared so future requests are not unnecessarily constrained.

Normal rejoin unaffected: 
Verify that a simple network reconnection (rejoin) does not add the current SFU to the exclusion list, since the client may want to reconnect to the same SFU.

Unit tests can mock SfuSocketState transitions and verify that migrate() is called after 2 DisconnectedTemporarily events, and that clearFailedSfuIds() is called on Connected.

#### General

#### Code & documentation

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking and migrating between multiple SFU sources during call transitions.
  * Implemented automatic failover mechanism for SFU connections with built-in retry logic.

* **Improvements**
  * Enhanced overall call resilience with improved edge server failure detection and recovery mechanisms.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -637,8 +649,28 @@ public class RtcSession internal constructor(
...
+                        }
+                        if (sfuConnectionRetryCount > MAX_SFU_CONNECTION_RETRIES) {
+                            logger.w {
+                                "[stateJob] Max retries reached for $sfuName, requesting new SFU via migrate()"
+                            }
+                            sfuConnectionRetryCount = 0
+                            sfuConnectionModule.socketConnection.disconnect()
+                            call.migrate()
+                        }
+                    }
```
rahul-lohra: Can we add unit-test for this case?

PratimMallick: added

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -236,6 +236,12 @@ public class Call(
     var sessionId = UUID.randomUUID().toString()
     internal val unifiedSessionId = UUID.randomUUID().toString()
 
+    /**
+     * SFU IDs (edge names) we failed to connect to (e.g. SFU_FULL). Sent in migrating_from_list
+     * when requesting new credentials so the coordinator can exclude them.
+     */
+    private val failedSfuIds = Collections.synchronizedList(mutableListOf<String>())
```
rahul-lohra: Why don't we use a thread safe Set ? Like `ConcurrentHashMap.newKeySet<String>()`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1482,14 +1493,43 @@ public class Call(
         return clientImpl.muteUsers(type, id, request)
     }
 
+    /** Adds the given SFU ID (edge name) to the failed list (for migrating_from_list). */
+    private fun addFailedSfuId(sfuId: String) {
+        if (sfuId.isBlank()) return
+        synchronized(failedSfuIds) {
+            if (!failedSfuIds.contains(sfuId)) {
+                failedSfuIds.add(sfuId)
+            }
```
rahul-lohra: This double locking feels unnecessary. We should either switch to a thread-safe `Set` or remove the explicit `synchronized` block. What do you think?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1832,19 +1862,35 @@ public class RtcSession internal constructor(
...
+            sfuConnectionModule.socketConnection.simulateNetworkError(
+                error = io.getstream.result.Error.NetworkError(
+                    message = "Simulated SFU_FULL (code 700)",
+                    serverErrorCode = 700,
+                    statusCode = 700,
+                ),
+                reconnectStrategy = WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE,
+            )
+        }
+    }
```
rahul-lohra: Do you think methods that simulate errors should be part of the SDK’s default behavior?

They’re definitely useful for testing (e.g., in the demo app), but ideally we shouldn’t be shipping or executing this kind of logic in consumer apps unless explicitly needed.

One approach we could consider is gating these behind a compile-time flag so they’re only active in local/debug builds and effectively no-op (and stripped) in release.

```kotlin
buildTypes {
    debug {
        buildConfigField("boolean", "DEBUG_TOOLS_ENABLED", "true")
    }
    release {
        buildConfigField("boolean", "DEBUG_TOOLS_ENABLED", "false")
    }
}
```

```kotlin
// a helper method
internal inline fun debugOnly(block: () -> Unit) {
    if (BuildConfig.DEBUG_TOOLS_ENABLED) {
        block()
    }
}

// Your actual debeg method
 internal fun simulateSfuFull()  = debugOnly { 
// simulation logic
   } 
```

This way the demo app can still use these methods, while the production SDK stays clean.

Curious to hear your thoughts—does this direction make sense or do you see a better way to handle this?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -652,43 +715,10 @@ public class RtcSession internal constructor(
...
-                logger.e(
-                    it.streamError.extractCause()
-                        ?: IllegalStateException("Error emitted without a cause on SFU connection."),
-                ) { "permanent failure on socket connection" }
-            }
-        }
+        // NOTE: Reconnect strategies (MIGRATE, REJOIN, FAST, DISCONNECT) are handled
+        // in the stateJob above via DisconnectedTemporarily.reconnectStrategy. All SFU
+        // errors are Error.NetworkError which routes through onVideoNetworkError → state
+        // machine → DisconnectedTemporarily, never through the errors() flow.
```
rahul-lohra: Do you think this comment would be better as KDoc (/** */) instead of inline // comments?

Since it’s describing a broader flow (reconnect strategies, error routing, state transitions), using KDoc with references (e.g., [DisconnectedTemporarily], [onVideoNetworkError]) could make it easier to read and navigate—especially since those references become clickable in IDE.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Yeah makes sense to wrap these in debug flags . good idea

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: yep makes sense

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Yeah ConcurrentHashMap.newKeySet<String>() can be good choice here, since we want to avoid duplicates and it gives thread saftely while iterating as well

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: resolved with thread safe set

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -637,8 +667,62 @@ public class RtcSession internal constructor(
...
+                                logger.w {
+                                    "[stateJob] SFU sent DISCONNECT strategy for $sfuName"
+                                }
+                                sfuConnectionRetryCount = 0
+                                sfuConnectionModule.socketConnection.disconnect()
+                                call.state._connection.value = RealtimeConnection.Disconnected
+                            }
+
+                            else -> {
+                                // FAST or UNSPECIFIED: HealthMonitor handles the retry
```
aleksandar-apostolov: The old `errorJob` explicitly called `call.fastReconnect()` when the server sent `WEBSOCKET_RECONNECT_STRATEGY_FAST`. This new code puts FAST in the `else` branch alongside UNSPECIFIED — it increments the retry counter and relies on HealthMonitor. Does the HealthMonitor actually perform a fast reconnect here, or just a generic retry? If not, this would be a behavioral regression where the server says "fast reconnect" and the client no longer does one.

stream-video-android-core/src/main/kotlin/io/getstream/android/video/generated/models/JoinCallRequest.kt
```diff
@@ -51,6 +51,9 @@ data class JoinCallRequest (
     @Json(name = "migrating_from")
     val migratingFrom: kotlin.String? = null,
 
+    @Json(name = "migrating_from_list")
+    val migratingFromList: kotlin.collections.List<kotlin.String>? = null,
```
aleksandar-apostolov, line 55: This file is under `generated/models/` — would the next API codegen run overwrite this manual addition? Is the server-side spec already updated to include `migrating_from_list`, or is this patched ahead of the API?

PratimMallick: `migrating_from_list` is already in the server-side API spec. 
Added this manually because when we were doing the automated way to update the models, it turned out to be a breaking change as we have exposed a lot of models publically and an overall update would lead to breaking changes since model names have changed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: The HealthMonitor will only reconnect the raw WebSocket — it does not restart ICE or re-send a JoinRequest with ReconnectDetails (as done by the fastReconnect() method)

So this is a behavioral difference. Adding a case for RECONNECT_FAST

PratimMallick: Although develop branch had another issue where the errorJob never got trigerred atall for any Server side error messages. Hence it was never honoured and used to get the same HealthMonitor WebSocket reconnect treatment.

aleksandar-apostolov (approved):
FAST strategy fix looks good. LGTM.

PR #1633, opened 2026-03-19
Ensure smoother `Ringing` to `Active` transition

rahul-lohra:
### Goal

Ensure a responsive transition to `RingingState.Active`, where a user is not rendered to others unless their own local media is available.

### Previously

In ringing calls, we transitioned to `RingingState.Active` based on signaling events—either when the call was accepted by the other participant or when call.join() completed.

JoinAndRing | User | Logic to set RingingState.Active |
| --- | --- | --- | 
Yes | Caller | After `CallAcceptedEvent` |
Yes | Callee | After `join()` |
No | Caller | After `CallAcceptedEvent` |
No | Callee | After `join()` |

-------------------------------------------------

> 
> ### Current Approach
> 
> We previously considered transitioning to `RingingState.Active` only after both `Publisher` and `Subscriber` RTCPeerConnections are **connected to the SFU**.
> 
> This approach aimed to ensure media readiness before activating the UI.

### Problem

The Both Peers Connected strategy introduces one key issue:
	•	**Privacy Risk**: There can still be edge cases where a user is rendered remotely before they have a confirmed local preview, violating a strict privacy expectation.

Additionally, this approach does not align well with established real-world patterns.

Apps like WhatsApp and Telegram:
	•	Render the local camera preview immediately during the outgoing ringing state
	•	Show local preview on the callee side right after call acceptance
	•	Transition to remote media once available

This avoids the “active UI but no media” gap while maintaining privacy.

-------------------------------------------------

### Solution

Adopt the Publisher Connected strategy:
	•	Transition to RingingState.Active when the Publisher reaches PeerConnectionState.CONNECTED
	•	Ensure the user’s local media is available before exposing them to others
	•	Avoid waiting for Subscriber connection to reduce latency

This provides a better balance between:
	•	Privacy guarantees
	•	Faster perceived call setup
	•	Alignment with industry-standard UX patterns

| JoinAndRing | User | Logic to set RingingState.Active |
| --- | --- | --- | 
Yes | Caller | After `CallAcceptedEvent` & both one publisher & subscriber is connected with SFU |
Yes | Callee | After `join()` & publisher is connected with SFU  |
No | Caller | After `CallAcceptedEvent` & when publisher is connected with SFU |
No | Callee | After `join()` & both  publisher and subscriber is connected with SFU |

```kotlin
internal enum class TransitionToRingingStateStrategy {
    LEGACY_BEHAVIOUR,
    PUBLISHER_CONNECTED    
}
```
We can use any strategy to experiment, by default we will select `PUBLISHER_CONNECTED`
⸻

### Implementation

Core Changes | Reason |
| --- | --- | 
| Make `Call.session` observable | Need to observe publisher's and subscriber's connection state |
| Make `Publisher` & `Subscriber` both observable | Need to observe their connection state |
| Add new logic `transitionToActiveState` logic | Need to observe `PeerConnection` state with `timeout` before transition to RingingState.Active | 

```kotlin

//We will remove all the unused strategy once validation is done and will keep the one which is getting used 

internal enum class TransitionToRingingStateStrategy {
    LEGACY_BEHAVIOUR,
    PUBLISHER_CONNECTED    
}

private const val PEER_CONNECTION_OBSERVER_TIMEOUT = 5_000L

internal class ActiveStateGate(
    private val coroutineScope: CoroutineScope,
    private val previousRingingStates: Set<RingingState>,
    private val strategy: TransitionToRingingStateStrategy = TransitionToRingingStateStrategy.PUBLISHER_CONNECTED,
    private val timeoutMs: Long = PEER_CONNECTION_OBSERVER_TIMEOUT,
) {
    private val logger by taggedLogger("ActiveStateGate")
    private var peerConnectionObserverJob: Job? = null

    internal fun awaitAndTransition(currentRingingState: RingingState, call: Call, onReady: () -> Unit) {
        // check file changes
    }

    private fun observePeerConnection(call: Call, onReady: () -> Unit, strategy: TransitionToRingingStateStrategy) {
        if (peerConnectionObserverJob?.isActive == true) return

        peerConnectionObserverJob = coroutineScope.launch {
            val start = System.currentTimeMillis()

            val result = withTimeoutOrNull(timeoutMs) {
                call.session
                    .filterNotNull()
                    .flatMapLatest { session ->
                        when (strategy) {
                         // check file changes
                         }
                    .first()
            }
            onReady()
        }
    }

    fun cleanup() {...}
```

### 🎨 UI Changes

None, just the time to render `RingState.Active` has increased

### Testing
1. Perform audio-call (normal and joinAndRing) and observe that there is almost zero waiting time in audio transfer

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added peer connection state monitoring during call state transitions with configurable connection strategies (LEGACY_BEHAVIOUR or PUBLISHER_CONNECTED) and automatic 5-second timeout handling.

* **Improvements**
  * Enhanced call session state management for improved reactivity and reliability during ringing-to-active transitions.
  * Improved session lifecycle handling and state tracking across call operations.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
@@ -91,7 +91,10 @@ internal class ActiveStateGate(
                             }
 
                             TransitionToRingingStateStrategy.FIST_PACKET_RECEIVED ->
-                                subscriberFlow.filter { it == PeerConnection.PeerConnectionState.CONNECTED }
+                                session.subscriber
+                                    .filterNotNull()
+                                    .flatMapLatest { it.firstRtpPacketArrivedWithinTimeout }
+                                    .filter { it }
                                     .map { "first_packet" to it }
 
                             TransitionToRingingStateStrategy.ANY_PEER_CONNECTED -> {
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
@@ -27,6 +27,7 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.merge
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withTimeoutOrNull
 import org.webrtc.PeerConnection
@@ -135,7 +136,9 @@ internal class ActiveStateGate(
                     "[observeConnection-$strategy] Timeout after ${duration}ms"
                 }
             }
-            onReady()
+            if (isActive) {
+                onReady()
+            }
         }
     }
 
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
@@ -112,6 +112,7 @@ internal class ActiveStateGate(
             }
             if (isActive) {
                 onReady()
+                cleanup()
             }
         }
     }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
@@ -19,14 +19,12 @@ package io.getstream.video.android.core
 import io.getstream.log.taggedLogger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withTimeoutOrNull
@@ -37,7 +35,7 @@ private const val PEER_CONNECTION_OBSERVER_TIMEOUT = 5_000L
 internal class ActiveStateGate(
     private val coroutineScope: CoroutineScope,
     private val previousRingingStates: Set<RingingState>,
-    private val strategy: TransitionToRingingStateStrategy = TransitionToRingingStateStrategy.BOTH_PEER_CONNECTED,
+    private val strategy: TransitionToRingingStateStrategy = TransitionToRingingStateStrategy.PUBLISHER_CONNECTED,
     private val timeoutMs: Long = PEER_CONNECTION_OBSERVER_TIMEOUT,
 ) {
     private val logger by taggedLogger("ActiveStateGate")
@@ -73,51 +71,14 @@ internal class ActiveStateGate(
                             .filterNotNull()
                             .flatMapLatest { it.state }
 
-                        val subscriberFlow = session.subscriber
-                            .filterNotNull()
-                            .flatMapLatest { it.state }
-
                         when (strategy) {
-                            TransitionToRingingStateStrategy.NONE -> {
+                            TransitionToRingingStateStrategy.LEGACY_BEHAVIOUR -> {
                                 emptyFlow<Int>()
                                     .map { "none" to it }
                             }
-                            TransitionToRingingStateStrategy.DEBUG_PUBLISHER_CONNECTED -> {
+                            TransitionToRingingStateStrategy.PUBLISHER_CONNECTED -> {
                                 publisherFlow.filter { it == PeerConnection.PeerConnectionState.CONNECTED }
                                     .map { "publisher" to it }
-                            }
-                            TransitionToRingingStateStrategy.DEBUG_SUBSCRIBER_CONNECTED -> {
-                                subscriberFlow.filter { it == PeerConnection.PeerConnectionState.CONNECTED }
-                                    .map { "subscriber" to it }
-                            }
-
-                            TransitionToRingingStateStrategy.DEBUG_FIST_PACKET_RECEIVED ->
-                                session.subscriber
-                                    .filterNotNull()
-                                    .flatMapLatest { it.debugFirstRtpPacketArrivedWithinTimeout }
-                                    .filter { it }
-                                    .map { "first_packet" to it }
-
-                            TransitionToRingingStateStrategy.DEBUG_ANY_PEER_CONNECTED -> {
-                                merge(
-                                    publisherFlow.map { "publisher" to it },
-                                    subscriberFlow.map { "subscriber" to it },
-                                ).filter { (_, state) ->
-                                    state == PeerConnection.PeerConnectionState.CONNECTED
-                                }
-                            }
-
-                            TransitionToRingingStateStrategy.BOTH_PEER_CONNECTED -> {
-                                combine(publisherFlow, subscriberFlow) { pub, sub ->
-                                    if (
-                                        pub == PeerConnection.PeerConnectionState.CONNECTED &&
-                                        sub == PeerConnection.PeerConnectionState.CONNECTED
-                                    ) {
-                                        "both" to PeerConnection.PeerConnectionState.CONNECTED
-                                    } else {
-                                        null
-                                    }
-                                }.filterNotNull()
                             }
                         }
                     }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
@@ -20,6 +20,7 @@ import io.getstream.log.taggedLogger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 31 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -232,7 +232,11 @@ public class Call(
     private var isDestroyed = false
 
     /** Session handles all real time communication for video and audio */
-    internal var session: RtcSession? = null
+    private val _session: MutableStateFlow<RtcSession?> = MutableStateFlow(null)
+
+    @InternalStreamVideoApi
+    public val session: StateFlow<RtcSession?> = _session
```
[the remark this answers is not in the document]
rahul-lohra: Fixed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -676,7 +679,7 @@ public class CallState(
     private var speakingWhileMutedResetJob: Job? = null
     private var autoJoiningCall: Job? = null
     private var ringingTimerJob: Job? = null
-
+    private var peerConnectionObserverJob: Job? = null
```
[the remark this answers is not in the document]
rahul-lohra: Fixed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
```diff
@@ -0,0 +1,131 @@
...
+                logger.d {
+                    "[observeConnection-$strategy] $source reached $state in ${duration}ms"
+                }
+            } else {
+                logger.w {
+                    "[observeConnection-$strategy] Timeout after ${duration}ms"
+                }
+            }
+            transitionToActiveState()
+        }
```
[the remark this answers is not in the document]
rahul-lohra: As per your suggestion do you think checking `CoroutineScope.isActive` inside the coroutine make any sense?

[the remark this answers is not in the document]
rahul-lohra: Thank you for clarifying my doubt, I have added `isActive` check in a new commit

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -486,6 +486,8 @@ public class CallState(
     private val _ringingState: MutableStateFlow<RingingState> = MutableStateFlow(RingingState.Idle)
     public val ringingState: StateFlow<RingingState> = _ringingState
 
+    private val previousRingingStates = ConcurrentHashMap.newKeySet<RingingState>()
```
aleksandar-apostolov, line 489: `previousRingingStates` accumulates across the lifetime of the `CallState` instance. If a `Call` object is reused (ring → active → idle → ring again), wouldn't the stale states from the first call affect the gate logic on the second ring? Should this be cleared in `cleanup()` or when transitioning back to `Idle`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
```diff
@@ -0,0 +1,110 @@
...
+                    .filterNotNull()
+                    .flatMapLatest { session ->
+
+                        val publisherFlow = session.publisher
+                            .filterNotNull()
+                            .flatMapLatest { it.state }
+
+                        when (strategy) {
+                            TransitionToRingingStateStrategy.LEGACY_BEHAVIOUR -> {
+                                emptyFlow<Int>()
```
aleksandar-apostolov, line 89: The `LEGACY_BEHAVIOUR` strategy emits from `emptyFlow()`, which means it always falls through to the 5-second timeout. The actual legacy behavior was an immediate transition. Is this intentional — effectively a 5-second delay for "legacy" — or should it call `onReady()` immediately to match what the old code did?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
```diff
@@ -673,4 +683,42 @@ internal class Subscriber(
     fun trackIdToParticipant(): Map<String, String> = trackIdToParticipant.toMap()
 
     fun isEnabled() = enabled
+
+    /**
+     * Will be removed after testing is done
+     */
+    internal fun debugPollFirstPacket() {
```
aleksandar-apostolov: `debugPollFirstPacket()` runs `getStats()` every 700ms for up to 60 seconds on every subscriber creation. The comments say "will be removed after testing is done." Is the plan to remove this before merge, or should it be gated behind a debug/test flag?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -687,10 +689,10 @@ public class Call(
         }
         monitorPublisherPCStateJob?.cancel()
         monitorPublisherPCStateJob = scope.launch {
-            session?.publisher?.iceState?.collect {
+            session.value?.publisher?.value?.iceState?.collect {
```
aleksandar-apostolov: `session.value?.publisher?.value?.iceState?.collect` takes a one-time snapshot. If `publisher` is still `null` when `monitorSession` runs (publisher is created asynchronously after SFU connects), the ICE restart monitoring would be silently skipped. Now that `publisher` is a `StateFlow`, would it make sense to `flatMapLatest` on it here so monitoring starts when publisher becomes available?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1511,11 +1516,11 @@ public class Call(
 
     fun cleanup() {
         // monitor.stop()
-        session?.cleanup()
+        session.value?.cleanup()
         shutDownJobsGracefully()
```
aleksandar-apostolov, line 1531: `state.cleanup()` was added at the end of `leave()`, which cleans up the `ActiveStateGate`. But `Call.cleanup()` doesn't call `state.cleanup()`. Could there be a path where `cleanup()` is called without `leave()`, leaving the gate's coroutine job alive?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
```diff
@@ -0,0 +1,110 @@
...
+                                    .map { "none" to it }
+                            }
+                            TransitionToRingingStateStrategy.PUBLISHER_CONNECTED -> {
```
aleksandar-apostolov, line 92: The description table says "both publisher & subscriber is connected with SFU" for every scenario, but the `PUBLISHER_CONNECTED` strategy only observes publisher state — subscriber isn't checked. Which one is correct — the table or the code? I guess code.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
rahul-lohra: Yes it should get cleaned up

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
rahul-lohra: Will be removed. No need to include that logic on this PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
```diff
@@ -673,4 +683,42 @@ internal class Subscriber(
...
+                while (!debugFirstRtpPacketArrivedWithinTimeout.value) {
+                    debugCheckStats()
+                    delay(700)
+                }
+            }
+            debugFirstRtpPacketPollingJob?.cancel()
+        }
+    }
+
+    private fun debugCheckStats() {
```
PratimMallick: Do we need this function anymore?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
```diff
@@ -0,0 +1,110 @@
...
+    private val timeoutMs: Long = PEER_CONNECTION_OBSERVER_TIMEOUT,
+) {
+    private val logger by taggedLogger("ActiveStateGate")
+    private var peerConnectionObserverJob: Job? = null
+
+    internal fun awaitAndTransition(currentRingingState: RingingState, call: Call, onReady: () -> Unit) {
+        logger.d { "[awaitAndTransition], ringingState: $currentRingingState" }
+        val isIncomingOrOutgoing =
+            previousRingingStates.any { it is RingingState.Incoming || it is RingingState.Outgoing }
+        if (isIncomingOrOutgoing && currentRingingState !is RingingState.Active) {
```
PratimMallick: Maybe one scenario is missing in this if else condition 

when `isIncomingOrOutgoing` = `true` and `currentRingingState` is `RingingState.Active`  --> It does not enter if or else ..  should we should cover this scenario?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -761,12 +763,12 @@ public class Call(
...
         logger.d { "[fastReconnect] Reconnecting, reconnectAttepmts:$reconnectAttepmts" }
-        session?.prepareReconnect()
+        session.value?.prepareReconnect()
         this@Call.state._connection.value = RealtimeConnection.Reconnecting
-        if (session != null) {
+        if (session.value != null) {
             reconnectStartTime = System.currentTimeMillis()
 
-            val session = session!!
+            val session = session.value!!
```
PratimMallick: I would rename this variable to `currentSession` because this is just the snapshot and not the actual session flow

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
rahul-lohra: Yes on `LEGACY_BEHAVIOUR` we should call old code right away

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
rahul-lohra: Will be removed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
rahul-lohra: The code is correct. I cannot find the table that you have mentioned

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: Sure

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
aleksandar-apostolov: Screenshot 2026-03-31 at 15 42 52

aleksandar-apostolov: But since the code is correct I'll just update the table in the PR description.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: Fair, we can invoke `activeStateGate.cleanUp()` right after we have transitioned to `RingingState.Active`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
rahul-lohra: Thank you

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: Correct. I will update it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
rahul-lohra: If the `currentRingingState` is `RingingState.Active` then we don't need to do anything.

The purpose of this class is invoke callback `onReady()` when we want transition into `RingingState.Active`

aleksandar-apostolov (approved):
All feedback addressed. LGTM.

PR #1637, opened 2026-03-31
Fix a crash when initialising `StreamVideoClient` with a `UserType.Guest`

rahul-lohra:
### Goal

Fix a crash in the Android SDK when initialising `StreamVideoClient` with a `UserType.Guest` user and no token. On iOS and React, the SDK silently calls /video/guest to obtain a JWT automatically — Android should behave the same way. 

Additionally, the demo app incorrectly treated Guest and Anonymous as the same login path. This PR separates the two user types with correct labels and distinct entry points in the login UI.

### Implementation

SDK — two bug fixes:

1. StreamVideoBuilder — token blank validation was applied unconditionally to all user types. Scoped the check to UserType.Authenticated only. Guest users pass no token; the SDK internally calls POST /video/guest to fetch a JWT via guestUserJob.

2. CoordinatorAuthInterceptor — when token was blank (guest/anonymous), the interceptor still appended an Authorization: header with an empty string, causing the server to return HTTP 403. Added a guard so the Authorization header is only added when the token is non-blank.

### 🎨 UI Changes
 
image

image

### Testing

1. Sign out
2. Choose Sign in as guest
3. The guest user do not have the role of creating a call or joining an ended call.
4. Use another authenticated user to join a video with new ID
5. Join that room with the guest user
6. The guest user should be able to join the video room

## Summary by CodeRabbit

* **New Features**
  * Added guest user sign-in option to the login interface, providing an alternative way to access video calls
  * Guest users can now connect and participate in calls


demo-app/src/main/kotlin/io/getstream/video/android/util/StreamVideoInitHelper.kt
```diff
@@ -263,7 +357,7 @@ object StreamVideoInitHelper {
             user = user,
             token = token,
             connectionTimeoutInMs = 12_000L,
-            loggingLevel = loggingLevel,
+            loggingLevel = loggingLevel.copy(httpLoggingLevel = HttpLoggingLevel.BODY),
```
aleksandar-apostolov: Was the `httpLoggingLevel = HttpLoggingLevel.BODY` change intentional? This logs full HTTP request/response bodies for every API call across all user types — not just guest. If it was added for debugging the guest auth flow, should it be reverted before merge?

demo-app/src/main/kotlin/io/getstream/video/android/util/StreamVideoInitHelper.kt
```diff
@@ -180,6 +182,98 @@ object StreamVideoInitHelper {
...
+
+        isInitialising = false
+    }
+
+    /**
+     * Initialises the [StreamVideo] SDK for a [UserType.Anonymous] user.
+     * Anonymous users do not establish a WebSocket connection and receive no real-time events.
+     * They are suitable for passive viewing of livestreams.
+     */
+    suspend fun loadSdkForAnonymous(
```
aleksandar-apostolov: `loadSdkForAnonymous()` is defined here but never called — the "anonymous" login button in `LoginScreen.kt` still uses the old `login(true, null)` path. Is this meant to be wired up, or should it be removed?

demo-app/src/main/kotlin/io/getstream/video/android/util/StreamVideoInitHelper.kt
rahul-lohra: It will be removed

demo-app/src/main/kotlin/io/getstream/video/android/util/StreamVideoInitHelper.kt
rahul-lohra: It will be removed

aleksandar-apostolov (approved):
SDK fixes look good. LGTM.

PR #1644, opened 2026-04-14
Improve pinned participant state on rejoin with new session

rahul-lohra:
### Goal

Solves the gap where:

pinned participant leaves
joins again with another session
existing participants have obsolete pins information, and render the user unpinned

`ParticipantJoined` SFU event now will carry `is_pinned` property so you can use it for updating the client-side state correctly

### Implementation

React [PR](https://github.com/GetStream/stream-video-js/pull/2190)

- Add a property `is_pinned` in ParticipantJoinedEvent

**Core logic** 

```kotlin

fun handleEvent(event: VideoEvent) {
   is ParticipantJoinedEvent ->  updateServerSidePins(internalParticipants, event)
}

internal fun updateServerSidePins(internalParticipants: Map<SessionId, ParticipantState>, event: ParticipantJoinedEvent) {
    val participantSessionId = event.participant.session_id
    if (internalParticipants.containsKey(participantSessionId)) {
        if (event.isPinned) {
                updateServerSidePins(updatedList)
            }
        } else {
            _serverPins.value = _serverPins.value.filter { it.key != participantSessionId }
        }
    }
}

```

There are auto generated files, which are not the core part of this PR
1. SignalLostSignalingServiceDecorator.kt
2. SignalingServiceTracerDecorator.kt
3. go files

### 🎨 UI Changes

None

### Scenario A: Pin propagation on join
- [ ] Start a new call from Web with User A
- [ ] User B joins the call
- [ ] On User A, perform "Pin Everyone" on User B
- [ ] User C joins the call
- [ ] Verify User C sees User B pinned
Status: Passed ✅
---
### Scenario A – Rejoin Flow (User C)
- [ ] User C leaves the call
- [ ] User C rejoins
- [ ] Verify `On User C` still sees User B pinned
Status: Passed ✅
---
### Scenario A – Rejoin Flow (User B)
Step 1: User B leaves
- [ ] User B leaves the call
- [ ] Verify `on User C` that no participants are pinned
Status: Passed ✅
---
Step 2: User B rejoins flow
- [ ] User B rejoins the call
- [ ] Verify `on User B` that User B appears pinned
Observed:
- User B is not pinned
- Spotlight UI is not rendered
Root Cause:
- Pins from `JoinCallResponseEvent` contains a different sessionId
- Leads to mismatch while restoring pinned state
- Server-side issue (known) 
Status: Failed (expected due to known issue) ‼️
---
Step 3: User B rejoins flow
State on User C
- [ ] Verify `on User C` still sees User B pinned
Status: Passed ✅
---
## Notes
- Inconsistency observed:
  - User B (self view) → not pinned
  - User C (remote view) → sees User B pinned
  - Due to sessionId mismatch on rejoin (server-side)
---
## Summary
- Pin state works for new joins and non-pinned user rejoins
- Known issue: pinned user rejoin fails due to sessionId inconsistency


generate-sfu.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Step 3: Checkout to the correct branch, tag, or commit
+if [ "$REFERENCE_TYPE" == "branch" ]; then
+    git checkout "$REFERENCE_VALUE"
+elif [ "$REFERENCE_TYPE" == "tag" ]; then
+    git fetch --tags
+    git checkout "tags/$REFERENCE_VALUE"
+elif [ "$REFERENCE_TYPE" == "commit" ]; then
+    git fetch --depth=1 origin "$REFERENCE_VALUE"
+    git checkout "$REFERENCE_VALUE"
```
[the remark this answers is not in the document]
rahul-lohra: Ignore it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1096,8 +1104,14 @@ public class CallState(
                     // Remove any pins for the participant
                     unpin(sessionId)
                 }
-
+                Log.d(
+                    "Noob",
+                    "[ParticipantLeftEvent], _serverPins: ${_serverPins.value.map { it.key }.joinToString(
+                        ",",
+                    ) }",
+                )
```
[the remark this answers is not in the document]
rahul-lohra: Removed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1252,10 +1266,19 @@ public class CallState(
...
+            }
+
         // Update participants that are still in the call
         val pinnedInCall = pins.filter {
             internalParticipants.containsKey(it.sessionId)
         }
+        Log.d(
+            "Noob",
+            "[updateServerSidePins] 2, pins: $pins, internalParticipantsText: $internalParticipantsText, pinnedInCall: $pinnedInCall",
+        )
```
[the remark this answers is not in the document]
rahul-lohra: Removed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1458,6 +1502,9 @@ public class CallState(
     }
 
     public fun upsertParticipants(participants: List<ParticipantState>) {
+        val participantNames = participants.joinToString(",") {
+            "[name:${it.name.value}, sessionId:${it.sessionId}]"
+        }
```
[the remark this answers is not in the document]
rahul-lohra: removed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1265,6 +1272,27 @@ public class CallState(
...
+                val pinUpdate =
+                    PinUpdate(event.participant.user_id, participantSessionId)
+                val tempPinUpdateList = _serverPins.value.map { it.value.it }
+                val participantIsNotPresent =
+                    tempPinUpdateList.none { it.sessionId == participantSessionId }
+                if (participantIsNotPresent) {
+                    val updatedList = tempPinUpdateList.toMutableList().apply {
+                        add(pinUpdate)
+                    }
+                    updateServerSidePins(updatedList)
```
aleksandar-apostolov, line 1288: `updateServerSidePins(updatedList)` rebuilds the entire `_serverPins` map with `OffsetDateTime.now()` for every entry — so when one new pinned participant joins, every other pinned user's timestamp also gets refreshed. Is that intentional?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1265,6 +1272,27 @@ public class CallState(
...
+                }
+            } else {
+                _serverPins.value = _serverPins.value.filter { it.key != participantSessionId }
```
aleksandar-apostolov, line 1291: The new `participantSessionId` can't already be in `_serverPins` (sessionIds are unique per session) — so this filter looks like a no-op for the `isPinned == false` branch. Defensive, or am I missing a case where the sessionId could repeat?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1099,6 +1105,7 @@ public class CallState(
                 }
 
                 if (_serverPins.value.containsKey(sessionId)) {
+                    _serverPins.value = _serverPins.value.filter { it.key != event.participant.session_id }
```
aleksandar-apostolov, line 1108: Could this be the actual root cause of the "Failed: User B rejoin" scenario in the PR description, rather than the sessionId mismatch? JS doesn't do this — its leave handler just removes the participant locally, leaving the server-side pin alone, which is why JS rejoin works. Android calls `unpinForEveryone` here, so when User B leaves the server is told to unpin them; by the time they rejoin, `is_pinned` on the server is `false` and the new event handler has nothing to act on.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1265,6 +1272,27 @@ public class CallState(
         }
     }
 
+    internal fun updateServerSidePins(internalParticipants: Map<SessionId, ParticipantState>, event: ParticipantJoinedEvent) {
```
aleksandar-apostolov, line 1275: The new method is internal so directly testable. Worth adding a few cases to test the new behaviour?

rahul-lohra: Yes

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
rahul-lohra: Will remove this code

rahul-lohra:
Going with better implementation: https://github.com/GetStream/stream-video-android/pull/1683

PR #1645, opened 2026-04-15
Unify reconnect pipeline and harden network transition handling

PratimMallick:
### Goal

To Unify the reconnect pipeline and harden network transition handling

### Summary
- Replaces the scattered `fastReconnect()` / `rejoin()` / `migrate()` call sites with a single `Call.reconnect(strategy, reason)` entry point that owns all retry logic and strategy escalation (FAST → REJOIN, MIGRATE → REJOIN)
- Fixes call drops during network transitions (cellular ↔ WiFi) caused by false disconnect signals and stale leave timers
- Centralizes RtcSession teardown via `cancelActiveWork()` to prevent stale background jobs from triggering unwanted reconnects on old sessions
- Adds `RetryableSignalingServiceDecorator` with `StreamRetryPolicy` / `StreamRetryProcessor` for robust SFU API call retries with proper error propagation

#### Unified reconnect loop (`Call.kt`)
Previously, `fastReconnect()`, `rejoin()`, and `migrate()` were independent methods scheduled through `StreamSingleFlightProcessorImpl`. Multiple concurrent triggers (stateJob, NetworkStateListener, SfuSocket errors, HealthMonitor) could race and queue conflicting reconnect paths.
Now all callers funnel through `reconnect(strategy, reason)` which:
- Acquires a `Mutex` via `tryLock()` — concurrent triggers are dropped, not queued
- Runs a retry loop with automatic escalation: FAST → REJOIN after 3 attempts or deadline, MIGRATE → REJOIN on failure
- Handles `PeerConnectionNotUsableException` (stale WebRTC peer connections) as an escalation trigger
- Exits cleanly on `Connected`, `ReconnectingFailed`, `Disconnected`, or max attempts
#### Network transition reliability (`NetworkStateProvider`, `Call.kt`)
- `onLost` now checks actual connectivity before signaling disconnect — prevents false events during transport handoffs (OS releases cellular *after* WiFi is already active)
- Leave timer in `onDisconnected` guards against leaving when the connection has already recovered
#### RtcSession cleanup (`RtcSession.kt`)
- `cancelActiveWork()` centralizes cancellation of stateJob, eventJob, muteStateSyncJob, participantsMonitoringJob, and serialProcessor — every teardown path (migration, reconnect, rejoin, leave) goes through it
- stateJob is now a thin forwarding layer: `DisconnectedTemporarily` → `call.reconnect(strategy)`, `DisconnectedPermanently` → `call.reconnect(REJOIN)` — no more inline escalation logic or retry counters
- Network-aware guards skip reconnect triggers when the network is down or the call is already disconnected
- `fastReconnect` uses `withTimeout` + `state().first { Connected }` instead of callback-based `whenConnected`; throws `PeerConnectionNotUsableException` on stale peer connections instead of calling rejoin directly
- Migration follows a phased approach: `enterMigration()` → new session connect → `finalizeMigration()` — old peer connections stay alive during transition
#### SFU API retry layer (`RetryableSignalingServiceDecorator`)
- Replaces `SignalLostSignalingServiceDecorator` with a decorator that retries transient SFU API failures via `StreamRetryProcessor`
- Separates error propagation: SFU response errors → `onSfuApiError` (maps to reconnect strategy), network failures → `onSfuNetworkFailure` (triggers FAST reconnect)
- New `StreamRetryPolicy` and `StreamRetryProcessor` provide configurable linear/exponential backoff
#### Socket layer hardening (`HealthMonitor`, `SfuSocket`, `SfuSocketStateService`)
- `HealthMonitor` accepts `isNetworkAvailable` lambda — skips reconnect when network is down
- `SfuSocketStateService`: `NetworkDisconnected` stays parked on socket errors (no futile retry loops); handles `NetworkAvailable` event for recovery
- `SfuSocket`: cancels stale `socketListenerJob` on reconnect; cleans up old WebSocket before creating new one

### Testing

- [x] `ReconnectEscalationTest` — verifies FAST → REJOIN escalation, `PeerConnectionNotUsableException` handling, and mutex non-deadlock
- [x] `SfuConnectionRetryTest` — stateJob forwarding, network-aware guards, `DisconnectedPermanently` escalation
- [x] `ReconnectAttemptsCountTest` — reconnect attempt counter increments correctly through unified loop
- [x] `ReconnectDetailsTest` — reconnect details populated correctly for FAST/REJOIN/MIGRATE
- [x] `RetryableSignalingServiceDecoratorTest` — retry behavior, terminal error propagation, network failure callbacks
- [x] `StreamRetryPolicyTest`, `StreamRetryProcessorTest` — policy configuration and processor retry semantics
- [x] `SignalingServiceTracerDecoratorKtTest` — tracing decorator integration
- [x] Manual: cellular → WiFi and WiFi → cellular transitions verified stable with no call drops

#### General

#### Code & documentation

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

## Summary by CodeRabbit

* **New Features**
  * Added `ReconnectingFailed` connection state to indicate when reconnection attempts are exhausted.

* **Bug Fixes**
  * Improved reconnection robustness with automatic retry logic for transient failures.
  * Enhanced network state handling to conditionally manage reconnection attempts based on connectivity.
  * Better error recovery for signaling and SFU communication failures.

* **Tests**
  * Added comprehensive test coverage for reconnection escalation, retry behavior, and network state scenarios.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +775,321 @@ public class Call(
...
-                val (prevSessionId, subscriptionsInfo, publishingInfo) = session.currentSfuInfo()
-                val reconnectDetails = ReconnectDetails(
-                    previous_session_id = prevSessionId,
-                    strategy = WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE,
-                    announced_tracks = publishingInfo,
-                    subscriptions = subscriptionsInfo,
-                    from_sfu_id = oldSfuName,
-                    reconnect_attempt = reconnectAttepmts,
+        val loc = location
+            ?: throw IllegalStateException("No location available for migrate")
```
rahul-lohra: Do you think this throw can cause the app to crash?
If yes, then we should look for some alternative

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/RetryableSignalingServiceDecorator.kt
```diff
@@ -0,0 +1,152 @@
...
+        crossinline block: suspend () -> T,
+    ): T {
+        var lastResponse: T? = null
+
+        val result = retryProcessor.retry(policy) {
+            val response = block()
+            lastResponse = response
+            val error = errorExtractor(response)
+            if (error != null && error.should_retry && !error.requiresReconnect()) {
+                throw SfuRetryableException(error)
```
rahul-lohra, line 162: Can we return with failure result? Its better than throwing exception as we need to remember to catch it

rahul-lohra:
@coderabbit review specifically for unhandled exceptions and coroutine misuse in this PR

rahul-lohra:
@coderabbitai review specifically for unhandled exceptions and coroutine misuse in this PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
```diff
@@ -0,0 +1,97 @@
...
+        throw IllegalStateException("Check your policy: $policy")
+    }
+}
+
+@Suppress("TooGenericExceptionCaught")
+private inline fun <T> runCatchingCancellable(block: () -> T): Result<T> {
+    return try {
+        Result.success(block())
+    } catch (ce: CancellationException) {
+        throw ce
```
rahul-lohra, line 93: I think its better to return result failure, we don't want to the app. Wdyt?

rahul-lohra:
@coderabbitaireview in this PR. Can you review for the following
1. The code should flow in uni-direction way
2. Enclosing class should need keep the reference of outer class. You can provide suggestion to provide callbacks or flow

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: No this is called from from try block and the exception gets caught in the catch block of the caller which then determines the next step of what to do . This doesnt reach the app

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
PratimMallick: If wrapped  in Result.failure(ce) instead, the function would return normally — the coroutine runtime would think the work completed successfully, and cancellation would silently stop propagating.

PratimMallick: When a user leaves a call, call.scope is cancelled. That should cancel any in-flight SFU signaling calls (setPublisher, sendAnswer, etc.) running through RetryableSignalingServiceDecorator. If StreamRetryProcessor swallowed the CancellationException into a Result.failure:

1. The retryCall in RetryableSignalingServiceDecorator would hit result.getOrElse 
2. The throwable would be a CancellationException, not a SfuRetryableException
3. It would call onNetworkFailure(throwable) — triggering a reconnect for a call the user intentionally left
4. Then throw throwable would re-throw it anyway, but the damage (spurious reconnect) is done

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +775,321 @@ public class Call(
...
+                if (clientImpl.leaveAfterDisconnectSeconds > 0 &&
+                    elapsedMs / 1000 > clientImpl.leaveAfterDisconnectSeconds
+                ) {
+                    logger.w { "[reconnect] Disconnection timeout reached — giving up" }
+                    state._connection.value = RealtimeConnection.ReconnectingFailed
+                    break
+                }
+
+                if (currentStrategy != WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_FAST) {
+                    reconnectAttepmts++
```
aleksandar-apostolov: Nit: typo — `reconnectAttepmts` should be `reconnectAttempts` ('m' and 'p' are swapped). Same pattern with `reconnectDeadlineMils` → `reconnectDeadlineMillis`. Both show up in log messages, so they'll propagate into production logs.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1853,48 +1828,75 @@ public class RtcSession internal constructor(
...
+    /**
+     * Cancels all active background jobs and stops the serial processor so that
+     * no stale SFU API calls can trigger unwanted reconnects. Every teardown
+     * path (migration, reconnect, rejoin, leave) goes through this first.
+     *
+     * @param cancelEventJob whether to also cancel [eventJob]. Migration defers
+     *   this to [finalizeMigration] because the event loop is still needed while
+     *   the new session is being established.
+     */
+    private fun cancelActiveWork(cancelEventJob: Boolean = true) {
```
aleksandar-apostolov, line 1942: Question: the old `prepareRejoin()` also called `sfuConnectionModule.socketConnection.disconnect()` and closed `publisher`/`subscriber` peer connections. Those calls aren't in `cancelActiveWork()` — are they handled by every caller separately, or were they intentionally dropped?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1853,48 +1828,75 @@ public class RtcSession internal constructor(
...
-                    }
-                }
+        logger.d { "[fastReconnect] Connecting RTC, $request" }
+        sfuConnectionModule.socketConnection.connect(request)
+        try {
+            withTimeout(SocketActions.DEFAULT_SOCKET_TIMEOUT) {
+                sfuConnectionModule.socketConnection.state().first { it is SfuSocketState.Connected }
             }
+        } catch (e: Exception) {
+            sfuTracer.trace("fast-reconnect-connection-failed", "${e.message}")
```
aleksandar-apostolov: Question: this catches all `Exception` including `CancellationException` — `sendCallStats()` would then run on a cancelled coroutine. Would it be safer to let `CancellationException` fly without side effects?

```suggestion
        } catch (e: CancellationException) {
            throw e
        } catch (e: Exception) {
            sfuTracer.trace("fast-reconnect-connection-failed", "${e.message}")
            sendCallStats()
            throw e
        }
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +775,321 @@ public class Call(
...
+        logger.d { "[reconnectRejoin] reconnectAttempts=$reconnectAttepmts" }
+        state._connection.value = RealtimeConnection.Reconnecting
+        val loc = location
+            ?: throw IllegalStateException("No location available for rejoin")
+        reconnectStartTime = System.currentTimeMillis()
+
+        val joinResponse = joinRequest(location = loc)
+        if (joinResponse is Success) {
+            val cred = joinResponse.value.credentials
+            val oldSession = this.session.value!!
```
aleksandar-apostolov: Question: `session.value` could theoretically be null under rapid reconnect. Safer to use an explicit check?

```suggestion
            val oldSession = this.session.value
                ?: throw IllegalStateException("No active session available for rejoin")
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -407,12 +404,12 @@ public class Call(
     private var leaveTimeoutAfterDisconnect: Job? = null
     private var lastDisconnect = 0L
     private var reconnectDeadlineMils: Int = 10_000
```
aleksandar-apostolov: Question: this default was 30s (`sfuReconnectTimeoutMillis`) before — now it's 10s. Line 693 shows it gets overwritten by the server's `fastReconnectDeadlineSeconds` after join, so the default only applies before the first join response. Is 10s enough for users on slow networks. Also wasn't the old constant public API?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/RetryableSignalingServiceDecorator.kt
PratimMallick: The throw at line 130 is intentional and can't be replaced with a failure result — it's the mechanism that drives the retry loop.

The processor retries by catching exceptions — that's the only signal it has to distinguish "this attempt failed, try again" from "this attempt succeeded, return the value." If the block returned a Result<T> instead of throwing, the processor would see it as a successful return of a Result.failure value and stop retrying.

This processor is taken from core, so not changing the contract of RetryProcessor

PratimMallick: The throw is fully contained — it never escapes retryCall():after retries exhaust, runCatchingCancellable wraps it in Result.failure, and getOrElse at line 135 recovers lastResponse so callers always get a normal SFU response object back. It's internal control flow, not an unhandled exception risk.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: fixed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: The socket disconnect and publisher/subscriber close were not dropped — they're in cleanup(), which is called by each path that needs full teardown:

-  **_Rejoin_**: `prepareRejoin()` → `cancelActiveWork() `→ `leaveWithReason()` → `cleanup() `**(full teardown)**

- **_Migrate_**: `enterMigration() → cancelActiveWork(cancelEventJob=false), then finalizeMigration() → cleanup()` (deferred teardown — keeps event loop alive until new session connects)

- **_Fast reconnect_**: `prepareReconnect() → cancelActiveWork() + socket disconnect only`. Peer connections are intentionally kept open since fast reconnect reuses them on the same SFU.

`cancelActiveWork() `is intentionally limited to cancelling jobs and the serial processor. It's the common subset shared by all three paths. The socket/peer-connection teardown varies per strategy, so it stays in the caller-specific methods.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Good catch — fixed. Added `catch (e: CancellationException) { throw e } `before the generic catch. This avoids running `sendCallStats()` and logging a misleading trace when the coroutine was cancelled . `TimeoutCancellationException` from `withTimeout` is also a `CancellationException` so it skips the side effects too, but that's acceptable — the reconnect loop in `Call.kt` already logs the failure and handles escalation.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: `sfuReconnectTimeoutMillis` was already dead code on `develop` — declared but never referenced. It was a leftover from a previous reconnect implementation. The `reconnectDeadlineMillis` (10s default, overwritten by server) is the value that actually drives FAST-vs-REJOIN decisions,

PratimMallick: Adding back the `sfuReconnectTimeoutMillis` so that there's no breaking change with a `@deprecated` annotation

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: OKay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
```diff
@@ -0,0 +1,97 @@
...
+        policy: StreamRetryPolicy,
+        block: suspend (attempt: StreamRetryAttemptInfo) -> T,
+    ): Result<T> = runCatchingCancellable {
+        var delayMs = policy.initialDelayMillis
+        var attempt = 1
+        var previousError: Throwable? = null
+        while (attempt <= policy.maxRetries) {
+            if (delayMs > 0) {
+                delay(delayMs)
+            }
```
rahul-lohra, line 55: Do you think this whole function logic should be unit testable? I am aware of private visibility. But unit-test can safe guard us against future misuses. What do you think?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
rahul-lohra: okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/RetryableSignalingServiceDecorator.kt
rahul-lohra: okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
```diff
@@ -0,0 +1,97 @@
...
+private class StreamRetryProcessorImpl(private val tag: String) : StreamRetryProcessor {
+
+    override suspend fun <T> retry(
+        policy: StreamRetryPolicy,
+        block: suspend (attempt: StreamRetryAttemptInfo) -> T,
+    ): Result<T> = runCatchingCancellable {
+        var delayMs = policy.initialDelayMillis
+        var attempt = 1
+        var previousError: Throwable? = null
+        while (attempt <= policy.maxRetries) {
```
rahul-lohra, line 52: If max retries is zero then the while logic is skipped

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +783,325 @@ public class Call(
...
+
+            val isMigrate = strategy ==
+                WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE
+            state._connection.value = if (isMigrate) {
+                RealtimeConnection.Migrating
+            } else {
+                RealtimeConnection.Reconnecting
+            }
+
+            val reconnectStartTime = System.currentTimeMillis()
```
rahul-lohra: Do you think we should change the variable name as it conflicts with class level variable reconnectStartTime at line 255?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +783,325 @@ public class Call(
...
+                if (clientImpl.leaveAfterDisconnectSeconds > 0 &&
+                    elapsedMs / 1000 > clientImpl.leaveAfterDisconnectSeconds
+                ) {
+                    logger.w { "[reconnect] Disconnection timeout reached — giving up" }
+                    state._connection.value = RealtimeConnection.ReconnectingFailed
+                    break
+                }
+
+                if (currentStrategy != WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_FAST) {
+                    reconnectAttempts++
```
rahul-lohra: Right now we have two variables with similar names
1. attempt (local variable) Line 850
2. reconnectAttempts (class level variable)

Do you think we should do something about it? Because a reader has to trace both to understand what's happening.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +783,325 @@ public class Call(
...
+                        WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE ->
+                            reconnectMigrate()
+
+                        WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_DISCONNECT -> {
+                            logger.w { "[reconnect] DISCONNECT requested — leaving call" }
+                            leave("SFU:DISCONNECT")
+                            break
+                        }
+                    }
+                    break
```
rahul-lohra: Do we need this `break`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +783,325 @@ public class Call(
...
+                        WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN ->
+                            reconnectRejoin(reason)
+
+                        WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE ->
+                            reconnectMigrate()
+
+                        WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_DISCONNECT -> {
+                            logger.w { "[reconnect] DISCONNECT requested — leaving call" }
+                            leave("SFU:DISCONNECT")
+                            break
```
rahul-lohra: Do we need this `break`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +783,325 @@ public class Call(
...
+     * Reuses the existing session ID — no previous_session_id needed since the
+     * SFU already knows this participant.
      */
-    suspend fun fastReconnect(reason: String = "unknown") = schedule("fast") {
-        logger.d { "[fastReconnect] Reconnecting, reconnectAttepmts:$reconnectAttepmts" }
-        session.value?.prepareReconnect()
-        this@Call.state._connection.value = RealtimeConnection.Reconnecting
-        if (session.value != null) {
-            reconnectStartTime = System.currentTimeMillis()
+    private suspend fun reconnectFast(reason: String) {
```
rahul-lohra: The functions which returns `Unit` and capable of throwing exceptions internally is not good in the following way. 

Whenever I have to invoke a function which returns `Unit` then I don't want to look inside the implementation unless there is reason for it. 

The function contract is simple - it should do what the name says, the return type should indicate how the caller should handle it

I will recommend add return type of `Result<Unit>` to these new functions
```kotlin
private suspend fun reconnectFast(reason: String): Result<Unit> 
private suspend fun reconnectRejoin(reason: String): Result<Unit>
private suspend fun reconnectMigrate(): Result<Unit>
```
Now returning the `Result` type will right away tell that functions needs additional handing for failure cases.

I am aware that there is try catch added to handle these exceptions on the caller side. But we have 2 problems
1. The invoked functions isn't telling you that it can fail through its contract. So you need to either remember it or always look inside
2. If we keep it this way then we will forget after 2-3 months which methods needs to be in try-catch unless we look inside

So wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryPolicy.kt
```diff
@@ -0,0 +1,128 @@
...
+    companion object {
+        /**
+         * Creates a policy with accelerating backoff.
+         *
+         * Note: the growth formula is `prev + retry * backoffStepMillis` which
+         * produces quadratic growth (e.g. 250 → 750 → 1500 → 2500), not true
+         * exponential doubling. This matches the stream-core-android retry
+         * behavior and is kept for compatibility.
+         */
+        fun exponential(
```
rahul-lohra, line 55: This function is not used at the moment. Do you think we can remove it? We can always add it later if needed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: FFixed this by the TOCTOU changes

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
PratimMallick: `StreamRetryProcessorTest.kt `has 7 tests that exercise all the key behaviors through the StreamRetryProcessor(tag) factory function: success, retry-and-recover, max retries exhaustion, minRetries gating, giveUp early exit, CancellationException propagation, and attempt info carrying previous errors. The private visibility of StreamRetryProcessorImpl isn't a problem since the factory function exposes it for testing. The runCatchingCancellable helper is implicitly tested via the CancellationException test.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
PratimMallick: The factory methods` (exponential, linear, fixed)` all call `validate()` which enforces` minRetries >= 1 and maxRetries >= minRetries`, so maxRetries is always >= 1 through the normal API. If someone constructs a `StreamRetryPolicy` directly with` maxRetries = 0`, the while loop is skipped and the fallthrough at line 84 throws` IllegalStateException("Check your policy: ...")`, which `runCatchingCancellable` wraps into `Result.failure` — so it doesn't crash, it returns a failure with a descriptive message. The guard is already there. 
This is taken as it is from android-core

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: yes good catch. 
Renamed the local to `loopStartTime` and updated both references

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: correct .. make the local variable inside the loop as `loopIteration` and also adding comments on both the variables so that its use is made clear

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +783,325 @@ public class Call(
...
+
+    /**
+     * Rejoin a call. Creates a new session ID and joins as a new participant.
+     * previous_session_id is set so the SFU can transfer state (tracks,
+     * subscriptions) from the old session to the new one.
+     */
+    private suspend fun reconnectRejoin(reason: String) {
+        logger.d { "[reconnectRejoin] reconnectAttempts=$reconnectAttempts" }
+        state._connection.value = RealtimeConnection.Reconnecting
+        val loc = location
```
rahul-lohra, line 1070: These IllegalStateException are caught under the umbrella of Exception. Do you all `IllegalStateException` must be `retried` ?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: This is the success path. If `reconnectFast/reconnectRejoin/reconnectMigrate` returns without throwing, the reconnect succeeded. The break exits the `while (true)` loop. Without it, the loop would go back to the top and try to reconnect again despite already being connected.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: redundant, but communicates intent ("we're done, leave immediately")

PratimMallick: It's a readability choice — makes it explicit that DISCONNECT doesn't fall through to the success-path break. I'd keep it for clarity since the when block is large and a reader might not immediately see the break at end of when.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryPolicy.kt
PratimMallick: keeping it as a building block, so we can make a change easily later without taking in the changes from android-core

PratimMallick: but adding comments to linear and fixed policies

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Good catch : 

All precondition guards in `reconnectFast`, `reconnectRejoin`, `reconnectMigrate` now throw `ReconnectPreconditionException` instead of `IllegalStateException`.

Catch block now has three layers in order:

1. `CancellationException` → re-throw (coroutine cancellation)
2. `ReconnectPreconditionException` → log + set ReconnectingFailed + break (terminal, no retry)
3. `Exception` → retry with escalation (transient failures)

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt
rahul-lohra: > This is taken as it is from android-core

This is a good thing but as a code reviewer we must flag issues in the PR, even if it is taken from core or outside. 

> The factory methods (exponential, linear, fixed) all call validate() which enforces minRetries >= 1 and maxRetries >= minRetries

Its good thing, but that is not the problem. 

The problem I am trying to highlight is that the method signature proudly claims that pass whatever `StreamRetryPolicy` so it is which is not restricted to just `(exponential, linear, fixed)`  and there is only 1 semi-validation


Now we have 2 policy validators
1. private fun `StreamRetryPolicy.validate(): StreamRetryPolicy {..}` ~ There is no contract to validate a new instace of `StreamRetryPolicy`. So it means I must remember it

2. Inside `fun retry` - `attempt <= policy.maxRetries`. This is getting used as 2 things 1. while loop and StreamRetryPolicy validation  but the validation does not cover all cases. 
We should have proper validation inside retry

Additionally, we have a free hand to create and pass any values of `StreamRetryPolicy`. We should have put the restrictions here like we can use `IntRange`


```kotlin
import androidx.annotation.IntRange

internal data class StreamRetryPolicy(
    val minRetries: @IntRange(from = 1) Int,
    val maxRetries: @IntRange(from = 1) Int,
    val minBackoffMills: @IntRange(from = 1) Long,
    val maxBackoffMills: @IntRange(from = 1) Long,
    val initialDelayMillis: @IntRange(from = 0) Long,
    val giveUpFunction: (retry: Int, cause: Throwable) -> Boolean,
    val nextBackOffDelayFunction: (retry: Int, previousDelay: Long) -> Long,
) 
 ```
Wdyt?

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt (hunk 13 above)
Comment (rahul-lohra) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt]: oh, actually this `break` was hard a bit hard to follow up.

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt (hunk 14 above)
Comment (rahul-lohra) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt]: okay

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt (hunk 9 above)
Comment (rahul-lohra) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt:55]: Since it is already tested then we are good

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt (hunk 10 above)
Comment (aleksandar-apostolov) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt:52]: This is valid in principle, but low risk in practice this PR. We will fix this in core and validate the retry policy on the consume border. For the purpose of video this is internal anyway.

Comment (rahul-lohra) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/retry/StreamRetryProcessor.kt:52]: Okay

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt [hunk 18]
```diff
@@ -778,184 +807,383 @@ public class Call(
...
-                    this.session.value?.connect(reconnectDetails, currentOptions)
+        val loc = location
+            ?: return ReconnectOutcome.PreconditionNotMet("No location available for migrate")
+        val oldSession = session.value
+            ?: return ReconnectOutcome.PreconditionNotMet("No active session for migrate")
+        reconnectStartTime = System.currentTimeMillis()
+        addFailedSfuId(oldSession.sfuName)
+
+        val joinResponse = try {
+            joinRequest(location = loc, migratingFrom = oldSession.sfuName)
```
rahul-lohra: `joinRequest` returns a `Result` Why is this in a `try-catch`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +807,383 @@ public class Call(
...
+            cred.server.url,
+            cred.server.wsEndpoint,
+            cred.token,
+            cred.server.edgeName,
+            cred.iceServers.map { ice -> ice.toIceServer() },
+        )
+        this.session.value = newSession
+
+        return try {
+            val result = newSession.internalConnect(reconnectDetails, currentOptions)
```
rahul-lohra: Looking at the function signature I don't get any sign that it carries a hidden logic of throwing exception. The function signature is dishonest. This does not seems scalable.

I will try to look for a way to resolve this.

A quick way to resolve this would be changing the method name to something like 

`newSession.internalConnectWithException` so the method caller will know he needs to look inside for exception

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +807,383 @@ public class Call(
...
-                }
-                state._connection.value = RealtimeConnection.Reconnecting
+        val loc = location
+            ?: return ReconnectOutcome.PreconditionNotMet("No location available for rejoin")
+        val oldSession = session.value
+            ?: return ReconnectOutcome.PreconditionNotMet("No active session for rejoin")
+        reconnectStartTime = System.currentTimeMillis()
+
+        val joinResponse = try {
+            joinRequest(location = loc)
```
rahul-lohra: `joinRequest` returns a Result. Could you please explain why is this in a try-catch?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -778,184 +807,383 @@ public class Call(
...
-            logger.d { "[fastReconnect] [RealtimeConnection.Disconnected], call_id:$id" }
-            this@Call.state._connection.value = RealtimeConnection.Disconnected
+    private suspend fun reconnectFast(reason: String): ReconnectOutcome {
+        logger.d { "[reconnectFast] reconnectAttempts=$reconnectAttempts" }
+        val currentSession = session.value
+            ?: return ReconnectOutcome.PreconditionNotMet("No active session for fast reconnect")
+
+        try {
+            val stats = collectStats()
+            currentSession.sendCallStats(stats)
```
rahul-lohra: Could you please explain why is this in a `try-catch`?
I raised this question because I don't get any signal from the method signature that it needs to be put in a try-catch

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -808,24 +792,36 @@ public class RtcSession internal constructor(
...
-            capabilities = call.clientCapabilities.values.toList(),
-            source = ParticipantSource.PARTICIPANT_SOURCE_WEBRTC_UNSPECIFIED,
-        )
+        when (val result = internalConnect(reconnectDetails, options)) {
+            is SfuConnectionResult.Connected -> Unit
+            is SfuConnectionResult.PeerConnectionStale ->
+                throw PeerConnectionNotUsableException()
+            is SfuConnectionResult.Failed ->
+                throw result.error
+        }
```
rahul-lohra: We can make this function return a type. So why are we still throwing exceptions?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1831,70 +1851,74 @@ public class RtcSession internal constructor(
...
+
+        cancelActiveWork()
+
+        // Mark disconnected immediately — late ICE candidates will be routed
+        // to iceTricklePendingEvents instead of being sent to the now-defunct SFU.
+        _sfuSfuSocketState.value = SfuSocketState.Disconnected.Stopped
+
+        // Tell the old SFU we're leaving so it evicts immediately
+        // instead of waiting for the reconnect grace period.
+        leaveWithReason(reason)
```
rahul-lohra: `leaveWithReason` event is silently killed before it's sent because of `cleanup()`

Image

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: yes, not needed here .. removing them

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: yes right..not needed here removing them

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: changing the internalConnect method not to throw anything

PratimMallick: we need a try-finally here because` finalizeMigration()` must run even on cancellation. The internalConnect can throw a cancellationException, on which we should do the cleanUp of migration()

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Since this is a public method, returning a type would break binary compatibility. 
The only caller of this funciton was already catching exception earlier although this method was not throwing anything

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: yes not needed , removed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: yeah good catch. Actually the same problem happens when leave is called from Call class as well

PratimMallick: Fixed it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Deprecated it and used internalConnect everywhere

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnection.kt
```diff
@@ -130,6 +130,11 @@ open class StreamPeerConnection(
         }
     }
 
+    fun isClosed(): Boolean {
```
rahul-lohra: I think we can make it internal?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/module/SfuConnectionModule.kt
```diff
@@ -45,7 +45,8 @@ internal class SfuConnectionModule(
     override val connectionTimeoutInMs: Long,
     override val lifecycle: Lifecycle,
     override val tracer: Tracer,
-    val onSignalingLost: (Error) -> Unit,
+    val onSfuApiError: (Error) -> Unit,
+    val onSfuNetworkFailure: (Throwable) -> Unit,
```
rahul-lohra: I think we should write the meaning of these 2 new apis above the class as kdoc. Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -808,24 +805,38 @@ public class RtcSession internal constructor(
...
+            is SfuConnectionResult.Failed -> throw result.error
+        }
+    }
+
+    /**
+     * Connects to the SFU and suspends until the connection is established or
+     * fails. Returns a typed [SfuConnectionResult] so callers can react to the
+     * outcome without catching exceptions.
+     */
+    internal suspend fun internalConnect(
```
rahul-lohra: Just a tip (optional): writing `conectInternal` is easy to infer via IDE suggestion than `internalConnect`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/network/NetworkStateProvider.kt
```diff
@@ -53,8 +53,7 @@ public class NetworkStateProvider(
 
         override fun onLost(network: Network) {
             logger.d { "[callback#onLost] #network; onLost" }
-            isConnected = false
-            listeners.onDisconnected()
+            notifyListenersIfNetworkStateChanged()
```
rahul-lohra: Can you elaborate what was wrong earlier?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/HealthMonitor.kt
```diff
@@ -45,6 +45,7 @@ internal class HealthMonitor(
     private val userScope: UserScope,
     private val checkCallback: suspend () -> Unit,
     private val reconnectCallback: suspend () -> Unit,
+    private val isNetworkAvailable: () -> Boolean = { true },
```
rahul-lohra: Why default value is set to `true`. Why it has a default value in first case?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/SfuSocketStateService.kt
```diff
@@ -182,40 +155,12 @@ internal class SfuSocketStateService(initialState: SfuSocketState = SfuSocketSta
                     }
 
                     else -> {
-                        logger.e { "Cannot handle event $event while being in inappropriate state $state" }
+                        logger.d { "Cannot handle event $event while being in inappropriate state $state" }
```
rahul-lohra: Is changing from `logger.e` to `logger.d` intentional?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -764,11 +797,13 @@ public class Call(
...
 
     internal suspend fun collectStats(): CallStatsReport {
-        val publisherStats = session.value?.getPublisherStats()
-        val subscriberStats = session.value?.getSubscriberStats()
-        state.stats.updateFromRTCStats(publisherStats, isPublisher = true)
-        state.stats.updateFromRTCStats(subscriberStats, isPublisher = false)
-        state.stats.updateLocalStats()
+        val publisherStats = runCatching { session.value?.getPublisherStats() }.getOrNull()
+        val subscriberStats = runCatching { session.value?.getSubscriberStats() }.getOrNull()
+        runCatching {
```
rahul-lohra, line 804: Why we do we have `runCatching` in the code now? It was not there before.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -787,184 +822,367 @@ public class Call(
...
+                    -> reconnectFast(reason)
+
+                    WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN ->
+                        reconnectRejoin(reason)
+
+                    WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE ->
+                        reconnectMigrate()
+
+                    WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_DISCONNECT ->
+                        error("Handled above")
```
rahul-lohra: I understand how it is handled. Technically it works. 

Using `error()` for a valid strategy path feels slightly misleading since `DISCONNECT` is an expected control signal rather than an exceptional case.

The way short-circuit design is used here for `WEBSOCKET_RECONNECT_STRATEGY_DISCONNECT` seems unusual. We introduced 2 decision paths for the same state, whereas a single unified decision flow would be sufficient and easier to maintain.

I can give you a solution: 
Add a new `ReconnectOutcome.Disconnect`

Handle it in `when(outcome)` ~ Line 952

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -1012,14 +1229,14 @@ public class Call(
...
             safeCall {
-                session.value?.sfuTracer?.trace(
-                    "leave-call",
-                    "[reason=$reason, error=${disconnectionReason?.message}]",
-                )
+                session.value?.sfuTracer?.trace("leave-call", leaveReason)
                 val stats = collectStats()
                 session.value?.sendCallStats(stats)
             }
+            // Must complete before cleanup() cancels the session's supervisor job.
```
rahul-lohra, line 1264: Are we sure that the `session.value?.sendLeaveEvent(leaveReason)` depends on session's supervisor job? Because I don't see it

rahul-lohra (commented):
Please see all above comments

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/utils/ApiCalls.kt
```diff
@@ -16,55 +16,33 @@
...
-        } catch (e: IOException) {
-            Failure(
-                Error.ThrowableError(
-                    e.message ?: "IOException",
-                    e,
-                ),
-            )
-        }
+@Suppress("TooGenericExceptionCaught")
+internal suspend inline fun <T : Any> sfuCall(
```
rahul-lohra: `sfuCall` seem to know nothing about sfu :sweat_smile: . It appears like a simple `safeCall` . 

The name suggests it should be used only with SFUs but the implementation tells me - use it anywhere

Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/RetryableSignalingServiceDecorator.kt
```diff
@@ -0,0 +1,195 @@
...
+        crossinline block: suspend () -> T,
+    ): T {
+        var lastResponse: T? = null
+
+        val response = retryProcessor.retry(policy) {
+            val response = block()
+            lastResponse = response
+            val error = errorExtractor(response)
+            if (error != null && error.should_retry && !error.isSessionFatal()) {
+                throw SfuRetryableException(error)
```
rahul-lohra, line 162: I don't see we are catching `SfuRetryableException` any more. So do we need it ?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnection.kt
PratimMallick: yes agreed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Good tip

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/network/NetworkStateProvider.kt
PratimMallick: The old onLost handler unconditionally set isConnected = false and called listeners.onDisconnected(), regardless of whether the device actually lost all connectivity. The new code delegates to notifyListenersIfNetworkStateChanged() — the same path used by onAvailable and onCapabilitiesChanged — which checks real connectivity via isConnected() before notifying listeners.

In practice, during a WiFi → cellular handoff, the behavior is correct. There's a genuine ~300ms window where the device has no connectivity (WiFi is gone, cellular hasn't been announced yet), so isConnected() correctly returns false and onDisconnected fires. When onAvailable(cellular) arrives shortly after, onConnected fires and triggers a FAST reconnect (elapsed time is well within reconnectDeadlineMillis).

The key improvement is that all three callbacks now go through one unified path that derives state from reality rather than hardcoding it — so edge cases like onLost firing for a non-primary secondary network won't cause spurious disconnect signals.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/HealthMonitor.kt
PratimMallick: yes we dont need a default value. Added the same isNetworkAvailable() call to CordinatorSocket as well

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/SfuSocketStateService.kt
PratimMallick: Yes, this is intentional. The else part happens during normal operation when events arrive during transitions (e.g., a stale health-check event arriving after a disconnect). Logging it as logger.e caused noise in error monitoring and customer logs. logger.d keeps it visible for debugging without alarming. 

The defaultHandler is the FSM's fallback — it fires when an event arrives in a state that doesn't have a specific onEvent handler for it.

The cases in else are  all expected, benign situations — stale or redundant events arriving in states that intentionally don't care about them. The FSM correctly returns state (no transition) for all of them.

Logging these as logger.e was generating noise in error monitoring and customer logs for what is normal operational behavior during connection teardown and transitions. logger.d keeps them visible for debugging without treating them as errors.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: The `runCatching` was added because `getPublisherStats() / getSubscriberStats()` can throw if the peer connection is already closed or in a transitional state during reconnection. Previously this could crash the stats-reporting coroutine, which in turn could interfere with the leave flow. Wrapping in runCatching makes stats collection best-effort — failures are logged as warnings but don't disrupt the call lifecycle.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: yes that's a better way . making the change

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: `sendLeaveEvent` internally uses the SFU socket's `sendEvent()`, which writes to the WebSocket. When `cleanup()` runs, it calls `session.value?.disconnect()` which cancels the session's supervisor job and closes the SFU socket. If sendLeaveEvent hasn't completed before `cleanup()`, the socket gets closed mid-send, and the leave event is lost. That's why the comment and ordering exist — sendLeaveEvent must complete before `cleanup()`. The dependency isn't on the supervisor job directly, but on the resources (SFU socket) that the supervisor job's cancellation tears down.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/utils/ApiCalls.kt
PratimMallick: yes right, changign it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/RetryableSignalingServiceDecorator.kt
PratimMallick: It's not caught explicitly because StreamRetryProcessor catches Throwable generically — the class is used as a signal, not for type-matching. That said, the name could be more generic (e.g., RetrySignalException) since the retry processor itself isn't SFU-specific. 

Adding a KDoc for now, can think of a better way/name later on

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/utils/ApiCalls.kt
rahul-lohra: Okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/utils/RetryableSignalingServiceDecorator.kt
rahul-lohra: Okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: Okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/common/HealthMonitor.kt
rahul-lohra: Okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
rahul-lohra: Okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnection.kt
rahul-lohra: Okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/SfuSocketStateService.kt
PratimMallick: Deleting this log

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/internal/network/NetworkStateProvider.kt
PratimMallick: done

rahul-lohra (approved):
LGTM

PR #1646, opened 2026-04-16
Improve Telecom enabled audio routing

rahul-lohra:
### Goal

Improve Telecom based audio output routing
For telecom based calls, the the audio output routing is owned by telecom which overrides `AudioManager.setCommunicationDevice()`

### Implementation

For telecom registered calls, when a user wants to switch to speaker then we will coordinate with Telecom API to switch the audio output via
```kotin
telecomCall.processAction(TelecomCallAction.SwitchAudioEndpoint(...))
```

### 🎨 UI Changes

None

### Testing

Depends on PR: https://github.com/GetStream/stream-video-android/pull/1650

1. Click on direct call 
2. select user to make audio/video call

3. Check if telecom is enabled for audio/video call (its enabled by default in demo-app)
like this
```kotlin
        val callServiceConfigRegistry = CallServiceConfigRegistry()
        callServiceConfigRegistry.apply {
            register(
                CallType.AudioCall.name,
                DefaultCallConfigurations.audioCall.copy(enableTelecom = false),
            )

        return StreamVideoBuilder(
            callServiceConfigRegistry = callServiceConfigRegistry,
```

2. Make an audio/video-call
3. Once the call accepted, then try to toggle speaker and see the behaviour

## Summary by CodeRabbit

* **New Features**
  * Integrated Android Telecom framework for audio endpoint routing during device selection.

* **Bug Fixes**
  * Modified microphone enable/disable control flow to provide more immediate response.

* **Tests**
  * Added comprehensive test coverage for microphone management state transitions, device selection behaviors, USB audio handling, and audio bitrate configuration.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
@@ -941,9 +941,6 @@ class MicrophoneManager(
             var capturedOnAudioDevicesUpdate = onAudioDevicesUpdate
 
             if (setupCompleted) {
-                logger.d {
-                    "[setup] Already complete (preferSpeaker=$preferSpeaker), invoking callback immediately"
-                }
                 capturedOnAudioDevicesUpdate?.invoke()
                 capturedOnAudioDevicesUpdate = null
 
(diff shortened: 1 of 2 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
```diff
@@ -648,12 +650,10 @@ class MicrophoneManager(
      * Enable or disable the microphone
      */
     fun setEnabled(enabled: Boolean, fromUser: Boolean = true) {
```
aleksandar-apostolov, line 654: `enforceSetup` was wrapping this before — do `enable()` and `disable()` already guard against being called before `setup()`? If not, calling `setEnabled()` before setup could NPE on the audio track.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
```diff
@@ -678,6 +686,46 @@ class MicrophoneManager(
...
+     * Telecom endpoint API so the routing is actually honoured.
+     */
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun switchTelecomEndpointForDevice(device: StreamAudioDevice?) {
+        val telecomCall = mediaManager.call.state.jetpackTelecomRepository
+            ?.currentCall?.value as? TelecomCall.Registered ?: return
+
+        val targetEndpointType = when (device) {
+            is StreamAudioDevice.Speakerphone -> CallEndpointCompat.TYPE_SPEAKER
+            is StreamAudioDevice.Earpiece -> CallEndpointCompat.TYPE_EARPIECE
```
aleksandar-apostolov, line 704: This silently skips unmapped device types (e.g. USB). Is this intentional not to support the USB device type?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
rahul-lohra: `enable()` and `disable()` already enforce setup internally, so the guard is still preserved.

rahul-lohra:
```kotlin
internal fun enable(fromUser: Boolean = true) {
        enforceSetup {
            if (fromUser) {
                _status.value = DeviceStatus.Enabled
            }
            mediaManager.audioTrack.trySetEnabled(true)
        }
    }
    
  fun disable(fromUser: Boolean = true) {
        enforceSetup {
            if (fromUser) {
                _status.value = DeviceStatus.Disabled
            }
            mediaManager.audioTrack.trySetEnabled(false)
        }
    }
    
```

aleksandar-apostolov: Awesome.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/MediaManager.kt
rahul-lohra: Good point, I need to test this PR with a USB attached mic first.

rahul-lohra: We investigated this and found that some external USB microphones like _DJI mic 3_ require additional handling independent of telecom. For now, we’ve decided to skip implementing audio routing changes for external USB microphones in telecom-based calls. We can revisit this if a concrete requirement arises.

aleksandar-apostolov (approved):

PR #1648, opened 2026-04-16
Move audio switching logic to execute on another thread

rahul-lohra:
### Goal

1. Move audio switching logic to execute on another thread
2. Remove duplicate IPC calls

We were making 4 IPC calls when toggling the speaker
```
→ SpeakerManager.setSpeakerPhone(true)
  → MicrophoneManager.select(Speakerphone)
    → AudioSwitchHandler.selectDevice(Speakerphone)
        → audioSwitch.selectDevice()           ← triggers enumerateDevices()
            → since state=ACTIVATED: activate() ← IPC #1: clearCommunicationDevice()
                                                  IPC #2: setCommunicationDevice()
        → audioSwitch.activate()  [explicit]   ← IPC #3: clearCommunicationDevice()
                                                  IPC #4: setCommunicationDevice()
```
This was causing 50-60 framedrops

### Implementation

We have changed few internal implementation
1. We switch the audio switch work from `main thread` to a new thread `audio-switch-thread`
2. Removed the duplicate IPC calls

We replaced `HandlerThread` with `Couroutine` backed by `SingleThreadExecuter`. Its lifecycle is cleaned up when `StreamClient` is cleaned up

We separated AudioSwitch core logic from concurrency. So we moved the core logic in `AudioSwitchController` and concurrency logic in `AudioHandler`

AudioExecutionContext - A new class to provide coroutine scope for `AudioSwitch`

### 🎨 UI Changes

None

### Testing

Depends on PR: https://github.com/GetStream/stream-video-android/pull/1650, https://github.com/GetStream/stream-video-android/pull/1646

1. Click on direct call 
2. select user to make audio/video call


rahul-lohra:
@coderabbitai Analyse this PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/AudioExecutionContext.kt
```diff
@@ -0,0 +1,47 @@
...
+        }.asCoroutineDispatcher()
+    }
+
+    private val scope: CoroutineScope by lazy {
+        CoroutineScope(SupervisorJob() + dispatcher)
+    }
+
+    fun createChildScope(): CoroutineScope {
+        return CoroutineScope(scope.coroutineContext + Job())
+    }
```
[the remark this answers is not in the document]
rahul-lohra: Done as per your suggestion

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/AudioHandler.kt
```diff
@@ -39,8 +40,12 @@ public interface AudioHandler {
 }
 
 /**
- * TODO: this class should be merged into the Microphone Manager
+ * TODO Remove this class on v2
  */
```
aleksandar-apostolov, line 44: Should the `ReplaceWith` target be `AudioSwitchDecorator` instead of `AudioSwitchControllerDecorator`? IDE users who apply the quick-fix will get a broken import.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/AudioSwitchDecorator.kt
```diff
@@ -0,0 +1,67 @@
...
+    private var scope: CoroutineScope? = null
+
+    /**
+     * Synchronized to make scope initialization and replacement atomic.
+     * Prevents races where multiple scopes are created or a cancelled scope
+     * is reused, leading to silent no-op executions.
+     * Not a hot path, so the overhead is negligible.
+     */
+    @Synchronized
+    private fun ensureScope(): CoroutineScope? {
```
aleksandar-apostolov, line 38: Would it make sense to inject `AudioExecutionContext` via constructor instead of pulling it from the `StreamVideo` singleton here? The controller is already injected — this is the only dependency that reaches through the global. Would simplify testing too (no `mockkObject(StreamVideo)` needed).

rahul-lohra: Yes, injecting `AudioExecutionContext` via constructor would be cleaner and improve testability.

However, both `MediaManagerImpl` and `MicrophoneManager` currently expose public constructors, so introducing a new constructor dependency would be a breaking API change across the chain:

```
Call → MediaManagerImpl (public constructor) → MicrophoneManager (public constructor)
```

Because of that constraint, we can’t safely push this dependency through constructors without widening the public API surface or forcing changes downstream.

If we decide to make these constructors internal in the future, then we can move this to proper constructor injection.

A cleaner long-term fix would be:
	•	Deprecate the current public constructors
	•	Introduce new constructors (or factory APIs) with proper dependency injection
	•	Make the actual constructors internal once migration is complete

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/AudioHandler.kt
rahul-lohra: Done updated with this 

```kotin
@Deprecated(
    "Replaced by AudioSwitchDecorator + AudioSwitchController which provide coroutine-backed audio handling",
    ReplaceWith(
        "AudioSwitchDecorator(AudioSwitchController(context, preferredDeviceList, audioDeviceChangeListener))",
    ),
)
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/audio/AudioSwitchDecorator.kt
aleksandar-apostolov: Fine.

aleksandar-apostolov (approved):

PR #1652, opened 2026-04-20
Improve USB audio device selection and cleanup lifecycle

PratimMallick:
### Goal

Fixes two issues in USB device selection:
1. If the USB device was connected before joining the call, it wasnt getting selected automatically as the AudioDeviceManager of webrtc wasnt ready when the usb devices discovery ran 
2. On removing of usb mic mid call, the selection wasnt cleared leading to non selection when it was connected back again mid call

### Implementation

- Fix NPE crash when clearing USB device selection — `clearUsbDeviceSelection()` was passing `null` 
  to the WebRTC ADM's `setPreferredInputDevice`, which internally calls `getId()` on the null reference. 
  Now resets state directly since Android handles audio fallback automatically on device removal.

- Queue USB mic preference when ADM is not yet initialized — if `setPreferredAudioInputDevice` is 
  called before the factory is created, the preference is stored and applied once the ADM becomes available.

- Remove unnecessary `@RequiresApi(Build.VERSION_CODES.M)` annotations and version gate in `cleanup()` — 
  the `AudioDeviceCallback` APIs used here are safe without the annotation since the callback is only 
  registered when the API is available.
### Testing

- [ ] Connect a USB input-only microphone (e.g. Rode Wireless Go II) to an Android device
- [ ] Join a call and verify logs show `[reapplyUsbDevicePreference] Re-applying USB device after AudioSwitch activation: <device name>` after each AudioSwitch callback
- [ ] Verify audio is captured from the USB mic, not the built-in phone mic
- [ ] Toggle speakerphone on/off and verify USB mic input is preserved
- [ ] Disconnect the USB mic and verify fallback to built-in mic works correctly
- [x] Run `./gradlew :stream-video-android-core:testDebugUnitTest` — all tests pass

#### General

#### Code & documentation

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

## Bug Fixes
- Fixed an issue where the selected USB audio input device would unexpectedly reset when audio settings or modes changed during an active call. The app now properly maintains your preferred USB input device selection and ensures it remains active consistently, even as different audio configurations are applied or detected.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
@@ -97,6 +97,8 @@ public class StreamPeerConnectionFactory(
     // Provider function to check if microphone is enabled
     private var microphoneEnabledProvider: (() -> Boolean)? = null
 
+    internal var onAudioRecordStartCallback: (() -> Unit)? = null
+
     /**
      * Set to get callbacks when audio input from microphone is received.
      * This can be example used to detect whether a person is speaking
@@ -175,6 +177,12 @@ public class StreamPeerConnectionFactory(
 
     private var adm: JavaAudioDeviceModule? = null
 
+    @Volatile
+    private var pendingPreferredInputDevice: AudioDeviceInfo? = null
+
+    @Volatile
+    private var hasPendingPreferredDevice = false
+
     private fun createFactory(): PeerConnectionFactory {
         PeerConnectionFactory.initialize(
             PeerConnectionFactory.InitializationOptions.builder(context)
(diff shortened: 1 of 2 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -345,17 +359,23 @@ public class StreamPeerConnectionFactory(
...
-            audioLogger.i {
-                "[setPreferredAudioInputDevice] Set preferred input device: ${deviceInfo?.productName}"
+            val currentAdm = adm
+            if (currentAdm != null) {
+                currentAdm.setPreferredInputDevice(deviceInfo)
+                audioLogger.i {
+                    "[setPreferredAudioInputDevice] Set preferred input device: ${deviceInfo?.productName}"
+                }
+            } else {
+                pendingPreferredInputDevice = deviceInfo
```
rahul-lohra, line 374: Do you think we should return something else instead of `true`?
I think this function can return 3 results - True, Queued, Fail

PratimMallick: I feel this is fine for now as the caller of the method does not need to know the internal implementation. It just needs to know whether the api was able to process the request or not. if later we see a need that the caller has to made a decision based on whther the request was queued or not we can add the third state

rahul-lohra: Cool

rahul-lohra (approved):

Issue #1657, opened 2026-04-27
RemoteParticipant displays mirroring screen


ataballyyeva:
**Describe the bug**
when Android user joins call first, User B video shows mirrored

**SDK version**
-1.22.0

**To Reproduce**
Pre-conditions: User A on android device

1. User A  start call
2. User A join into Call room first
3. User B join into call room

**Expected behavior**
User B video should always show in normal video view. Attached screenshot that shows both local and joined users in mirror mode.

**Device:**
 - Pixel 7A, Pixel 9 PRO
 - OS16, OS15
**Screenshots**

<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/ae4bed21-32af-476e-a647-abc76252771d" />

ataballyyeva:
I locally added some of Stream components into demo app to reproduce the issue, just copy and paste the MainActivity in demo app with below version:

import android.os.Bundle
import android.widget.Toast
import androidx.activity.ComponentActivity
import androidx.activity.compose.setContent
import androidx.compose.foundation.background
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.foundation.layout.padding
import androidx.compose.foundation.layout.size
import androidx.compose.runtime.collectAsState
import androidx.compose.runtime.getValue
import androidx.compose.runtime.mutableStateOf
import androidx.compose.runtime.remember
import androidx.compose.runtime.setValue
import androidx.compose.ui.Alignment
import androidx.compose.ui.Modifier
import androidx.compose.ui.draw.clip
import androidx.compose.ui.graphics.Color
import androidx.compose.ui.unit.IntSize
import androidx.compose.ui.unit.dp
import io.getstream.video.android.compose.permission.LaunchCallPermissions
import io.getstream.video.android.compose.theme.VideoTheme
import io.getstream.video.android.compose.ui.components.call.activecall.CallContent
import io.getstream.video.android.compose.ui.components.call.controls.ControlActions
import io.getstream.video.android.compose.ui.components.call.controls.actions.FlipCameraAction
import io.getstream.video.android.compose.ui.components.call.controls.actions.ToggleCameraAction
import io.getstream.video.android.compose.ui.components.call.controls.actions.ToggleMicrophoneAction
import io.getstream.video.android.compose.ui.components.call.renderer.FloatingParticipantVideo
import io.getstream.video.android.compose.ui.components.call.renderer.LayoutType
[...]

PR #1665, opened 2026-04-30
Add mirrorStream support to ParticipantVideo for video mirroring

rahul-lohra:
### Goal

We currently do not provide any API to mirror participant video streams. This limitation was raised in Stream Video Android Issue https://github.com/GetStream/stream-video-android/issues/1657

Mirroring is a common requirement, especially for:

* front camera preview parity (self-view)
* UI consistency across platforms

### Implementation

This PR introduces a new parameter `mirrorMode` in `ParticipantVideo`, allowing developers to control horizontal mirroring of the rendered video stream.

```kotlin
public enum class MirrorMode { AUTO, ALWAYS, NEVER }

mirrorStream: MirrorMode = MirrorMode.AUTO

val mirror = when (mirrorStream) {
    MirrorMode.AUTO -> cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId
    MirrorMode.ALWAYS -> true
    MirrorMode.NEVER -> false
}
```

The mirroring is applied at the rendering layer using the underlying WebRTC renderer, ensuring minimal performance overhead.

### Usage

```kotlin

CallContent(
    videoRenderer = { modifier, call, participant, style ->
        val isSelf = participant.sessionId == me?.sessionId
        ParticipantVideo(mirrorMode = if (isSelf) MirrorMode.ALWAYS else MirrorMode.NEVER)
     }
} 
```

Notes

* Mirroring is handled at the renderer level (not via UI transforms), avoiding additional composition cost.
* Defaults are preserved; existing behavior remains unchanged unless explicitly enabled.
* Provides flexibility without exposing rendering internals to the UI layer.

### 🎨 UI Changes

None

### Testing

Join a video call with 2 people, see that the other user is not mirrored

## Summary by CodeRabbit

* **New Features**
  * Improved video participant rendering with enhanced mirroring behavior. Remote participants' video streams are now properly mirrored for optimal display orientation, while local participant video is rendered appropriately without unnecessary mirroring adjustments in video calls.


stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
```diff
@@ -260,14 +265,14 @@ public fun ParticipantVideoRenderer(
     val video by participant.video.collectAsStateWithLifecycle()
     val cameraDirection by call.camera.direction.collectAsStateWithLifecycle()
     val me by call.state.me.collectAsStateWithLifecycle()
-    val mirror by remember(cameraDirection) {
+    val mirror by remember(cameraDirection, mirrorStream) {
         mutableStateOf(
-            cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId,
+            mirrorStream || (cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId),
         )
```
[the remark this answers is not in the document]
rahul-lohra: Updated

aleksandar-apostolov:
**Q: Should `mirrorStream` be an enum rather than a `Boolean`?**

Looking at the resulting logic:

```kotlin
val mirror = mirrorStream || (cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId)
```

The new parameter is **add-only**: a caller can force mirroring on by passing `true`, but cannot force it off — passing `false` falls through to the implicit front-camera self-view rule. That means there's no way to say *"I'm the local user with the front camera, and I don't want my self-view mirrored."*

Would it be worth modeling this as an explicit tri-state instead?

```kotlin
public enum class MirrorMode { AUTO, ALWAYS, NEVER }

mirrorStream: MirrorMode = MirrorMode.AUTO

val mirror = when (mirrorStream) {
    MirrorMode.AUTO -> cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId
    MirrorMode.ALWAYS -> true
    MirrorMode.NEVER -> false
}
```

Pros: each call site reads itself (`MirrorMode.AUTO` vs. `MirrorMode.NEVER`), and we don't burn the param name on a `Boolean` we'd later want to upgrade. A `Boolean?` would also work as a smaller change, but a public Compose API named `mirrorStream: Boolean?` is harder to reason about than an enum at every call site.

Curious what you'd prefer here — or whether the add-only semantics are intentional for backwards compat?

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
```diff
@@ -484,6 +484,7 @@ fun CallScreen(
                                         ),
                                     )
                                 },
+                                mirrorStream = (participant.sessionId != me?.sessionId),
```
aleksandar-apostolov: The PR description's testing note says *"see that the other user is not mirrored"*, but this line passes `mirrorStream = true` for every remote participant — which forces remote streams to be mirrored (their right hand appears on our left). Was the intent to drop this line entirely and rely on the default `false`, or was the condition meant to be inverted? Setting it on remote participants typically isn't what we want, since their stream is already correctly oriented as received from the SFU.

stream-video-android-ui-compose/api/stream-video-android-ui-compose.api
```diff
@@ -1421,8 +1421,8 @@ public final class io/getstream/video/android/compose/ui/components/call/rendere
...
-	public static final fun ParticipantVideoRenderer (Lio/getstream/video/android/core/Call;Lio/getstream/video/android/core/ParticipantState;Lio/getstream/video/android/compose/ui/components/video/VideoScalingType;Lkotlin/jvm/functions/Function3;Landroidx/compose/runtime/Composer;II)V
+	public static final fun ParticipantVideo (Lio/getstream/video/android/core/Call;Lio/getstream/video/android/core/ParticipantState;Landroidx/compose/ui/Modifier;Lio/getstream/video/android/compose/ui/components/call/renderer/VideoRendererStyle;Lkotlin/jvm/functions/Function4;Lkotlin/jvm/functions/Function4;Lio/getstream/video/android/compose/ui/components/video/VideoScalingType;ZLkotlin/jvm/functions/Function3;Lkotlin/jvm/functions/Function4;Lkotlin/jvm/functions/Function6;Landroidx/compose/runtime/Composer;III)V
```
aleksandar-apostolov: The new `Boolean` parameter is inserted in the middle of the argument list, so the binary signature changes (`Z` added at position 8, default-args bitmask grows from `II` → `III`). Source-level callers using named arguments are fine, but anything compiled against the old `.aar` will break at link time. Could we move `mirrorStream` to the end of the parameter list to keep ABI stability — or is this PR intentionally part of a v2 break?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
```diff
@@ -110,6 +110,7 @@ import kotlinx.coroutines.delay
  * @param labelContent Content is shown that displays participant's name and device states.
  * @param connectionIndicatorContent Content is shown that indicates the connection quality.
  * @param videoFallbackContent Content is shown the video track is failed to load or not available.
+ * @param mirrorStream Whether to mirror the video stream horizontally.
```
aleksandar-apostolov: Tiny one — the new `@param mirrorStream` is between `videoFallbackContent` and `reactionContent` in the KDoc, but in the actual signature it sits between `scalingType` and `videoFallbackContent`. Should we re-order the doc to match the signature?

aleksandar-apostolov: Should the KDoc mention that `mirrorStream` is OR'd with the implicit front-camera self-view rule? A reader passing `false` might expect mirroring to be off, but the implicit `Front && self` mirror still applies — worth flagging in the docs (this is also what the top-level comment about an enum ties into).

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
```diff
@@ -260,14 +265,11 @@ public fun ParticipantVideoRenderer(
     val video by participant.video.collectAsStateWithLifecycle()
     val cameraDirection by call.camera.direction.collectAsStateWithLifecycle()
     val me by call.state.me.collectAsStateWithLifecycle()
-    val mirror by remember(cameraDirection) {
-        mutableStateOf(
-            cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId,
-        )
-    }
+    val mirror =
+        mirrorStream || (cameraDirection == CameraDirection.Front && me?.sessionId == participant.sessionId)
```
aleksandar-apostolov: We're dropping the previous `remember(cameraDirection) { mutableStateOf(...) }` wrapper. The cost is trivial (a Boolean OR on each recomposition), and arguably the new code is more correct — the old `remember` only keyed on `cameraDirection`, so it would have missed updates to `me?.sessionId` or `participant.sessionId`. Just confirming this was intentional?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
```diff
@@ -260,14 +265,11 @@ public fun ParticipantVideoRenderer(
...
     val videoRendererConfig = remember(mirror, scalingType, videoFallbackContent) {
         videoRenderConfig {
-            mirrorStream = mirror
+            this.mirrorStream = mirror
```
aleksandar-apostolov: The `this.` qualifier here is needed because the new function parameter shadows the builder's `mirrorStream` property. Should we rename the parameter (or change to the enum from the top-level comment) so this shadowing goes away? Same name at two layers is a small footgun for future readers.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
rahul-lohra: Yes, this is intentional

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
rahul-lohra: We can proceed with the enum proposal
```kotlin
 public enum class MirrorMode { AUTO, ALWAYS, NEVER }
 ```

stream-video-android-ui-compose/api/stream-video-android-ui-compose.api
rahul-lohra: Thanks for pointing, this PR is not a breaking change

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
rahul-lohra: Resolved, please check new kdoc

rahul-lohra: Resolved, please check new kdoc

stream-video-android-ui-compose/api/stream-video-android-ui-compose.api
rahul-lohra: Resolved

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallScreen.kt
rahul-lohra: Removed

aleksandar-apostolov (approved):
All review feedback addressed cleanly:

- `Boolean` → `MirrorMode` enum (`AUTO`/`ALWAYS`/`NEVER`) — exactly the right shape
- Binary compat preserved via `@Deprecated` overload delegating to the new signature
- KDoc reordered to match signature, with explicit documentation of `AUTO` semantics
- Demo app no longer force-mirrors remote participants
- Parameter rename eliminated the `this.mirrorStream` shadowing

Nice iteration. LGTM.

PR #1669, opened 2026-05-05
Add RingingCallJoinInterceptor hook for ringing call activation

rahul-lohra:
### Goal

Add a join interception hook that lets apps delay or abort call entry after the join response has been applied locally, so participant readiness/status can be synchronized before the user is treated as fully in-call. Mirrors the iOS CallJoinInterceptor API. [Swift PR link](https://github.com/GetStream/stream-video-swift/pull/1108) 

### Implementation

```kotlin
public interface CallJoinInterceptor {
      //Called when the SDK is ready to transition to [RingingState.Active].
     // Throw [CallJoinInterceptionException] to abort the join — the SDK will leave the call cleanly and surface the failure.
      public suspend fun callReadyToJoin(call: Call) // will run for max 5 second
}

class Call {
    suspend fun join(
        ....
        callJoinInterceptor: CallJoinInterceptor? = null,
    ):Result<RtcSession> {
    ...
     state.callJoinInterceptor = callJoinInterceptor
    }
}        
```

- New `CallJoinInterceptor` interface with one suspend method `callReadyToJoin(call)`. Returning from it triggers the transition; the SDK auto-proceeds after 5 s if it doesn't return. 
- Integrators can throw `CallJoinInterceptionException` from `callReadyToJoin` to safely  abort the call.
- `Call.join()` accepts an optional `callJoinInterceptor` that will run for all types of call.
- `CallJoinInterceptor` is invoked right before transitioning to `RingingState.Active` with timeout of 5 seconds. 

## Usage

```kotlin
val interceptor = object : CallJoinInterceptor {
    override suspend fun callReadyToJoin(call: Call) {
        // 1. Signal that this client is ready and waiting
        call.sendCustomEvent(
            mapOf(
                "type" to "client_ready",
                "user_id" to currentUserId,
            ),
        )

        // 2. Wait for the corresponding "ready_to_join" event
        //    (sent by the server, an admin, or the other party)
        call.events
            .filterIsInstance<CustomVideoEvent>()
            .first { it.custom["type"] == "ready_to_join" }
    }
}

call.join(....callJoinInterceptor = interceptor)
```

### 🎨 UI Changes

None

### Testing
Todo

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
@@ -751,7 +751,7 @@ public class CallState(
     internal var incomingNotificationData = IncomingNotificationData(emptyMap())
     private val ringingLogger by taggedLogger("RingingState")
 
-    internal var ringingCallJoinInterceptor: RingingCallJoinInterceptor? = null
+    internal var ringingCallActivationInterceptor: RingingCallActivationInterceptor? = null
 
     fun handleEvent(event: VideoEvent) {
         logger.d { "[handleEvent] ${event::class.java.name.split(".").last()}" }
@@ -1386,7 +1386,7 @@ public class CallState(
             activeStateGate.awaitAndTransition(
                 ringingState.value,
                 call,
-                ringingCallJoinInterceptor,
+                ringingCallActivationInterceptor,
             ) {
                 _ringingState.value = state
                 activeStateGate.cleanup()
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
@@ -751,6 +751,8 @@ public class CallState(
     internal var incomingNotificationData = IncomingNotificationData(emptyMap())
     private val ringingLogger by taggedLogger("RingingState")
 
+    internal var ringingCallJoinInterceptor: RingingCallJoinInterceptor = DefaultRingingCallJoinInterceptor
+
     fun handleEvent(event: VideoEvent) {
         logger.d { "[handleEvent] ${event::class.java.name.split(".").last()}" }
 
@@ -1381,7 +1383,11 @@ public class CallState(
         ringingLogger.d { "Update: $state" }
 
         if (state is RingingState.Active) {
-            activeStateGate.awaitAndTransition(ringingState.value, call) {
+            activeStateGate.awaitAndTransition(
+                ringingState.value,
+                call,
+                ringingCallJoinInterceptor,
+            ) {
                 _ringingState.value = state
                 activeStateGate.cleanup()
             }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
@@ -751,7 +751,7 @@ public class CallState(
     internal var incomingNotificationData = IncomingNotificationData(emptyMap())
     private val ringingLogger by taggedLogger("RingingState")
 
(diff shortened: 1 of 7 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1381,7 +1383,11 @@ public class CallState(
...
 
         if (state is RingingState.Active) {
-            activeStateGate.awaitAndTransition(ringingState.value, call) {
+            activeStateGate.awaitAndTransition(
+                ringingState.value,
+                call,
+                ringingCallJoinInterceptor,
+            ) {
                 _ringingState.value = state
                 activeStateGate.cleanup()
```
aleksandar-apostolov, line 1392: `Call.join()` writes `ringingCallJoinInterceptor` on `CallState`, but it's never cleared after the transition completes. Integrator interceptors typically capture UI/Activity context in lambdas — for a long-lived `Call` this is a real retention path past the join.

```suggestion
                activeStateGate.cleanup()
                ringingCallJoinInterceptor = null
```

aleksandar-apostolov:
The PR description says "Mirrors the iOS CallJoinInterceptor API" ([Swift PR #1108](https://github.com/GetStream/stream-video-swift/pull/1108)), but the implementations diverge on three points that are worth resolving before a public API ships, since renaming and changing semantics post-release is painful. The most concerning is scope: `Call.join()` exposes the interceptor parameter on every join, but for a livestream or direct join Android silently ignores it — no error, no log, no compile-time signal.

1. **Scope** — iOS runs the interceptor for all joins; Android only for ringing calls (`Incoming`/`Outgoing` history). Either match iOS, or move the parameter off the general `Call.join()` signature so it can't be silently dropped.
2. **Abort semantics** — iOS lets the interceptor throw to abort the join (`async throws` → `CallJoinInterceptionError`, call left cleanly). Android catches and proceeds, so an integrator who needs to bail out has no path.
3. **Method name** — iOS uses `callReadyToJoin(_ call: Call)`; Android uses `callReadyToJoinWithTimeout(call: Call)`. iOS deliberately dropped the suffix because the timeout is implementation detail (already documented in the KDoc).

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -751,6 +751,8 @@ public class CallState(
     internal var incomingNotificationData = IncomingNotificationData(emptyMap())
     private val ringingLogger by taggedLogger("RingingState")
 
+    internal var ringingCallJoinInterceptor: RingingCallJoinInterceptor? = null
```
aleksandar-apostolov: Add `@Volatile` — written from one coroutine context, read from another, no memory fence guarantees the reader sees the write.

```suggestion
    @Volatile
    internal var ringingCallJoinInterceptor: RingingCallJoinInterceptor? = null
```

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/ActiveStateGateTest.kt
```diff
@@ -359,27 +346,201 @@ class ActiveStateGateTest {
...
+                coroutineScope = this,
+                previousRingingStates = setOf(incoming),
+                timeoutMs = 5_000L,
+            )
+            val transitioned = mutableListOf<Unit>()
+
+            sut.awaitAndTransition(incoming, call, interceptor) { transitioned += Unit }
+            pubState.value = PeerConnection.PeerConnectionState.CONNECTED
+
+            assertEquals(1, interceptorCalled.size)
```
aleksandar-apostolov, line 401: Test name claims ordering but the assertions only check both fired once — would pass even if `onReady` fired first. Record both into a single `List<String>` and assert the order to lock the contract the name promises.

rahul-lohra:
Going with a separate PR: https://github.com/GetStream/stream-video-android/pull/1679
It showcases an example usage of `CallJoinInterceptor` with logic mirrored from iOS using `ActiveStateGate

PR #1683, opened 2026-05-15
Improve pinned participant state on rejoin with new session

rahul-lohra:
### Goal

Server-side pin synchronization was previously tightly coupled to `sessionIds` only. This created issues during participant reconnection flows where a participant could reconnect with a new `sessionId` while still representing the same user.

The goal of this change is to make server pin resolution resilient to reconnect scenarios while preserving the server-defined pin priority order.

Additionally:

- Improve determinism and testability of timestamp handling by introducing `TimeProvider`
- Preserve stable pin ordering across updates
- Avoid ambiguous reconnect matching when multiple pins share the same userId
- `ParticipantJoined` SFU event now will carry `is_pinned` property so you can use it for updating the client-side state correctly

### Implementation

React [PR](https://github.com/GetStream/stream-video-js/pull/2190)

- Add a property `is_pinned` in ParticipantJoinedEvent

## Pin Resolution

A new PinManager has been introduced as the central owner of:

* local pin state
* server pin state
* reconnect-aware pin resolution

The main entry points are:

image

## New PinManager class (pinning/PinManager.kt)

- Centralises all pin mutations — `pin()`, `unpin()`, `setServerPins()`, `onParticipantJoined()` — that were previously scattered across `CallState`
- Mirrors JS SDK's setServerSidePins: assigns `pinnedAt = now + (pins.size - index)` so the server's priority order is preserved as a descending timestamp sequence
- Preserves an existing pinnedAt when a participant is already server-pinned (JS guard: typeof participant.pin?.pinnedAt !== 'number')
- Reconnection support: builds a dual-key lookup (`sessionId` → exact match, `userId` → fallback). Duplicate userId entries are dropped so the wrong participant is never pinned on reconnect

## CallState cleanup

- `pin()` / `unpin()` delegate to `PinManager`
- `updateServerSidePins()` private function removed — replaced by `PinManager.setServerPins()`
- `SessionId` typealias introduced for readability

There are auto generated files, which are not the core part of this PR
1. `SignalLostSignalingServiceDecorator.kt`
2. `SignalingServiceTracerDecorator.kt`
3. go files

### 🎨 UI Changes

None

### Scenario A: Pin propagation on join
- [ ] Start a new call from Web with User A
- [ ] User B joins the call
- [ ] On User A, perform "Pin Everyone" on User B
- [ ] User C joins the call
- [ ] Verify User C sees User B pinned
Status: Passed ✅
---
### Scenario A – Rejoin Flow (User C)
- [ ] User C leaves the call
- [ ] User C rejoins
- [ ] Verify `On User C` still sees User B pinned
*Status*: Passed ✅
---
### Scenario A – Rejoin Flow (User B)
Step 1: User B leaves
- [ ] User B leaves the call
- [ ] Verify `on User C` that no participants are pinned
*Status*: Passed ✅
---
Step 2: User B rejoins flow
- [ ] User B rejoins the call
- [ ] Verify `on User B` that User B appears pinned
Observed:
- User B is not pinned
- Spotlight UI is rendered
*Status*: Passed (expected due to known issue) ✅
*Reason*: Pins from JoinCallResponseEvent may contain a different sessionId, but the fallback userId matching successfully restores the pin association.

---
Step 3: User B rejoins flow
State on User C
- [ ] Verify `on User C` still sees User B pinned
*Status*: Passed ✅
---

## Summary by CodeRabbit

* **New Features**
  * Added metrics and telemetry support for media performance monitoring and RTP statistics tracking.
  * Enhanced participant pinning functionality for improved video call organization and state management.
  * Added support for Python and Vision Agents SDK types.

* **Improvements**
  * Expanded participant information tracking with WebRTC version details and media negotiation identifiers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-video-android/pull/1683)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/pinning/PinManager.kt
```diff
@@ -0,0 +1,178 @@
...
+        _localPins.value = localPins.value - sessionId
+    }
+
+    fun setServerPins(pins: List<PinUpdate>) {
+        _serverPins.value = resolveServerPins(pins, buildPinnedAtLookup(pins))
+    }
+
+    fun onParticipantJoined(event: ParticipantJoinedEvent) {
+        val sessionId = event.participant.session_id
+        if (participants().containsKey(sessionId) && event.isPinned) {
```
aleksandar-apostolov: `CallState.handleEvent` debounces `ParticipantJoinedEvent`s into `pendingParticipantsJoined` when there are more than 8 participants — the new participant isn't added to `internalParticipants` until the debounce fires. But `pinManager.onParticipantJoined(event)` runs immediately after, so this guard sees `containsKey(sessionId) == false` and silently drops the `isPinned` flag.

So in a large call (≥8 participants — exactly when pinning matters most), a freshly-joined pinned user wouldn't get pinned until the next `PinsUpdatedEvent` rebuilds state from the server. Is that intentional, or should the pin be buffered alongside the participant in the debounce path?

aleksandar-apostolov (approved):
LGTM

PR #1704, opened 2026-06-03
Adopt createGuest's server-issued identity for guest users

aleksandar-apostolov:
Stacked on #1703. Merge after parent.

### Goal
Refs [AND-1202](https://linear.app/stream/issue/AND-1202/guest-user-type-is-broken) — keep the SDK's in-memory guest user in sync with the server-issued identity returned by `createGuest`, matching the JS SDK's `connectUser(response.user, response.access_token)` semantics.

The SDK previously discarded `response.user` and only kept `response.accessToken`. If the server resolves a guest id to a different value than what the integrator passed in, the WS auth payload (built from `client.user.id`) and the JWT `user_id` claim drift apart.

### Implementation
- `setupGuestUser` now maps `response.user` via `UserResponse.toUser()`, re-applies `UserType.Guest`, and writes the result back to `StreamVideoClient.user`.
- `StreamVideoClient.user` becomes `var`; `userId` becomes `get() = user.id` so every reader (`CallState`, `ParticipantState`, ringing-state checks) picks up the new identity.
- `CoordinatorSocketConnection.user` becomes `var`, set from the `connectData`/`data` argument inside `connect()`/`reconnect()` so `onCreated()`'s WS auth payload reflects the latest user.

### Testing
- New tests in `StreamVideoClientTest`:
  - `userId tracks the current user reference` — the computed property.
  - `setupGuestUser adopts the server-issued user identity on success` — drives `setupGuestUser` with a mocked `api.createGuest` returning a different id; asserts `client.user.id`, `client.userId`, and preserved `UserType.Guest`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed guest user identity synchronization to correctly adopt server-issued credentials, preventing potential identity mismatches after guest account creation.
  * Improved user identity tracking to dynamically reflect the current user state rather than a fixed value captured at initial connection.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
@@ -126,9 +126,11 @@ import io.getstream.video.android.core.utils.safeSuspendingCallWithResult
 import io.getstream.video.android.core.utils.toEdge
 import io.getstream.video.android.core.utils.toQueriedCalls
 import io.getstream.video.android.core.utils.toQueriedMembers
+import io.getstream.video.android.core.utils.toUser
 import io.getstream.video.android.model.ApiKey
 import io.getstream.video.android.model.Device
 import io.getstream.video.android.model.User
+import io.getstream.video.android.model.UserType
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
@@ -165,7 +167,7 @@ internal const val defaultAudioUsage = AudioAttributes.USAGE_VOICE_COMMUNICATION
 internal class StreamVideoClient internal constructor(
     override val context: Context,
     internal val scope: CoroutineScope = ClientScope(),
-    override val user: User,
+    override var user: User,
     internal val apiKey: ApiKey,
     internal var token: String,
     private val lifecycle: Lifecycle,
@@ -207,7 +209,8 @@ internal class StreamVideoClient internal constructor(
 
     internal var guestUserJob: Deferred<Unit>? = null
 
-    public override val userId = user.id
+    public override val userId: String
+        get() = user.id
 
     private val logger by taggedLogger("Call:StreamVideo")
     private var subscriptions = mutableSetOf<EventSubscription>()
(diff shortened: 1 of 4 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -547,6 +550,10 @@ internal class StreamVideoClient internal constructor(
             response.onSuccess {
                 coordinatorConnectionModule.updateAuthType("jwt")
                 coordinatorConnectionModule.updateToken(it.accessToken)
+                // Adopt the server-issued user identity so the SDK's local user.id
+                // can't drift from the JWT's user_id claim (parity with the JS SDK,
+                // which connects with response.user from createGuest).
+                this@StreamVideoClient.user = it.user.toUser().copy(type = UserType.Guest)
```
PratimMallick, line 556: There is a `state.user`  as well in `ClientState`. I think we also need to update the `_user` there  otherwise it will contain older user. wdyt?

rahul-lohra, line 556: Looks good for now. I think in future maybe we should migrate to `UserRepository` instead of keeping it `var`

rahul-lohra (approved):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/coordinator/CoordinatorSocketConnection.kt
```diff
@@ -233,10 +234,12 @@ public open class CoordinatorSocketConnection(
     override suspend fun sendEvent(event: VideoEvent): Boolean = internalSocket.sendEvent(event)
 
     override suspend fun connect(connectData: User) {
+        user = connectData
```
rahul-lohra, line 237: I think its a risky coding pattern to keep updating user at every place where user is used.
Better alternative is to use `UserRepository`

where the writes can happen at one place and all other call-sites can use `UserRepository.getUser()`

We can make it more better by creating something like
`ReadableUserRepository`
and 
`WritableUserRepository`

StreamVideoClient will have reference to `WritableUserRepository` and all other classes will have reference to `ReadableUserRepository`

rahul-lohra (commented):
Please review the comments

PratimMallick (approved):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/coordinator/CoordinatorSocketConnection.kt
aleksandar-apostolov: Yes, this is the correct approach, I think its not a big change do this in this PR.

aleksandar-apostolov: https://github.com/GetStream/stream-video-android/pull/1708

Issue #1655, opened 2026-04-22
Me user video freezes when camera turned off or when turned on it keeps showing avatar sometimes


ataballyyeva:
**Describe the bug**
Me user video freezes when camera turned off or when turned on it keeps showing avatar sometimes. It happens only when microphone is on, I could not reproduce it in microphone muted version.

**SDK version**
- 1.22.0

**To Reproduce**

1. Start a call and join
2. Make microphone is on 
3. Tap the Video icon to turn it off
4. tap it a few more times (give it a second to change state each time)

**Expected behavior**
Video stream should have 2 states only, button should always match. If camera on show video, if camera is off show avatar.

**Device:**
 - Pixel 7a
 - Android version: OS16

**Screenshots**

<img width="458" height="692" alt="Image" src="https://github.com/user-attachments/assets/fcf467bd-edbf-46e6-b5bf-5debbbd4ac19" />

ataballyyeva:
I closed old issue it is not duplicate it is different.

ataballyyeva:
I could reproduce it with demo app just adding extra Stream components it helped me to reproduce it. Just replace MainActivity of the DemoApp with following:

import android.os.Bundle
import android.widget.Toast
import androidx.activity.ComponentActivity
import androidx.activity.compose.setContent
import androidx.compose.foundation.background
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.foundation.layout.padding
import androidx.compose.foundation.layout.size
import androidx.compose.runtime.collectAsState
import androidx.compose.runtime.getValue
import androidx.compose.runtime.mutableStateOf
import androidx.compose.runtime.remember
import androidx.compose.runtime.setValue
import androidx.compose.ui.Alignment
import androidx.compose.ui.Modifier
import androidx.compose.ui.draw.clip
import androidx.compose.ui.graphics.Color
import androidx.compose.ui.unit.IntSize
import androidx.compose.ui.unit.dp
import io.getstream.video.android.compose.permission.LaunchCallPermissions
import io.getstream.video.android.compose.theme.VideoTheme
import io.getstream.video.android.compose.ui.components.call.activecall.CallContent
import io.getstream.video.android.compose.ui.components.call.controls.ControlActions
import io.getstream.video.android.compose.ui.components.call.controls.actions.FlipCameraAction
import io.getstream.video.android.compose.ui.components.call.controls.actions.ToggleCameraAction
[...]

PR #1706, opened 2026-06-03
Send only the changed track in UpdateMuteStates

PratimMallick:
### Goal

Refs [AND-1216](https://linear.app/stream/issue/AND-1216/bug-self-video-freezes-on-last-frame-on-multiple-toggling-of-camera)
Fixes a local self-view freeze / video-vs-avatar desync that occurs when toggling a single track during a call (reported in #1655).

## Summary
When toggling one track — e.g. muting the mic while the camera stays on, or turning the camera off while the mic stays on — `RtcSession.setMuteState` was sending the **full** declarative mute-state map for all track types. Re-asserting an *unchanged* track as un-muted (e.g. `VIDEO muted=false` while only the mic changed) makes the SFU re-emit a redundant `TrackPublishedEvent` for that track. That redundant event carries a `published_tracks` snapshot that can be stale (captured before the concurrent unpublish was applied), which re-enables the just-disabled track and leaves the renderer stuck on the last frame / showing the avatar while no frames arrive.

This change sends only the mute state of the track that changed.

### Implementation

- `RtcSession.setMuteState` now sends a single `TrackMuteState` for the changed `trackType` instead of the full map. The local `muteState` map and retry logic are unchanged. On SFU (re)connect/migration, each enabled track is re-signalled individually via `listenToMediaChanges`, so the full state is still restored.

## Repro (before fix)

1. Join a call; turn on mic + camera.
2. With the mic on, toggle the camera off- on few times while keeping the mic unmuted
3. Self-view freezes on the last frame instead of showing the avatar; a subsequent on-toggle may show the avatar instead of live video.

The symmetric case (mic mute while video on) produced the same redundant-event pattern.

### Testing

- [x] Demo app: join, mic + camera on, toggle camera off with mic on → self-view shows avatar fallback correctly and restores live video on re-enable.
- [x] Mic mute while video on → no self-view desync.
- [x] SFU reconnect/migration mid-call → mute states for all enabled tracks restored correctly.

## Notes

The SFU re-emitting a publish event with a stale snapshot is still a latent robustness gap; a separate SFU-side ticket is being filed. A defensive client guard (a single-track event never mutating other tracks' enabled state) was intentionally left out of this PR per discussion.

## Bug Fixes
* Fixed mute/unmute operations to prevent unexpected renderer desynchronization and state inconsistencies when toggling audio or video track mute states during calls.


rahul-lohra (approved):
lgtm

aleksandar-apostolov (approved):

PR #1707, opened 2026-06-04
Auto-connect and register push device for guest users

aleksandar-apostolov:
Stacked on #1704. Merge after parent.

### Goal
Refs [AND-1202](https://linear.app/stream/issue/AND-1202/guest-user-type-is-broken) — `StreamVideoBuilder` previously ran auto-register-push and auto-connect only for `UserType.Authenticated`. Guest integrators had to write the same boilerplate (manual `registerPushDevice` + `connect()` after `build`) — boilerplate iOS and JS SDKs don't require.

Third behavior gap in the guest-mode series, stacked above #1703 (race fix) and #1704 (response.user adoption). Both are functional prerequisites: `registerPushDevice()` and `connectAsync()` await `guestUserJob`, which only works correctly with #1703.

### Implementation
- `StreamVideoBuilder.build()`: widen the gate at the auto-register/auto-connect block to `UserType.Authenticated || UserType.Guest`. Anonymous still excluded (no identity to register against).
- Both inner branches (`autoRegisterPushDevice` and `connectOnInit`) remain integrator-opt-out — Guest users can still set either flag to `false` to keep manual control of either path.

### Testing
- No dedicated unit test added. The change is a one-line widening of an existing condition; the methods it now invokes for Guest (`StreamVideoClient.connectAsync()`, `StreamNotificationManager.registerPushDevice()`) already have regression coverage in `StreamVideoClientTest` (added by #1703) and the existing notification-manager tests. `StreamVideoBuilderTest` is `@Ignore`d at class level for full builds (run via `isolatedTest`), so a direct gate test would duplicate the existing isolated coverage without unique signal.
- End-to-end functional verification on device pending — smoke test on a Guest build with default config recommended before lifting from Draft.

## Summary by CodeRabbit

* **New Features**
  * Guest users now automatically receive push notifications and WebSocket connection support, matching the capabilities previously available only to authenticated users.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
@@ -314,8 +314,11 @@ public class StreamVideoBuilder @JvmOverloads constructor(
             coordinatorConnectionModule.updateAuthType("anonymous")
         }
 
-        // Establish a WS connection with the coordinator (we don't support this for anonymous users)
-        if (user.type == UserType.Authenticated) {
+        // Auto-register push and open the coordinator WS for any user with an
+        // identity. Anonymous users don't have one, so neither path applies.
+        // Guest setup runs asynchronously inside StreamVideoClient — both
+        // registerPushDevice() and connectAsync() already await guestUserJob.
+        if (user.type == UserType.Authenticated || user.type == UserType.Guest) {
             scope.launch {
                 try {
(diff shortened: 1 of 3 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
PratimMallick, line 326: Since this doesn't go through the same flow which is fixed by PR #1703 , do will need a similar fix here as well before making the createDevice() API call to cordinator?

PR #1712, opened 2026-06-09
Stable participant count during large livestreams

aleksandar-apostolov:
### Goal

Fixes [AND-926](https://linear.app/stream/issue/AND-926/fix-fluctuating-participantcounts-when-large-number-of-participants) — the participant count displayed in `LivestreamPlayer` fluctuates wildly during large livestreams (e.g. 25k → 32k → 42k → 28k in 1-2 seconds, sometimes diverging by tens of thousands).

Root cause: post-join, two sources race to write `_participantCounts` — the SFU healthcheck (authoritative) and any coordinator event carrying a `CallSessionResponse` (stale snapshot). The existing guard `connection.value !is RealtimeConnection.Joined` only blocks updates during the brief `Joined` instant; once the connection promotes to `Connected`, every coordinator `participant_joined` / `participant_left` / `counts_updated` event re-writes the count from the stale snapshot.

### Implementation

`CallState.updateParticipantCounts()`:
- Replace `!is RealtimeConnection.Joined` with a new `isInCall()` helper covering `Joined`, `Connected`, `Reconnecting`, and `Migrating` — the post-join lifetime where the SFU healthcheck is the source of truth.
- Pre-join, the session-derived formula now takes `max(byRoleCount, participants.size)` before adding `anonymousParticipantCount` — React's monotonicity guard for fast joins.

Event entry points unchanged: `CallSessionParticipantJoinedEvent`, `CallSessionParticipantLeftEvent`, `CallSessionParticipantCountsUpdatedEvent`, and `didUpdateSession` still mutate the local session snapshot — they just no-op for count display once we're in the call.

### Testing

Four new unit tests added in `CallStateTest`, each verified failing against the original code before this fix landed, then passing after. They cover the three post-join clobber paths (`participant_joined`, `participant_left`, `counts_updated`) and the pre-join `max(byRoleCount, participants.size)` parity. Full `CallStateTest` suite still passes 13/13.

## Release Notes

* **Bug Fixes**
  * Improved participant count accuracy during active video calls by refining how the system reconciles participant data across different connection states, ensuring more reliable real-time participant counts.

* **Tests**
  * Added comprehensive test coverage for participant count behavior during various call states and connection transitions.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
@@ -1441,16 +1441,29 @@ public class CallState(
         session: CallSessionResponse? = null,
         sfuHealthCheckEvent: SFUHealthCheckEvent? = null,
     ) {
-        // When in JOINED state, we should use the participant from SFU health check event, as it's more accurate.
-
+        // Once we're in the call, the SFU health check is the single source of truth for
+        // participant counts. Coordinator session snapshots (counts_updated, participant_joined/left,
+        // and any event carrying a CallSessionResponse) carry a smaller, stale view at scale and
+        // must NOT clobber the SFU value — that produces wild fluctuations in livestreams.
         if (sfuHealthCheckEvent != null) {
             _participantCounts.value = sfuHealthCheckEvent.participantCount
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):
Looks good to me! ✅

rahul-lohra (approved):
lgtm

PR #1714, opened 2026-06-09
Fix self cancelling coroutine code (2/4)

rahul-lohra:
### Goal

Closes #AND-1223 Fix self cancelling coroutine code

> Please review and approve the PRs, but do not merge them yet.
> 
> Each PR is stacked on top of the previous one, so every subsequent PR uses the branch from the PR below it as its base. As a result, they are not currently targeting develop.

### Implementation

Found an internal issue: reconnection flow cancels its own coroutine mid-flight

What happens: When the SFU disconnects, the old session detects it and launches a rejoin on its own coroutine scope. During rejoin, we clean up the old session — which cancels that same coroutine scope. The rejoin coroutine is now cancelled but still running (cooperative cancellation), so it proceeds to connect the new session's WebSocket. The moment it hits the first suspension point (waiting for the socket state), it throws CancellationException.

Why the demo-app still works: The socket state is also being observed independently by a separate coroutine (stateJob) which is on the new session's scope. That coroutine correctly picks up the Connected state and updates all the call state, so the call recovers anyway.

Root cause: The reconnect coroutine should not be launched from old session's scope (which is destroyed as part of reconnection).

Here is what is happening
```kotlin
reconnectRejoin()
    → called via: coroutineScope.launch { call.reconnect(...) }  // RtcSession.kt:716
    →  coroutineScope is the OLD session's scope (child of supervisorJob)

    → oldSession.prepareRejoin("rejoin")           // Call.kt:1163
        → cancelActiveWork()                        // RtcSession.kt:2013
        → cleanup()                                 // RtcSession.kt:2014
            → supervisorJob.cancel()                // RtcSession.kt:1188
                ↑ supervisorJob is the PARENT of coroutineScope
                ↑ cancelling it cancels coroutineScope
                ↑ cancelling coroutineScope cancels the reconnect coroutine
                  (the very one running reconnectRejoin right now)

    → newSession.connectInternal(...)               // Call.kt:1181
        → withTimeoutOrNull { ... }
            → logger.d { "withTimeoutOrNull" }      // prints ✓ (not a suspension point)
            → state().first { ... }                 // FIRST suspension point
                → Kotlin checks isActive → false
                → throws CancellationException("Job was cancelled")  ✗
```
### Steps to reproduce
Join a Call
Toggle internet off for 8-9 seconds
You can observe the behaviour in log-cat

Screenshot is shared with code references

image

### 🎨 UI Changes

None

### Testing

Join a Call
Toggle internet off for 8-9 seconds
Watch log cat - there should no exception

## Summary by CodeRabbit

* **Refactor**
  * Improved scope management for call reconnection and rejoin operations to ensure they are properly tied to the call lifecycle, enhancing reliability of network recovery and session restoration.


aleksandar-apostolov:
Same self-cancellation bug remains in the `WebSocketEventLost` branch a few lines below the one fixed at line 712:

https://github.com/GetStream/stream-video-android/blob/d9b64476106f2eab4c864209bd39c92983c7642e/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt#L717-L723

`call.reconnect(...)` triggers `prepareRejoin → cleanup → supervisorJob.cancel()`, which cancels this `coroutineScope` mid-flight — exactly the chain the PR description documents. Per the rule stated in the description ("The reconnect coroutine should not be launched from old session's scope"), this should also go on `call.scope`.

Suggested one-liner:

```diff
                 is SfuSocketState.Disconnected.WebSocketEventLost -> {
                     logger.w { "[stateJob] HealthMonitor detected event loss for $sfuName — triggering reconnect" }
-                    coroutineScope.launch {
+                    call.scope.launch {
                         call.reconnect(
                             WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_FAST,
                             "SFU:healthcheck-timeout",
                         )
                     }
                 }
```

CodeRabbit flagged the same site as an "outside diff range" finding.

rahul-lohra:
Applied your suggestion

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1335,7 +1335,7 @@ public class RtcSession internal constructor(
             rejoin = {
                 logger.d { "[createPublisher] rejoin attempt, connection state: ${call.state.connection.value}" }
                 if (call.state.connection.value !is RealtimeConnection.Reconnecting) {
-                    coroutineScope.launch {
+                    call.scope.launch {
```
PratimMallick, line 1338: Here are not reconnecting or rejoining, then why do we need this on the call scope?

PratimMallick: It can actually run on any scope, but just asking if there was any need to change the scope?

aleksandar-apostolov (approved):
LGTM

PR #1718, opened 2026-06-10
Prevent "MediaSource has been disposed" crash when leaving a call

PratimMallick:
### Goal

Closes [AN-1228](https://linear.app/stream/issue/AND-1228/quantumquake-crash-fix-fatal-exception-javalangillegalstateexception)
Fix the production crash:

```
Fatal Exception: java.lang.IllegalStateException: MediaSource has been disposed.
   at org.webrtc.MediaSource.checkMediaSourceExists(MediaSource.java:69)
   at org.webrtc.PeerConnectionFactory.createAudioTrack(PeerConnectionFactory.java:490)
   at io.getstream.video.android.core.call.connection.StreamPeerConnectionFactory.makeAudioTrack(StreamPeerConnectionFactory.kt:679)
   at io.getstream.video.android.core.MediaManagerImpl.getAudioTrack(MediaManager.kt:1559)
   at io.getstream.video.android.core.MicrophoneManager$disable$1.invoke(MediaManager.kt:647)
   ...
   at com.twilio.audioswitch.AudioSwitch.enumerateDevices(AudioSwitch.kt:343)
   at com.twilio.audioswitch.AudioSwitch.start(AudioSwitch.kt:196)
   at io.getstream.video.android.core.audio.AudioSwitchController.start(AudioSwitchController.kt:38)
```

### How the crash was happening

`MediaManagerImpl`'s audio/video sources and tracks are created lazily and disposed on a different thread, with no synchronization. The crash is a thread-safety race during call teardown:

1. **Source exists.** During join/publish, `Publisher.newTrackFromSource(TRACK_TYPE_AUDIO)` reads `mediaManager.audioSource`, which lazily creates `_audioSource`. It builds its own sending track, so `MediaManagerImpl._audioTrack` stays `null`.
2. **A mic mute op is deferred.** The first `microphone.enable()/disable()` (e.g. from `updateMediaManagerFromSettings` at join) triggers a cold `AudioSwitch` setup. Its work lambda is **deferred** into the `audioDeviceChangeListener`, which runs later on the single-threaded `stream-audio-thread` once `AudioSwitch.start() → enumerateDevices()` completes.
3. **Leave disposes the source.** `leave()` → `internalLeave` launches `cleanup()` on `clientImpl.scope` (IO). `cleanup()` → `disposeTracksAndSources()` disposes `_audioSource`.
4. **Deferred lambda resumes.** On `stream-audio-thread`, the deferred `disable` runs the `audioTrack` getter: `_audioTrack == null`, so it calls `makeAudioTrack(source = audioSource)`. If this lands in the window where `_audioSource` was **disposed but not yet nulled** (or otherwise interleaves with disposal), WebRTC's `checkMediaSourceExists` throws `MediaSource has been disposed`.

The two operations run on different threads (`stream-audio-thread` for the deferred getter vs. the client IO scope for disposal) with no lock, and the old `disposeTracksAndSources()` disposed the sources first and nulled the references afterward — exposing a disposed-but-non-null source.

### Implementation

All changes are in `MediaManagerImpl` (+ mic/camera mute call sites) in `MediaManager.kt`:

- **Single reentrant lock (`mediaLock`)** guards every lazy getter/setter for `audioSource`, `audioTrack`, `videoSource`, `videoTrack`, `screenShareVideoSource`, `screenShareTrack`, and `disposeTracksAndSources()`. Disposal now disposes **and** nulls references atomically, so a racing getter can never observe a disposed-but-non-null source — it either runs fully before disposal (valid source) or after (sees `null` and recreates a fresh one). The lock is reentrant so the track getters can still call their source getters.
- **Terminal `released` flag** set under the lock in `cleanup()`. After teardown, the mic/camera mute paths must not lazily rebuild a track (which would also resurrect the `peerConnectionFactory` via its lazy getter and leak native objects that nothing disposes post-cleanup).
- **`runOnAudioTrackIfAvailable` / `runOnVideoTrackIfAvailable`** helpers now back `microphone.enable()/disable()` and `camera.enable()/disable()`. They create/use the track under the lock during an active call, but **no-op (with an error log)** once `released`. This is what neutralizes the exact crash path: the deferred mic-`disable` after leave becomes a clean no-op instead of recreating a track from a disposed source.

`disposeTracksAndSources()` intentionally does **not** set `released`, because `Call.recreateFactoryAndAudioTracks()` legitimately disposes and then recreates tracks before joining.

### Testing

- Reproduced the crash deterministically before the fix using a manual repro hook (a 5s delay in the `audioTrack` getter between creating the source and calling `makeAudioTrack`, then pressing **Leave** in the demo app during the window) — confirmed the exact `MediaSource has been disposed` stack trace. The repro hook is not part of this PR.
- Verified the fix on-device with join → leave flows (including fast leave during call setup) — no crash.
- Confirmed no behavior change for normal media usage: getters remain lazy ("create once, reuse"), and `recreateFactoryAndAudioTracks()` (pre-join audio-bitrate-profile change) still disposes and recreates correctly.
- `spotlessCheck` and the pre-push build/binary-compatibility checks pass.

### Notes / follow-ups

- A separate follow-up PR is planned to **decouple `microphone.enable()/disable()` from `enforceSetup`** so toggling the capture track no longer waits on `AudioSwitch` device enumeration. That addresses the deferral coupling at its root; this PR fixes the concurrency/lifecycle safety that should remain regardless.
- The public `audioTrack`/`videoTrack` setters are also guarded by `mediaLock` for memory-visibility/ordering consistency with the lock-protected getters and disposal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved microphone and camera control reliability through enhanced thread-safe synchronization, preventing potential race conditions during enable/disable operations.
  * Enhanced resource cleanup during application shutdown to ensure proper disposal of media resources and prevent issues from concurrent access.


aleksandar-apostolov (approved):
LGTM

PR #1719, opened 2026-06-11
Revert audio switch local AAR (#1710) and reset version to allow re-release of 1.26.0

aleksandar-apostolov:
### Goal

Restore develop to a publishable state so we can re-release 1.26.0. The first attempt failed on Maven Central because of the local AAR in #1710.

Reverts AND-1218 for now — the audio switch fix returns as 1.26.1 once the fork is on Maven.

### Implementation

- Revert #1710
- Reset `gradle.properties` to `1.25.0` so the release workflow can bump back to `1.26.0`

### Testing

`./gradlew spotlessCheck` passes. Pure revert + version bump, no other changes.


PratimMallick (approved):
LGTM

PR #1721, opened 2026-06-16
Do not show error message while video is loading

### LINKED ISSUE #1656: CallLobby: Enabling camera briefly flashes "This track failed to be loaded in the session" error text
### **Describe the bug**

When toggling the camera on in `CallLobby`, the `DefaultMediaTrackFallbackContent` composable is briefly visible for a split second before the camera feed appears. This always shows the text "This track failed to be loaded in the session (sessionId)" — which looks like an error to the user — even though the camera initializes successfully almost instantly after.

### **SDK version**
`1.19.0`

### **To Reproduce**

1. Open a `CallLobby` screen with the camera initially disabled
2. Tap the camera toggle button to enable the camera
3. Observe a brief flash of "This track failed to be loaded in the session (...)" before the camera feed appears. Happens everytime

### **Expected behavior**
No error text should be shown during normal camera initialisation. The transition from camera-off to camera-on should show either a neutral state (dark background, user avatar, or nothing) before the video feed appears.

### **Device:**
All Android devices I've tested on (Pixel 8, Pixel 6 Pro, Samsung S8)

### **Screenshots**

Image

### **Potential Root Cause**
`VideoRenderer` unconditionally renders `fallbackContent` as the bottom-most layer of its `Box` on every composition (line 99 of `VideoRenderer.kt`):

```
// Show avatar always behind the video.
videoRendererConfig.fallbackContent.invoke(call)
```

The intent is that the actual video (rendered on top via `AndroidView`) covers this layer. However, when the camera is first enabled there is a brief window where `call.camera.mediaManager.videoTrack` is `null`— meaning no `AndroidView` is created — leaving `fallbackContent` as the only visible content.

The default `fallbackContent` (wired up in `VideoRendererConfigCreationScope`) is `DefaultMediaTrackFallbackContent`, which renders the misleading error string:

```
public var fallbackContent: @Composable (Call) -> Unit = {
    DefaultMediaTrackFallbackContent(modifier = Modifier, call = it)
}
```

`DefaultMediaTrackFallbackContent` is semantically named and worded as a failure state, but it is being used as a general-purpose backdrop that renders unconditionally — including during normal camera warm-up.

The comment on line 99 says "Show avatar always behind the video", suggesting the original intent was a neutral avatar placeholder, not an error message.

### **Suggested fix**

The `fallbackContent` default should be a visually neutral placeholder (e.g. a dark surface or user avatar) rather than an error string, since it is rendered unconditionally and will always be briefly visible when the camera is toggled on. The error string gives users the false impression that something has gone wrong.

Alternatively, the error string should only be shown when a genuine track failure has been detected, not as the always-on backdrop layer.

### Current Workaround
Override `CallLobby`'s `onRenderedContent` w/ a `VideoRenderer` w/ an overridden `fallbackContent`

```
VideoRenderer(
    modifier = Modifier.fillMaxSize(),
    call = lobbyCall,
    video = video,
    videoRendererConfig = videoRenderConfig { fallbackContent = { } }
)
```

PratimMallick:
### Goal
Closes [1235](https://linear.app/stream/issue/AND-1235/fix-show-loading-spinner-in-calllobby-while-camera-warms-instead-of)
Fix [#1656](https://github.com/GetStream/stream-video-android/issues/1656): when enabling the camera in CallLobby, users briefly saw "This track failed to be loaded in the session" even though the camera was initializing normally.

The default VideoRenderer fallback was an error message rendered unconditionally behind the video layer, so it flashed whenever the track was not yet visible.

### Implementation

-  Replaced the error Text in `DefaultMediaTrackFallbackContent` with a neutral `CircularProgressIndicator` on the standard tertiary background, and use it as the single default fallbackContent for `VideoRenderer`.

- Updated `CallLobby` to show a loading overlay until onRendered fires, so the spinner is visible during camera warm-up even when a videoTrack object already exists but the first frame has not rendered yet.

- Removed the `UserAvatarBackground` from the demo-app `CallLobbyScreen` so the lobby uses the SDK default behavior.

### 🎨 UI Changes

https://github.com/user-attachments/assets/ef457462-a563-4133-9b6c-12d336c942d6

### Testing

- [] ./gradlew :stream-video-android-ui-compose:testDebugUnitTest --tests "io.getstream.video.android.compose.CallLobbyTest" — passe

- [] Manual verification on device:
1. Open Call Lobby
2. Turn camera off → avatar is shown
3. Turn camera on → spinner appears briefly, then live video
4. Confirm the old error text no longer flashes

#### General

#### Code & documentation

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

### DIFF
--- a/demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyScreen.kt
+++ b/demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyScreen.kt
@@ -27,7 +27,6 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
@@ -70,13 +69,10 @@ import io.getstream.video.android.CallActivity
 import io.getstream.video.android.R
 import io.getstream.video.android.compose.theme.VideoTheme
 import io.getstream.video.android.compose.ui.components.avatar.UserAvatar
-import io.getstream.video.android.compose.ui.components.avatar.UserAvatarBackground
 import io.getstream.video.android.compose.ui.components.base.StreamButton
 import io.getstream.video.android.compose.ui.components.call.controls.ControlActions
 import io.getstream.video.android.compose.ui.components.call.lobby.CallLobby
 import io.getstream.video.android.compose.ui.components.call.lobby.buildDefaultLobbyControlActions
-import io.getstream.video.android.compose.ui.components.video.VideoRenderer
-import io.getstream.video.android.compose.ui.components.video.config.videoRenderConfig
 import io.getstream.video.android.core.Call
 import io.getstream.video.android.core.call.state.CallAction
 import io.getstream.video.android.core.call.state.ToggleCamera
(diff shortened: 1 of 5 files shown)


aleksandar-apostolov (approved):
LGTM

PR #1724, opened 2026-06-17
Improve video analytics by correctly sending ice state

rahul-lohra:
### Goal

Closes AND-1241
Correctly map raw ice state to `VideoAnalyticsIceState`

### Implementation

Introduce a new `VideoAnalyticsIceState` which will be mapped from Raw ICE state
```kotlin
internal enum class VideoAnalyticsIceState(val text: String) {
    CONNECTED("CONNECTED"), FAILED("FAILED"), NOT_CONNECTED("NOT_CONNECTED")
}
```
### 🎨 UI Changes

None

### Testing

Run 1-1 calls/group calls. Verify the local logs or kibana

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of connection state reporting during calls.
  * Enhanced fallback handling for missing connection state information.

* **Improvements**
  * Refined grace period logic for connection state transitions to increase analytics reliability.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/PeerConnectionAnalytics.kt
```diff
@@ -49,73 +63,105 @@ internal class PeerConnectionAnalytics(
...
-                                        subscriber.hashCode(),
-                                        role = PeerConnectionRole.SUBSCRIBE,
-                                        iceState = subscriber.iceState.value,
-                                        peerConnectionState = state,
-                                    )
+                    .mapLatest { pcState ->
+                        val iceState = if (pcState == PeerConnection.PeerConnectionState.CONNECTED) {
+                            // Give ICE a grace period to reach CONNECTED before reporting the
+                            // connected peer connection; otherwise report the current ICE state.
+                            withTimeoutOrNull(ICE_CONNECTED_GRACE_MILLIS) {
```
PratimMallick: If we have to send both PCState and ICEState to analytics, then why do we have to wait for ICE state to change before reporting the PC state? Can we not send them independently in which ever order they are reported by libwebrtc?

rahul-lohra: We wait because of this  - 

I observed that PC(PeerConnection) is reaching `PeerConnectionState.CONNECTED` state while the ICE state is still on `IceState.CHECKING` state which ideally should also be at least IceState.CONNECTED state.

> Can we not send them independently in which ever order they are reported by libwebrtc?

We cannot do it at the moment because the Analytics API does not support this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/PeerConnectionAnalytics.kt
```diff
@@ -162,3 +186,24 @@ internal class PeerConnectionAnalytics(
         observePeerConnections(session)
     }
 }
+internal fun PeerConnection.IceConnectionState?.toVideoAnalyticsIceState(): VideoAnalyticsIceState {
```
aleksandar-apostolov: Is the collapse of `NEW` / `CHECKING` / `DISCONNECTED` / `CLOSED` into one `NOT_CONNECTED` bucket intentional?

`DISCONNECTED` is the one I'd want to keep on the wire — it's the signal for "ICE was up, dropped mid-call," which is distinct from "never connected" (`NEW`/`CHECKING`). After this lands, queries like "% of calls with mid-call ICE drops" can't be answered for the new SDK. Old SDKs in the field still emit the raw values, so dashboards end up mixing formats.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/PeerConnectionAnalytics.kt
```diff
@@ -162,3 +186,24 @@ internal class PeerConnectionAnalytics(
...
+    }
+}
+
+private data class PeerConnectionSnapshot(
+    val peerConnectionHashCode: Int,
+    val peerConnectionState: PeerConnection.PeerConnectionState,
+    val iceState: VideoAnalyticsIceState,
+)
+
+internal enum class VideoAnalyticsIceState(val text: String) {
```
aleksandar-apostolov: Worth moving `VideoAnalyticsIceState` and `toVideoAnalyticsIceState` to their own file — `VideoAnalyticsIceState.kt` in the same package?

They're referenced from `CallAnalytics.kt`, `ClientEventFactory.kt`, `ClientEventReporter.kt`, and two test files. Defining a shared domain type at the bottom of a sibling observer file makes it harder to find. Zero import churn since the package is the same.

rahul-lohra: sure

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/PeerConnectionAnalytics.kt
PratimMallick: Backend has a check where it rejects any api calls having values other than `CONNECTED` , `NOT_CONNECTED` or `FAILED`

rahul-lohra: > Is the collapse of NEW / CHECKING / DISCONNECTED / CLOSED into one NOT_CONNECTED bucket intentional?

Yes

Actually we the **Join Analytics Backend API** only accepts these 3 values for ice-state
```kotlin
 CONNECTED, 
 FAILED, 
 NOT_CONNECTED
```
At the moment, it will throw an error if we pass something else. 

But I agree on why we need  to send `DISCONNECTED`. I will sync up with the backend on this

aleksandar-apostolov: Please check, I am not against it to collapse it, but we lose information this way.

rahul-lohra: We have decided to move forward with the current implementation for now. We will revisit this once Gulzar is back

rahul-lohra: cc @guldev25

PratimMallick (approved):

PR #1736, opened 2026-06-30
[WIP] Extend CallJoinInterceptor with will-join and did-join hooks

rahul-lohra:
### Goal

IOS PR: https://github.com/GetStream/stream-video-swift/pull/1176
Closes AND-1269

`CallJoinInterceptor` previously exposed a single gate, `callReadyToJoin()`, which only let integrators run (and optionally veto) work at the very last step of a join. That isn't enough for use cases that need to bracket the entire connection window — for example muting remote media while the peer connection is still negotiating and restoring it once the call is live. This PR adds two best-effort lifecycle hooks around the existing gate so integrators get a clear before/after pair.

### Implementation

**New public API**

- Add `CallJoinLifecycleInterceptor` (core/call/interceptor) extending  `CallJoinInterceptor` with `callWillJoin(call)` and `callDidJoin(call)`.

```kotlin
public interface CallJoinLifecycleInterceptor : CallJoinInterceptor {
    public suspend fun callWillJoin(call: Call)
    public suspend fun callDidJoin(call: Call)
}

```
  
- Deprecate `CallJoinInterceptor` (WARNING) pointing to the new interface.
  
**Lifecycle wiring**

- `Call.join()` invokes `callWillJoin(this)` once the join result is produced,  when the interceptor implements `CallJoinLifecycleInterceptor`.
- `ActiveStateGate.invokeInterceptor()` invokes `callDidJoin(call)` right after
  `onReady()` callback (which transitions to RingingState to Active)
  
  ``` kotlin
  call.join()
   │
   ├─▶ callWillJoin()      session ready, before Active
   │
   │   ⏳ wait: publisher CONNECTED
   │
   ├─▶ callReadyToJoin()   gate — suspend to delay · throw to abort · 5s max
   │
   ├─── RingingState.Active
   │
   └─▶ callDidJoin()       after Active


### Demo App
- DemoCallJoinInterceptor now implements the lifecycle interface: it silences incoming audio in `callWillJoin` and restores it in `callDidJoin`.

 ### Example Usage
 
 ```kotlin
 class DemoCallJoinInterceptor(
    private val previousRingingStates: Set<RingingState>,
) : CallJoinLifecycleInterceptor {

    override suspend fun callWillJoin(call: Call) {
        // Perform any setup before the join starts.
        // For example:
        // - Start observing participants.
        // - Temporarily mute incoming audio.
        // - Initialize resources needed during the join.
    }

    override suspend fun callReadyToJoin(call: Call) {        
        // - Exchange readiness signals between caller and callee.
        // - Wait until both sides are ready.
    }

    override suspend fun callDidJoin(call: Call) {
        // Clean up any temporary state.
        // For example:
        // - Stop observers.
        // - Re-enable incoming audio.
    }
}
```

```kotlin
call.join(callJoinInterceptor = DemoCallJoinInterceptor(previousRingingStates) ) 
```

Other
- Update stream-video-android-core.api with the new public surface.

### 🎨 UI Changes

None

### Testing

1. Go to Direct Call Screen
2. Enable Call Join First
3. Enable Use Call Join Interceptor
4. Choose a contact to make a call

## Summary by CodeRabbit

* **New Features**
  * Added call join lifecycle hooks for apps that need to react before and after a call fully joins.
  * Join handling now supports additional setup and cleanup steps around the join flow.

* **Bug Fixes**
  * Improved handling for participant audio during join, helping prevent audio from being active at the wrong time.

* **Chores**
  * The previous join interceptor is now deprecated in favor of the new lifecycle-based interceptor.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -624,6 +626,11 @@ public class Call(
                             "is joined. MediaManager will not be initialised with server settings."
                     }
                 }
+
+                if (callJoinInterceptor is CallJoinLifecycleInterceptor) {
+                    callJoinInterceptor.callWillJoin(this)
+                }
+
```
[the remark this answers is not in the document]
rahul-lohra: We don't need to guard `callWillJoin` unlike `callReadyToJoin`. It is there by design

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
```diff
@@ -83,7 +84,12 @@ internal class ActiveStateGate(
                 val shouldProceed = invokeInterceptor(call, interceptor)
                 if (!isActive) return@launch
 
-                if (shouldProceed) onReady()
+                if (shouldProceed) {
+                    onReady()
+                    if (interceptor is CallJoinLifecycleInterceptor) {
+                        interceptor.callDidJoin(call)
+                    }
+                }
```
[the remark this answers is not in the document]
rahul-lohra: Resolved

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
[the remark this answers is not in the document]
rahul-lohra: Its added now

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -624,6 +626,11 @@ public class Call(
                             "is joined. MediaManager will not be initialised with server settings."
                     }
                 }
+
+                if (callJoinInterceptor is CallJoinLifecycleInterceptor) {
+                    callJoinInterceptor.callWillJoin(this)
```
aleksandar-apostolov: Wrap this like `callDidJoin` in `ActiveStateGate` (rethrow `CancellationException`, log-and-proceed) — otherwise a throwing hook fails the whole join.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
```diff
@@ -83,7 +84,18 @@ internal class ActiveStateGate(
                 val shouldProceed = invokeInterceptor(call, interceptor)
                 if (!isActive) return@launch
 
-                if (shouldProceed) onReady()
+                if (shouldProceed) {
```
aleksandar-apostolov, line 86: Same `callDidJoin` try/catch at both `onReady()` sites — extract to a private helper.

demo-app/src/main/kotlin/io/getstream/video/android/DemoCallJoinInterceptor.kt
```diff
@@ -31,13 +39,31 @@ import kotlinx.coroutines.flow.first
...
     private val previousRingingStates: Set<RingingState>,
-) : CallJoinInterceptor {
+) : CallJoinLifecycleInterceptor {
     private val logger by taggedLogger("DemoCallJoinInterceptor")
 
     companion object {
         const val CALLER_READY_TO_JOIN_EVENT_TYPE = "caller_ready_join"
         const val CALLEE_READY_TO_JOIN_EVENT_TYPE = "callee_ready_join"
     }
+    var observeJob: Job? = null
```
aleksandar-apostolov: `observerScope` is never cancelled and `observeJob` keeps running if `callDidJoin` never fires (aborted/failed join) — remote audio stays muted. Tie the scope to a cancellable lifecycle and make `observeJob`/`observerScope` `private`. Integrators copy the demo.

aleksandar-apostolov (commented):
Nice feature. Two things before it hits the public API:

**1. Consider default methods instead of a new interface.** We build with `-Xjvm-default=all`, so we can add `callWillJoin`/`callDidJoin` directly to `CallJoinInterceptor` with empty default bodies — existing implementers keep compiling, `callReadyToJoin` stays required. That avoids introducing `CallJoinLifecycleInterceptor` and deprecating the base interface (which pushes a warning onto everyone implementing `callReadyToJoin`, even though it isn't going away). Fewer types, no deprecation churn.

**2. `callWillJoin` isn't best-effort.** In `Call.kt` it's unwrapped, so a throwing hook propagates out of `join()` and the successful join is never returned. `callDidJoin` is wrapped — `callWillJoin` should be too.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: Done

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ActiveStateGate.kt
rahul-lohra: Done

demo-app/src/main/kotlin/io/getstream/video/android/DemoCallJoinInterceptor.kt
rahul-lohra: Correct, its done

aleksandar-apostolov (approved):

rahul-lohra:
At the moment, we don’t have a clear direction for the `CallJoinInterceptor` work after observing the ~2-second delay in hearing first audio frame.
Our current suspicion is that the delay is caused by late delivery of the presence signal, which in turn delays the transition to the Active screen by about 2 seconds.

Putting this PR on draft mode for now. If there are no significant development then we will close this PR

rahul-lohra:
Closing this PR as there is no active development going on this direction at the moment
We will revisit it later when needed.

PR #1740, opened 2026-07-01
Report SFU WebSocket connection and join timeouts accurately

PratimMallick:
### Goal
Fixes [AN-1294](https://linear.app/stream/issue/AND-1294/fix-the-timeout-based-logic-of-sfu-ws-connection)

 This PR reworks SFU socket connect/timeout handling so every outcome is attributed precisely.

Issues this PR solves:

1. **Connect could hang on a silent SFU.** Once the transport WebSocket was upgraded, OkHttp no longer applied a timeout, so if the SFU never sent its `JoinCallResponse` we could wait indefinitely. There was no separate deadline for the post-open handshake.
2. **A successful-but-slow connect / socket timeout was wrongly surfaced as a hard join failure to the app.** There were effectively two competing timeouts (OkHttp's, plus a `withTimeoutOrNull` inside `connectInternal`) and the terminal state wasn't interpreted consistently.
3. **Opaque error messages.** Transport failures surfaced as `SOCKET_FAILURE`'s generic *"See stack trace in logs…"* instead of the real reason (e.g. `Failed to connect to /1.2.3.4:443`, `timeout`, SSL handshake errors), and the HTTP status code from a failed WS upgrade (401/429/5xx) was discarded.
6. **Join-error analytics were emitted during reconnects.** `connectInternal` reported join failures even when it ran as part of the internal reconnect loop, double-counting/false-positive join errors.
7. **Retry count was inaccurate.** There was no per-session notion of how many SFU WS retries an attempt actually took.
7. **`connectionTimeoutInMs` from `StreamVideoBuilder` was ignored** by the SFU socket path (a hardcoded 10s was used, and the post-open handshake had no configurable bound).

### Implementation

- **Two-phase connect with a new `WebSocketConnected` state.** The transport open (HTTP→WS upgrade) is bounded by OkHttp's connect timeout; a new `SfuSocketState.WebSocketConnected` marks "socket open, awaiting `JoinCallResponse`". `StreamWebSocket` now emits a `StreamWebSocketEvent.Open` on `onOpen` to drive this transition (coordinator socket treats it as informational).
- **Dedicated join-response timer.** While in `WebSocketConnected`, a timer (`joinResponseTimeoutMs`) bounds the wait for `JoinCallResponse`; on expiry it pushes a retryable `NetworkError` tagged `VideoErrorCode.SFU_JOIN_RESPONSE_TIMEOUT` → resolves to `DisconnectedTemporarily` instead of hanging.
- **Single recoverable path.** Removed the extra `withTimeoutOrNull` in `connectInternal`; all recoverable errors flow through one `DisconnectedTemporarily(error)` whose `NetworkError` carries the exact code + reason. `connectInternal` classifies the terminal state via `triggersReconnect()`/`networkErrorOrNull()`.
- **Real error surfaced.** `StreamWebSocket.onFailure` now builds `Error.NetworkError` directly with the throwable's message and the HTTP `response.code` when the upgrade failed on an HTTP response (falls back to unknown for pure transport failures).
- **Accurate timeout analytics.** `toAbortReason()` maps both the join-response timeout (by error code) and the transport connect timeout (by `SocketTimeoutException` cause — a type check, not string matching) to `REQUEST_TIMEOUT`; everything else stays `SFU_ERROR`.
- **Join-only analytics.** Join-error analytics moved out of `connectInternal` into `Call._join` via `sendJoinErrorAnalytics(...)`, so reconnect-driven connects don't report join errors. `SfuConnectionResult.Failure` now carries `recoverable` + `abortReason`; on a recoverable failure the join flow awaits the reconnect outcome (`didReconnectSucceed()`).
- **Per-session retry counter.** `RtcSession.sfuWsRetryCount` increments only on actual retries (a `connectInternal` with `reconnectDetails`), resets to 0 on `Connected`, and is passed to `onSfuWsCompleted`.
- **Configurable timeout.** `StreamVideoBuilder.connectionTimeoutInMs` now drives both the SFU OkHttp client and the join-response deadline.
- Renamed `SfuConnectionResult.Connected/Failed` → `Success/Failure`; renamed `awaitReconnectOutcome()` → `didReconnectSucceed()`.

### Testing

- [x] `./gradlew :stream-video-android-core:testDebugUnitTest --tests "io.getstream.video.android.core.socket.sfu.*" --tests "io.getstream.video.android.core.rtc.*"` — pass
- [x] `./gradlew :stream-video-android-core:compileDebugKotlin` — pass
- [x] `./gradlew :stream-video-android-core:spotlessApply` — clean
- Manual: reproduced an SFU connect timeout via Charles proxy and verified the connection resolves to a retryable disconnect and the reconnect loop takes over (no hang).

New/updated tests: `SfuSocketStateServiceTest` (new `WebSocketConnected`/join-response transitions), `JoinRecoverableFailureTest` (recoverable-vs-permanent join handling), `RtcSessionTest2` (`connectInternal` terminal-state classification), `FastReconnectIceRestartTest` (result rename).

#### General

#### Code & documentation
- [ ] Comparison screenshots added for visual changes (N/A — no UI changes)
- [x] Affected documentation updated (KDocs)
- [ ] Tutorial starter kit updated (N/A)
- [ ] Examples/guides starter kits updated (`stream-video-examples`) (N/A)

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit
- [x] Check the SDK Size Comparison table in the CI logs

## Summary by CodeRabbit

* **New Features**
  * Added clearer handling for WebSocket connection stages, including a new “WebSocket connected” state before the final join response arrives.
  * Introduced support for timeout-based recovery when the join response does not arrive in time.

* **Bug Fixes**
  * Improved reconnect behavior for recoverable connection failures.
  * Added better error reporting for SFU join timeouts and transport failures.

* **Documentation**
  * Updated connection timeout guidance and defaults for a more predictable setup experience.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -112,6 +112,7 @@ import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
@@ -770,22 +771,79 @@ public class Call(
         }
         session.value = localSession
 
+        if (session.value == null) {
+            return Failure(Error.GenericError("RtcSession was null during connection to sfu"))
+        }
+
         session.value?.let {
             state._connection.value = RealtimeConnection.Joined(it)
         }
 
-        when (val result = session.value?.connectInternal()) {
-            is SfuConnectionResult.Connected -> Unit
-            is SfuConnectionResult.Failed ->
-                return Failure(
-                    Error.GenericError(result.error.message ?: "RtcSession error occurred."),
-                )
-            null ->
-                return Failure(Error.GenericError("RtcSession was null during connect"))
+        // This is the SFU ws connection
+        val sfuConnectionResult = session.value!!.connectInternal()
+
+        when (sfuConnectionResult) {
+            is SfuConnectionResult.Success -> Unit
+            is SfuConnectionResult.Failure -> {
+                if (sfuConnectionResult.recoverable) {
+                    logger.w { "[_join] Recoverable SFU connection failure — awaiting reconnect outcome" }
+                    if (!didReconnectSucceed()) {
+                        logger.e { "[_join] Could not recover. Error : $sfuConnectionResult" }
+                        sendJoinErrorAnalytics(sfuConnectionResult)
+                        return Failure(
+                            Error.GenericError(
+                                sfuConnectionResult.error.message ?: "SFU connection failed",
+                            ),
+                        )
+                    }
+                } else {
+                    logger.e {
+                        "[_join] Got non recoverable error while connecting to SFU. Error : $sfuConnectionResult"
+                    }
+                    sendJoinErrorAnalytics(sfuConnectionResult)
+                    return Failure(
+                        Error.GenericError(
+                            sfuConnectionResult.error.message ?: "RtcSession error occurred.",
+                        ),
+                    )
+                }
+            }
         }
         client.state.setActiveCall(this)
         monitorSession(result.value)
         return Success(value = session.value!!)
+    }
+
+    /**
+     * Reports the SFU WebSocket join failure to analytics. Only called from the join
+     * flow ([_join]) so that reconnect-driven [RtcSession.connectInternal] failures are
+     * not counted as join errors. The retry count comes from the session's
+     * [RtcSession.sfuWsRetryCount]; the failure reason and abort code come straight from
+     * the [SfuConnectionResult.Failure] the connect attempt produced.
+     */
+    private fun sendJoinErrorAnalytics(failure: SfuConnectionResult.Failure) {
+        callAnalytics.sfuAnalytics.onSfuWsCompleted(
+            success = false,
+            retryCount = session.value?.sfuWsRetryCount?.get() ?: 0,
+            failureReason = failure.error.message,
+            failureCode = (failure.abortReason ?: AnalyticsCallAbortReason.SFU_ERROR).name,
+        )
+    }
+
+    /**
+     * Suspends until the reconnect loop triggered by a recoverable connection failure
+     * reaches a terminal state, returning `true` if the call recovered (became
+     * [RealtimeConnection.Connected]) and `false` otherwise
+     * ([RealtimeConnection.ReconnectingFailed] / [RealtimeConnection.Disconnected]).
+     */
+    private suspend fun didReconnectSucceed(): Boolean {
+        val terminal = state.connection.first {
+            it is RealtimeConnection.Connected ||
+                it is RealtimeConnection.ReconnectingFailed ||
+                it is RealtimeConnection.Disconnected
+        }
+        logger.d { "[_join] Reconnect after recoverable connection failure settled on $terminal" }
+        return terminal is RealtimeConnection.Connected
     }
 
     private fun Call.monitorSession(result: JoinCallResponse) {
@@ -1207,15 +1265,14 @@ public class Call(
             val result = newSession.connectInternal(
                 reconnectDetails,
                 currentOptions,
-                JoinAnalyticsModel(joinAnalyticsModel.retryAttempt),
             )
         ) {
-            is SfuConnectionResult.Connected -> {
+            is SfuConnectionResult.Success -> {
                 newSession.sfuTracer.trace("rejoin", reason)
                 monitorSession(joinResponse.value)
                 ReconnectOutcome.Success
             }
-            is SfuConnectionResult.Failed -> ReconnectOutcome.Failed(result.error)
+            is SfuConnectionResult.Failure -> ReconnectOutcome.Failed(result.error)
         }
     }
 
@@ -1291,14 +1348,13 @@ public class Call(
             val result = newSession.connectInternal(
                 reconnectDetails,
                 currentOptions,
-                JoinAnalyticsModel(joinAnalyticsModel.retryAttempt),
             )
             when (result) {
-                is SfuConnectionResult.Connected -> {
+                is SfuConnectionResult.Success -> {
                     monitorSession(joinResponse.value)
                     ReconnectOutcome.Success
                 }
-                is SfuConnectionResult.Failed -> ReconnectOutcome.Failed(result.error)
+                is SfuConnectionResult.Failure -> ReconnectOutcome.Failed(result.error)
             }
         } finally {
             oldSession.finalizeMigration()
@@ -2218,8 +2274,9 @@ public class Call(
     companion object {
         /** How many consecutive FAST reconnect failures are allowed before
          *  escalating to a full REJOIN. Kept small because each failed FAST
-         *  attempt can cost up to DEFAULT_SOCKET_TIMEOUT (10 s) waiting for
-         *  the WebSocket handshake to time out. */
+         *  attempt can cost up to the socket connection deadline before it gives
+         *  up: OkHttp's WebSocket-upgrade timeout, followed by the join-response
+         *  wait — both driven by StreamVideoBuilder.connectionTimeoutInMs (default 10s). */
         private const val MAX_FAST_RECONNECT_ATTEMPTS = 3
 
         /** Absolute upper bound on loop iterations across all strategies
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -2271,7 +2271,7 @@ public class Call(
          *  escalating to a full REJOIN. Kept small because each failed FAST
          *  attempt can cost up to the socket connection deadline before it gives
          *  up: OkHttp's WebSocket-upgrade timeout, followed by the join-response
-         *  wait — both driven by StreamVideoBuilder.connectionTimeoutInMs (default 10s). */
+         *  wait — both driven by StreamVideoBuilder.connectionTimeoutInMs. */
         private const val MAX_FAST_RECONNECT_ATTEMPTS = 3
 
         /** Absolute upper bound on loop iterations across all strategies
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -805,7 +805,14 @@ public class Call(
         }
         val connectedSession = session.value ?: return Failure(Error.GenericError("RtcSession was cleared during connection to sfu"))
         client.state.setActiveCall(this)
-        monitorSession(result.value)
+        // rejoin/migrate recovery swaps in a NEW session and already calls monitorSession()
+        // with the recovered join response. fastReconnect recovery — and the normal success
+        // path — keep the original session, which is not monitored anywhere else. Only
+        // (re)establish monitoring when the session is unchanged, using the response that
+        // still matches it, so we neither double-register nor monitor with a stale response.
+        if (connectedSession === localSession) {
+            monitorSession(result.value)
+        }
         return Success(value = connectedSession)
     }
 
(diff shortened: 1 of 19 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -770,24 +771,81 @@ public class Call(
...
-        when (val result = session.value?.connectInternal()) {
-            is SfuConnectionResult.Connected -> Unit
-            is SfuConnectionResult.Failed ->
-                return Failure(
-                    Error.GenericError(result.error.message ?: "RtcSession error occurred."),
-                )
-            null ->
-                return Failure(Error.GenericError("RtcSession was null during connect"))
+        // This is the SFU ws connection
+        val sfuConnectionResult = session.value!!.connectInternal()
```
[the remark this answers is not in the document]
PratimMallick: Good catch, fixed by adding a seperate state for session value in the when clause and returning a failure in that scenario

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -2218,8 +2274,9 @@ public class Call(
     companion object {
         /** How many consecutive FAST reconnect failures are allowed before
          *  escalating to a full REJOIN. Kept small because each failed FAST
-         *  attempt can cost up to DEFAULT_SOCKET_TIMEOUT (10 s) waiting for
-         *  the WebSocket handshake to time out. */
+         *  attempt can cost up to the socket connection deadline before it gives
+         *  up: OkHttp's WebSocket-upgrade timeout, followed by the join-response
+         *  wait — both driven by StreamVideoBuilder.connectionTimeoutInMs (default 10s). */
```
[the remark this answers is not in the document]
rahul-lohra: @PratimMallick 
You can update the KDoc to match [StreamVideoBuilder.connectionTimeoutInMs]. Also, shouldn’t we update the timeout description in the KDoc as well?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
```diff
@@ -84,7 +84,10 @@ import java.net.ConnectException
  * @property loggingLevel Represents and wraps SDK logging levels for Stream logger, HTTP interceptor and native WebRTC logging.
  * @property notificationConfig The configurations for handling push notification.
  * @property ringNotification Overwrite the default notification logic for incoming calls.
- * @property connectionTimeoutInMs Connection timeout in seconds.
+ * @property connectionTimeoutInMs Connection timeout in milliseconds. Applies both to the
+ * coordinator/SFU OkHttp clients (HTTP and WebSocket upgrade) and to the SFU join-response
+ * wait (time allowed for the SFU to deliver its JoinCallResponse after the socket opens).
+ * Defaults to 10s.
```
rahul-lohra: I think the text `Defaults to 10s` is incorrect, please re-check

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -188,6 +188,7 @@ internal class StreamVideoClient internal constructor(
     internal val appName: String? = null,
     internal val audioProcessing: ManagedAudioProcessingFactory? = null,
     internal val loggingLevel: LoggingLevel = LoggingLevel(),
+    internal val connectionTimeoutInMs: Long = 10_000,
```
rahul-lohra: Why `StreamVideoClient` and `StreamVideoBuilder` has different `connectionTimeoutInMs`

aleksandar-apostolov (commented):
Solid PR. Few things before merge:

1. `connectionTimeoutInMs` KDoc says "Defaults to 10s" but the default is now `5_000`. 5s is fine, just fix the doc. `StreamVideoClient` still defaults to `10_000`, so align those too, and add a changelog line since the coordinator/SFU connect budget effectively drops from 10s.

2. On the recovered path, once `didReconnectSucceed()` returns true, `_join` falls through to `setActiveCall` + `monitorSession(result.value)` + `Success`. But reconnect already created a new session and called `monitorSession` on it. Does this double-register, or hand back a response that doesn't match the recovered session?

3. `didReconnectSucceed()` waits on `state.connection.first { ... }` with no timeout. Reconnect is bounded so it should settle, but if it hits `Disconnected` transiently it returns false early, and if it never settles `_join` hangs. A timeout on the same budget would be safer.

Nit: the RtcSessionTest2 timeout test emits `DisconnectedTemporarily` directly, so the `WebSocketConnected -> NetworkError -> DisconnectedTemporarily` transition isn't actually tested. Worth a case in SfuSocketStateServiceTest.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
aleksandar-apostolov: KDoc says "Defaults to 10s" but the default below is `5_000`. Since 5s is intended, this should read 5s.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
aleksandar-apostolov: This internal default is still `10_000` while the builder now defaults to `5_000`. Align them so a directly-constructed client (e.g. in tests) doesn't diverge from real usage.

PratimMallick: yeah they should be same

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
PratimMallick: updated

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
PratimMallick: updated .. make the default to 5s at both places

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/state/SfuSocketStateEvent.kt
```diff
@@ -31,6 +31,12 @@ public sealed class SfuSocketStateEvent {
         val connectionType: WebsocketReconnectStrategy = WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_UNSPECIFIED,
     ) : SfuSocketStateEvent()
 
+    /**
+     * Event to notify that the transport WebSocket has opened (HTTP→WS upgrade
+     * completed), but before the SFU JoinCallResponse has been received.
+     */
+    object WebSocketConnected : SfuSocketStateEvent() { override fun toString() = "WebSocketConnected" }
```
rahul-lohra: I can see duplicate naming has been used to represent both Event & State 

- `SfuSocketState` has a `WebSocketConnected`
- `SfuSocketStateEvent` also has `WebSocketConnected` (maybe we can use `WebSocketConnectedEvent`)

Should we refactor the name to avoid future confusion between event emission and state updates?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -898,57 +930,87 @@ public class RtcSession internal constructor(
...
+    /**
+     * Whether [stateJob] reacts to this disconnect by launching a [Call.reconnect].
+     * Used by the initial-join flow to decide whether to await the recovery loop
+     * instead of failing permanently.
+     *
+     * MUST mirror the reconnect-triggering branches in [stateJob]; the exhaustive
+     * `when` (no `else`) makes a new [SfuSocketState.Disconnected] subtype a compile
+     * error here, forcing both sites to stay in sync.
+     */
+    private fun SfuSocketState.Disconnected.triggersReconnect(): Boolean = when (this) {
```
rahul-lohra: Don't you think the name of the method feels incorrect? 

Currently it appears that this is an action instead it is just retrieving a status

Maybe we can use `canTriggerReconnect` or something more appropriate

PratimMallick: yeah makes sense

PratimMallick:
From the comments of @aleksandar-apostolov 

> 1. `connectionTimeoutInMs` KDoc says "Defaults to 10s" but the default is now `5_000`. 5s is fine, just fix the doc. `StreamVideoClient` still defaults to `10_000`, so align those too, and add a changelog line since the coordinator/SFU connect budget effectively drops from 10s.
 Fixed
> 2. On the recovered path, once `didReconnectSucceed()` returns true, `_join` falls through to `setActiveCall` + `monitorSession(result.value)` + `Success`. But reconnect already created a new session and called `monitorSession` on it. Does this double-register, or hand back a response that doesn't match the recovered session?
Added a check for this. In case of fast reconnect we donot monitor seperately 
> 3. `didReconnectSucceed()` waits on `state.connection.first { ... }` with no timeout. Reconnect is bounded so it should settle, but if it hits `Disconnected` transiently it returns false early, and if it never settles `_join` hangs. A timeout on the same budget would be safer.
The reconnect loop always reaches a terminal state and hence we donot have to worry about this 'hung' scenario

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/socket/sfu/state/SfuSocketStateEvent.kt
PratimMallick: Renamed to WebSocketEstablished

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoBuilder.kt
PratimMallick: yes hcanged

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -898,57 +930,87 @@ public class RtcSession internal constructor(
...
+        logger.d {
+            "[connectInternal] SFU WS connect (retryCount=$retryCount, reconnect=${reconnectDetails?.strategy})"
+        }
         sfuConnectionModule.socketConnection.connect(request)
-        val terminalState = withTimeoutOrNull(SocketActions.DEFAULT_SOCKET_TIMEOUT) {
-            sfuConnectionModule.socketConnection.state().first {
-                it is SfuSocketState.Connected || it is SfuSocketState.Disconnected
-            }
+        val sfuSocketState = sfuConnectionModule.socketConnection.state().first {
+            it is SfuSocketState.Connected || it is SfuSocketState.Disconnected
```
rahul-lohra: Did we covered the failure case in which our retries will be exhausted? 

Look at this code where we do an early exit after 10 retries - 

```kotlin
if (loopIteration >= MAX_RECONNECT_ATTEMPTS) {
     logger.w { "[reconnect] Max reconnect attempts ($MAX_RECONNECT_ATTEMPTS) reached — giving up" }
     state._connection.value = RealtimeConnection.ReconnectingFailed
     break
}
```

PratimMallick: So the reconnect loop comes into action after the socket state has moved : 

socketState has the following states : 
`Connecting` -> `WebSocketConnected` -> `Connected` -> `Disconnected` 

`Disconnected` futher has 7 mini states of its own 

When the socket moves to any of the disconnected states the retry loop starts. Also parallely this code 

```
val sfuSocketState = sfuConnectionModule.socketConnection.state().first {
            it is SfuSocketState.Connected || it is SfuSocketState.Disconnected
            }
```


would collapse and go into the else part of the cide which returns a `SfuConnectionResult.Failure` to the caller of `connectInternal` which is `_join` in the joining scenario . The `_join` now awaits on the result of retry loop in the method `didReconnectSucceed` where we take care of the events which retry loop can emit which are  : 
          `ReconnectOutcome.Success` , `ReconnectOutcome.PreconditionNotMet` , `ReconnectOutcome.PeerConnectionStale` , `ReconnectOutcome.Failed`, 
          
So we have covered all the scenarios of retry loop

rahul-lohra: Nice 👍

aleksandar-apostolov (approved):

PR #1741, opened 2026-07-08
Recover initial join when connect safety-timeout leaves no reconnect

PratimMallick:
### Goal

Follow-up to #1740 and [AN-1294](https://linear.app/stream/issue/AND-1294/fix-the-timeout-based-logic-of-sfu-ws-connection) That PR added a safety-timeout to `RtcSession.connectInternal` so the SFU connect wait can no longer suspend forever. But there is still a gap on the **initial join** path: when the safety-timeout fires, the socket is left in a **non-terminal** state that `stateJob` does not react to, so no `Call.reconnect` is ever launched. `_join` then calls `didReconnectSucceed()` and waits for a recovery loop that will never start — blocking indefinitely.

This PR closes that gap and makes the abandoned connect attempt safe to leave behind.

### Implementation

- **`SfuConnectionResult.Failure.reconnectTriggered`** — new flag that tells the join flow whether a recovery loop has already been launched by `stateJob`:
  - `true` for terminal disconnect states `stateJob` reacts to (a `Call.reconnect` is already in flight — the caller must not start another).
  - `false` for the connect safety-timeout, which leaves the socket in a state `stateJob` ignores (nothing will drive recovery on its own).
- **Tear down the abandoned socket on safety-timeout** — the connect attempt is still in flight when the safety-timeout fires, so `connectInternal` now calls `socketConnection.disconnect()` before returning. This prevents a late `Connected`/`JoinResponse` from an abandoned attempt resurfacing through `stateJob` and resurrecting a dead session ("zombie socket").
- **`Call._join` self-triggers recovery** — on a recoverable failure with `reconnectTriggered == false`, `_join` now launches a `REJOIN` itself before awaiting the outcome (a full REJOIN, not FAST, because the initial join never completed — there is no negotiated media path to resume, so FAST would just burn attempts on `PARTICIPANT_NOT_FOUND`). When `reconnectTriggered == true` it defers to the existing loop as before.
- **Rename** `SfuSocketState.Disconnected.canTriggersReconnect()` → `triggersStateJobReconnect()` for clarity (it describes which states `stateJob` acts on).
- **demo-app** — align `connectionTimeoutInMs` with the SDK default (`5_000L`) instead of the hardcoded `12_000L`.

### Testing

- [x] `./gradlew :stream-video-android-core:testDebugUnitTest --tests "*.JoinRecoverableFailureTest" --tests "*.RtcSessionTest2"` — passing
- [x] `./gradlew :stream-video-android-core:spotlessApply :demo-app:spotlessApply` — clean

New/updated unit tests:
- `JoinRecoverableFailureTest`
  - `recoverable failure with reconnect already triggered awaits the existing loop` — asserts `_join` does **not** start its own `reconnect` when `reconnectTriggered == true`.
  - `recoverable failure with no reconnect triggered starts a REJOIN itself` — asserts `_join` launches a `REJOIN` when `reconnectTriggered == false`.
- `RtcSessionTest2` — safety-timeout test now verifies the abandoned socket is `disconnect()`-ed.

Manual: reproduced a silent connect hang (OkHttp `callTimeout`/`readTimeout` set to `0` locally + Charles breakpoint on the SFU WS upgrade). Confirmed the safety-timeout fires, the socket is torn down, `_join` self-triggers a REJOIN, and the flow ends in a bounded failure instead of hanging.

#### General

#### Code & documentation
- [x] Affected documentation updated (KDocs on `SfuConnectionResult.Failure`)

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

_No UI change — skipping._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video connection recovery when the initial websocket join fails, including clearer handling of whether a reconnect is already in progress.
  * Tightened timeout handling so stalled connections are cleanly disconnected and can’t later re-activate an abandoned session.
  * Reduced the initial connection timeout for faster failure detection.
* **Tests**
  * Updated recovery tests to cover both automatic reconnect and caller-initiated rejoin flows.
  * Added coverage for disconnect behavior when the safety timeout is reached.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
@@ -223,12 +223,16 @@ internal sealed class SfuConnectionResult {
     /**
      * The connection attempt failed.
      *
-     * @param recoverable `true` when the terminal socket state is one that
-     * RtcSession's `stateJob` reacts to by triggering `Call.reconnect` (e.g.
-     * a connection timeout or a temporary socket disconnect). In that case the
-     * initial-join flow can defer to that recovery loop and await its outcome
-     * instead of treating the failure as permanent. `false` for terminal
-     * failures with no recovery (auth errors, permanent disconnects, etc.).
+     * @param recoverable `true` when the failure can be recovered from, so the
+     * initial-join flow can await a recovery loop instead of treating it as
+     * permanent. `false` for terminal failures with no recovery (auth errors,
+     * permanent disconnects, etc.).
+     * @param reconnectTriggered `true` when the terminal socket state is one that
+     * RtcSession's `stateJob` reacts to by launching `Call.reconnect`, so a recovery
+     * loop is already being started and the caller must not start its own. `false`
+     * when nothing will drive recovery on its own (e.g. the connect safety-timeout,
+     * which disconnects the socket into a state `stateJob` ignores) — in that case
+     * the initial-join flow must trigger `Call.reconnect` itself.
      * @param abortReason the analytics abort reason derived from the disconnect
      * state's error code, or `null` when the terminal state carried no error.
      * The caller (the join flow) decides whether/when to report it, so that
@@ -237,6 +241,7 @@ internal sealed class SfuConnectionResult {
     data class Failure(
         val error: Exception,
         val recoverable: Boolean = false,
+        val reconnectTriggered: Boolean = false,
         val abortReason: AnalyticsCallAbortReason? = null,
     ) : SfuConnectionResult()
 }
@@ -949,14 +954,28 @@ public class RtcSession internal constructor(
             }
         }
         if (sfuSocketState == null) {
+            // Safety net: the socket never reached a terminal state (Connected or
+            // Disconnected) within the timeout. This happens when the socket layer's own
+            // timers fail to fire and it gets stuck in a non-terminal state — e.g. the
+            // HTTP→WS upgrade stalls (proxy/captive portal drops packets silently, half-open
+            // TCP after a network handoff) so OkHttp never opens or errors the socket, or the
+            // WS opens but the SFU never sends the JoinResponse and the join-response wait
+            // never completes. Without this branch, `.first { }` would suspend forever.
             val msg =
                 "SFU connection timed out waiting for socket state (${connectTimeoutMs}ms)"
             logger.w { "[connectInternal] $msg" }
             sfuTracer.trace("connect-failed", msg)
+            // The socket never reached a terminal state, so the connect attempt is still
+            // in flight. Tear it down before returning so a late Connected/JoinResponse
+            // can't resurface through stateJob and resurrect this abandoned session.
+            sfuConnectionModule.socketConnection.disconnect()
             sendCallStats()
             return SfuConnectionResult.Failure(
                 error = Exception(msg),
                 recoverable = true,
+                // We disconnected the socket into a state stateJob ignores, so no
+                // recovery loop will start on its own — the caller must trigger it.
+                reconnectTriggered = false,
                 abortReason = AnalyticsCallAbortReason.REQUEST_TIMEOUT,
             )
         }
@@ -974,9 +993,14 @@ public class RtcSession internal constructor(
             logger.w { "[connectInternal] $msg" }
             sfuTracer.trace("connect-failed", msg)
             sendCallStats()
+            // These are the exact states stateJob reacts to, so if this is recoverable
+            // a Call.reconnect is already being launched — the caller must not start another.
+            val triggersReconnect =
+                (sfuSocketState as? SfuSocketState.Disconnected)?.triggersStateJobReconnect() ?: false
             SfuConnectionResult.Failure(
                 error = Exception(msg),
-                recoverable = (sfuSocketState as? SfuSocketState.Disconnected)?.canTriggersReconnect() ?: false,
+                recoverable = triggersReconnect,
+                reconnectTriggered = triggersReconnect,
                 abortReason = networkError?.toAbortReason(),
             )
         }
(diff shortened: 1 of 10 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -991,7 +1015,7 @@ public class RtcSession internal constructor(
      * `when` (no `else`) makes a new [SfuSocketState.Disconnected] subtype a compile
      * error here, forcing both sites to stay in sync.
      */
-    private fun SfuSocketState.Disconnected.canTriggersReconnect(): Boolean = when (this) {
+    private fun SfuSocketState.Disconnected.triggersStateJobReconnect(): Boolean = when (this) {
```
rahul-lohra: Is this name refactor intentional? The function does not seem to trigger anythin.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -974,9 +993,14 @@ public class RtcSession internal constructor(
             logger.w { "[connectInternal] $msg" }
             sfuTracer.trace("connect-failed", msg)
             sendCallStats()
+            // These are the exact states stateJob reacts to, so if this is recoverable
+            // a Call.reconnect is already being launched — the caller must not start another.
+            val triggersReconnect =
```
rahul-lohra: I think this variable name which just holds a status is slightly misleading. Please re-check

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -237,6 +241,7 @@ internal sealed class SfuConnectionResult {
     data class Failure(
         val error: Exception,
         val recoverable: Boolean = false,
+        val reconnectTriggered: Boolean = false,
```
rahul-lohra: `reconnectTriggered` sounds like it represents an action that has already been executed after entering a failed state. Because of that, it feels a bit out of place to store it here.

Do you think reconnect-related state and actions would be better encapsulated elsewhere, or managed as part of a dedicated reconnect state instead? 
That would make the responsibility of this class a bit clearer.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Refactoring back to `canTriggerReconnect`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: the reconnection is actulally started by the stateJob that observes the socketState. But we need that info here so that we can tell the calling function about what happened about reconnection . I think a better name would be `willStateJobReconnect`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Making a seperate errorType which is created only in this scenario (okhttp has not timedout but we are cancelling it anyways and returning a SfuConnectionResult.Failure with a specific errro type.That way we are not introducing any new field in SfuConnectionResult.Failure

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -241,6 +243,28 @@ internal sealed class SfuConnectionResult {
...
+     * started on its own — the initial-join flow must start one itself.
+     */
+    class Timeout(message: String) : SfuConnectException(message)
+
+    /**
+     * The SFU socket reached a terminal [SfuSocketState.Disconnected] state. When
+     * the failure is recoverable, `stateJob` is already launching `Call.reconnect`,
+     * so the initial-join flow only needs to await the outcome.
+     */
+    class Disconnected(message: String) : SfuConnectException(message)
```
rahul-lohra: I think the new type names are a bit misleading. `Timeout` works because it describes an actual failure event. `Disconnected`, however, is already an SFU socket state, not really an exception or failure cause.

rahul-lohra: I see two possible ways to make the intent clearer:
Option A: model the failure cause
```kotlin
internal sealed interface SfuConnectFailureCause {
    data object SafetyTimeout : SfuConnectFailureCause
    data object TerminalSocketState : SfuConnectFailureCause
}
```
Then SfuConnectionResult.Failure can carry cause, while error remains the message/throwable carrier. This keeps RtcSession.connectInternal() responsible for reporting what happened, and lets Call._join() decide what recovery action to take.

Option B: model the recovery action
```kotlin
internal enum class SfuConnectRecovery {
    StartRejoin,
    AwaitExistingReconnect,
    None,
}
```
This matches the current call-site even more directly, because _join() only needs to know whether it should start REJOIN, await the reconnect loop already started by stateJob, or fail immediately. With this option, the existing recoverable boolean likely becomes unnecessary.

PratimMallick: Option B would make connectInternal do more than it should. that is it would also end of taking the decision that calling layer should take when any error happens

rahul-lohra:
All concerns resolved 👍

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
rahul-lohra: Decided to go with this RP: https://github.com/GetStream/stream-video-android/pull/1741

rahul-lohra (approved):
lgtm

PR #1747, opened 2026-07-17
Decompose Call into focused internal components

PratimMallick:
### Goal

closes [AND-1318](https://linear.app/stream/issue/AND-1318/decompose-call-class-into-smaller-components-so-that-it-becomes-a-thin)

`Call.kt` had grown to ~2,350 lines and mixed many unrelated responsibilities (REST API calls, media, session/RTC lifecycle, reconnect state machine, connectivity/ICE monitoring, events, stats, rendering, teardown). This made it hard to read, test, and evolve.

This PR decomposes `Call` into focused `internal` collaborators while keeping `Call` as a thin, **binary-compatible** public facade — no public API changes. `Call.kt` goes from **2,351 to 1,043 lines**.

#### Structure

```mermaid
flowchart LR
    Call["<b>Call</b><br/>public facade<br/>(binary-compatible)"]

    Orch["<b>Orchestrators</b><br/>drive a flow across siblings<br/><br/>CallJoinCoordinator<br/>CallReconnector<br/>CallLifecycleManager"]

    Svc["<b>Services</b><br/>each owns one concern<br/><br/>CallApiClient · CallMediaManager<br/>CallEventManager · CallStatsReporter<br/>CallRenderer · SessionMonitor<br/>CallIceConnectionMonitor<br/>CallConnectivityMonitor"]

    Sess["<b>CallSessionManager</b><br/>single writer for<br/>the RTC session"]

    Seams["<b>Seams</b><br/>the only route back to<br/>the Call instance<br/><br/>ClientCallRegistry<br/>RtcSessionFactory<br/>MediaManagerFactory"]

    Call -->|constructs, delegates| Orch
    Call -->|constructs, delegates| Svc
    Orch <-->|direct sibling calls| Svc
    Orch -->|read / write| Sess
    Svc -->|read| Sess
    Orch -.->|needs the Call instance| Seams
    Svc -.-> Seams
    Seams -.->|implemented by| Call
```

The rules the diagram encodes:

- **Ownership points one way.** `Call` constructs every component and delegates its public API down to them. No component holds a `Call` reference, and none of them can reach back up except through a seam.
- **Siblings call each other directly**, rather than bouncing through the facade. Where a component is constructed before a collaborator exists (`CallReconnector` and `CallLifecycleManager` are built before `CallState`, because the event pipeline depends on them), the dependency is injected as a lazy provider.
- **One writer for the session.** `CallSessionManager` exposes the RTC session as a read-only `StateFlow` with `setActiveSession()` as the sole write path, so join, reconnect and teardown can read it but only one place mutates it.
- **Three seams** cover the handful of operations that genuinely need the call object itself. `ClientCallRegistry` registers and deregisters the call in the client's ringing/active slots, telecom and per-call cleanup; `RtcSessionFactory` and `MediaManagerFactory` keep the `Call` argument of `RtcSession` and `MediaManagerImpl` at the facade boundary. `Call` implements them as anonymous objects capturing `this`, so the call identity never leaks into a component.
- **No dependency cycles.** `CallReconnector` reads connectivity straight from the connection module rather than through `CallConnectivityMonitor`, which would otherwise close a `ConnectivityMonitor → Reconnector → ConnectivityMonitor` cycle.

#### Components

| Component | Owns |
|---|---|
| `CallApiClient` | Coordinator REST wrappers |
| `CallEventManager` | `events` flow, subscriptions, dispatch |
| `CallStatsReporter` | Stats collection/reporting, `statsReport` / `statLatencyHistory` |
| `CallRenderer` | Renderer binding, visibility, screenshots, incoming media overrides |
| `CallMediaManager` | Peer-connection factory, media manager, audio pipeline, screen sharing |
| `CallSessionManager` | The RTC session, session ids, location, reconnect bookkeeping |
| `CallIceConnectionMonitor` | Publisher/subscriber ICE-restart monitoring |
| `CallConnectivityMonitor` | Network listener + leave-on-disconnect timeout |
| `CallJoinCoordinator` | Join flow, retry loop, session creation/connection |
| `CallReconnector` | FAST/REJOIN/MIGRATE state machine, failed-SFU tracking |
| `CallLifecycleManager` | Leave/end/cleanup, leave guard, destroyed flag |
| `SessionMonitor` | SFU signal/event observers and session monitoring |

`Call` keeps its existing member names as delegating accessors, so `RtcSession`, `CallState` and existing callers keep working. Facade one-liners that existed only to route a call from one component to another (`stopConnectionMonitors`, `stopStatsReporting`, `cancelSfuObservers`, `shutDownJobsGracefully`, `monitorSession`, `getFailedSfuIdsSnapshot`) are removed now that siblings talk directly.

#### Two production bugs found while repairing the test suite

- `CallMediaManager` evaluated `eglBase().eglBaseContext` to build an argument for `MediaManagerFactory.create`, forcing a real EGL context before the factory (and its test hook) ran. `Call` owns both the context and the factory, so the parameter is dropped and the factory resolves it itself.
- The reconnect loop reads connectivity from the connection module, but the tests' network injection had been repointed at `CallConnectivityMonitor`. The loop therefore polled the real provider and stalled on `isConnected()` without consuming an attempt. Injecting at the module restores three reconnect tests that had been failing silently since the first decomposition commit.

#### Merge with `develop`

Reconciled `develop`'s `Call` changes into the new components: typed SFU connect-failure recovery (`SfuConnectionResult` + `SfuConnectFailureCause`) now lives in `CallJoinCoordinator`, and the `unitTestRtcSessionFactory` hook replaces `TestInstanceProvider.rtcSessionCreator`.

### Testing

- [x] `./gradlew :stream-video-android-core:testDebugUnitTest` — pass (**978 tests, 0 failures**, 50 skipped)
- [x] `./gradlew :stream-video-android-core:apiCheck` — pass (public API unchanged / binary-compatible)
- [x] `./gradlew :stream-video-android-core:spotlessCheck` — pass
- [x] `./gradlew assembleDebug` — pass (all modules incl. demo-app)

New unit tests cover each extracted component: `CallApiClientTest`, `CallEventManagerTest`, `CallStatsReporterTest`, `CallRendererTest`, `CallMediaManagerTest`, `CallSessionManagerTest`, `CallIceConnectionMonitorTest`, `CallConnectivityMonitorTest`, `CallJoinCoordinatorTest`, `CallReconnectorTest`.

Two existing tests were rebuilt rather than patched, because they depended on internals that no longer exist:

- `JoinRecoverableFailureTest` spied `Call` and reflectively repointed a field inside `CallJoinCoordinator`. It now constructs the coordinator directly with mocked collaborators and drives `joinInternal`, covering the same three SFU connect-failure causes.
- `CallReconnectorTest` now builds the reconnector with its injected siblings and a real `CallSessionManager`, so the session and reconnect bookkeeping behave as they do in production instead of being stubbed.

Test-only reflection seams (`injectSession`, `injectMockNetwork`) are consolidated into a single `CallTestSeams.kt` rather than duplicated across test classes, so no test-only API was added to `Call`.

#### General

#### Code & documentation
- [x] Changelog is updated with client-facing changes (N/A — no client-facing/API changes)
- [ ] Comparison screenshots added for visual changes (N/A — no UI change)
- [x] Affected documentation updated (KDocs on every new component and seam)

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

Skipped — internal refactor with no user-facing change.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -594,111 +410,29 @@ public class Call(
...
-            AnalyticsCallAbortReason.RETRY_EXHAUSTED.name,
-            errorMessage,
+        return joinCoordinator.join(
+            create,
+            createOptions,
+            ring,
+            notify,
+            hintHighScaleLivestreamPublisher,
+            callJoinInterceptor,
         )
```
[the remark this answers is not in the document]
PratimMallick: Yes, fixed it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallConnectivityMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+ * Observes device network connectivity for a [Call] and drives the call's response:
+ * triggering a fast/rejoin reconnect when connectivity returns, and leaving the call if
+ * the device stays offline past the configured `leaveAfterDisconnectSeconds`.
+ *
+ * Also owns the call's subscription to the underlying [NetworkStateProvider].
+ */
+internal class CallConnectivityMonitor(
+    private val call: Call,
+) {
+    private val logger by taggedLogger("Call:ConnectivityMonitor:${call.type}:${call.id}")
```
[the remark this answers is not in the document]
PratimMallick: Was present earlier too, so will fix in a new PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallEventManager.kt
```diff
@@ -0,0 +1,98 @@
...
+                sub.filter?.let {
+                    if (it.invoke(event)) {
+                        sub.listener.onEvent(event)
+                    }
+                }
+            }
+        }
+
+        if (!events.tryEmit(event)) {
+            logger.e { "Failed to emit event to observers: [event: $event]" }
```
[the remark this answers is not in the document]
PratimMallick: Was present earlier too, so will fix in a new PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+                        val wasMigrating = currentStrategy ==
+                            WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE
+                        val pastFastReconnectDeadline = (System.currentTimeMillis() - loopStartTime) >
+                            call.reconnectDeadlineMillis
+                        val shouldEscalateToRejoin = wasMigrating ||
+                            pastFastReconnectDeadline
+
+                        if (shouldEscalateToRejoin) {
+                            currentStrategy = WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN
+                        }
```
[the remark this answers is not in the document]
PratimMallick: Was present earlier too, so will fix in a new PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallStatsReporter.kt
```diff
@@ -0,0 +1,91 @@
...
+            // Wait a bit before we start capturing stats
+            delay(reportingIntervalMs)
+
+            while (isActive) {
+                delay(reportingIntervalMs)
+                call.session.value?.sendCallStats(
+                    report = collectStats(),
+                )
+            }
+        }
```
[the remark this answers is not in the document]
PratimMallick: Was present earlier too, so will fix in a new PR

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/CallFieldDeclarationOrderTest.kt
```diff
@@ -49,15 +50,20 @@ class CallFieldDeclarationOrderTest {
...
         }
         val stateLine = lines.indexOfFirst {
             it.contains("val state = CallState(")
         }
 
+        assertThat(eventManagerLine).isGreaterThan(-1)
         assertThat(eventsLine).isGreaterThan(-1)
         assertThat(stateLine).isGreaterThan(-1)
+        assertThat(eventManagerLine).isLessThan(stateLine)
         assertThat(eventsLine).isLessThan(stateLine)
```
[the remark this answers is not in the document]
PratimMallick: Was present earlier too, so will fix in a new PR

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+                // Wait for network before doing anything else. Polls without
+                // consuming the attempt budget — the elapsed-time guard below
+                // will still fire if we wait too long.
+                if (!call.isNetworkConnected()) {
+                    logger.d {
+                        "[reconnect] Network unavailable — waiting for connectivity (loopIteration=$loopIteration)"
+                    }
+                    delay(RECONNECT_DELAY_MS)
+                    continue
+                }
```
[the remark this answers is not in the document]
PratimMallick: These were present earlier too, will make a newer PR for this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallConnectivityMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+                logger.d {
+                    "[NetworkStateListener#onDisconnected] #network; Leaving after being disconnected for ${clientImpl.leaveAfterDisconnectSeconds} (connection=$conn)"
+                }
+                call.leave(
+                    CallLeaveReason.Backend(
+                        cause = BackendCause.LEAVE_TIMEOUT_AFTER_DISCONNECT,
+                        message = message,
+                    ),
+                )
+            }
```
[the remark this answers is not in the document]
PratimMallick: It is not in the scope of this PR . this was present earlier too

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallRenderer.kt
```diff
@@ -0,0 +1,209 @@
...
+                // same thread as the videoframe is delivered will lead to a deadlock
+                // (needs investigation why)
+                call.scope.launch {
+                    track.video.removeSink(screenshotSink)
+                }
+                continuation.resume(bitmap)
+            }
+
+            track.video.addSink(screenshotSink)
+        }
```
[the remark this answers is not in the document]
PratimMallick: not in the scope of this PR, will make a seperate OR for this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallApiClient.kt
```diff
@@ -0,0 +1,402 @@
...
+ *
+ * This component holds no mutable call state — it is a stateless façade over the
+ * coordinator endpoints, extracted from [Call] to keep the public class focused.
+ */
+internal class CallApiClient(
+    private val call: Call,
+) {
+    private val logger by taggedLogger("Call:ApiClient:${call.type}:${call.id}")
+
+    private val clientImpl get() = call.clientImpl
```
rahul-lohra: Could have passed this as direct argument

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallApiClient.kt
```diff
@@ -0,0 +1,402 @@
...
+    private val type get() = call.type
+    private val id get() = call.id
+    private val state get() = call.state
```
rahul-lohra: Same here, we could have passed state in the constructor argument itself

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallApiClient.kt
```diff
@@ -0,0 +1,402 @@
...
+        logger.d { "[reject] #ringing; rejectReason: $reason, call_id:$id" }
+        return clientImpl.reject(type, id, reason)
+    }
+
+    fun collectUserFeedback(
+        rating: Int,
+        reason: String? = null,
+        custom: Map<String, Any>? = null,
+    ) {
+        call.scope.launch {
```
rahul-lohra: I am aware this is just code movement, however I think this logic should not run on call scope but instead on client scope or user scope

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallConnectivityMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+            val elapsedTimeMils = System.currentTimeMillis() - lastDisconnect
+            logger.d {
+                "[NetworkStateListener#onConnected] #network; no args, elapsedTimeMils:$elapsedTimeMils, lastDisconnect:$lastDisconnect, reconnectDeadlineMils:${call.reconnectDeadlineMillis}"
+            }
+            val strategy = if (lastDisconnect > 0 && elapsedTimeMils < call.reconnectDeadlineMillis) {
+                WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_FAST
+            } else {
+                WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN
+            }
+            call.reconnect(strategy, "NetworkStateListener#onConnected")
```
rahul-lohra: can we directly call `reconnector.reconnect(strategy, reason)` so that our code will follow UDF

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallConnectivityMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+                    logger.d {
+                        "[NetworkStateListener#onDisconnected] #network; Already reconnected ($conn) — not leaving"
+                    }
+                    return@launch
+                }
+                val message = "Leaving after being disconnected for ${clientImpl.leaveAfterDisconnectSeconds}"
+                logger.d {
+                    "[NetworkStateListener#onDisconnected] #network; Leaving after being disconnected for ${clientImpl.leaveAfterDisconnectSeconds} (connection=$conn)"
+                }
+                call.leave(
```
rahul-lohra: How about we directly call `lifecycle.leave(reason)` here, to follow UDF

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallConnectivityMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Observes device network connectivity for a [Call] and drives the call's response:
+ * triggering a fast/rejoin reconnect when connectivity returns, and leaving the call if
+ * the device stays offline past the configured `leaveAfterDisconnectSeconds`.
+ *
+ * Also owns the call's subscription to the underlying [NetworkStateProvider].
+ */
+internal class CallConnectivityMonitor(
+    private val call: Call,
```
rahul-lohra: We can pass CallState and ClientImpl directly here

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallEventManager.kt
```diff
@@ -0,0 +1,98 @@
...
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.launch
+import java.util.Collections
+
+/**
+ * Owns the event pipeline for a [Call]: the shared [events] flow, the set of legacy
+ * [EventSubscription]s, and the dispatch / handling of incoming [VideoEvent]s.
+ */
+internal class CallEventManager(
+    private val call: Call,
```
rahul-lohra: Low pointer: The encapsulated class should not be aware of the class that contains it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+
+    suspend fun joinInternal(
+        create: Boolean = false,
+        createOptions: CreateCallOptions? = null,
+        ring: Boolean = false,
+        notify: Boolean = false,
+        hintHighScaleLivestreamPublisher: Boolean? = null,
+        joinAnalyticsModel: JoinAnalyticsModel,
+    ): Result<RtcSession> {
+        call.nonFastReconnectAttempts = 0
```
rahul-lohra: Why not directly use `sessionManager.nonFastReconnectAttempts`. It will do 2 things
1. UDF
2. We don't need to unnecessary reference `call.nonFastReconnectAttempts`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.launch
+import stream.video.sfu.models.WebsocketReconnectStrategy
+
+/**
+ * Drives the join flow for a [Call]: permission checks, the bounded retry loop, the
+ * underlying join request to the coordinator, and creation + connection of the [RtcSession].
+ */
+internal class CallJoinCoordinator(
```
rahul-lohra, line 51: This class contains too many method invokation which backlinks to Call class, this breaks UDF at a lot of places. 

Can we fix this?
I will pinpoint to exact places and their solution

rahul-lohra: Just find all text with `call.`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+                    "To start this service and send the appropriate audio/video tracks the permissions are required,\n" +
+                    "otherwise the service will fail to start, resulting in a crash.\n" +
+                    "You can re-define your permissions and their expected state by overriding the [permissionCheck] in [StreamVideoBuilder]\n"
+            }
+        }
+        // if we are a guest user, make sure we wait for the token before running the join flow
+        clientImpl.guestUserJob?.await()
+
+        // Ensure factory is created with the current audioBitrateProfile before joining
+        call.ensureFactoryMatchesAudioProfile()
```
rahul-lohra: We can use - `media.ensureFactoryMatchesAudioProfile()`

And we can pass media on the constructor arg

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+
+        // the join flow should retry up to 3 times
+        // if the error is not permanent
+        // and fail immediately on permanent errors
+        state._connection.value = RealtimeConnection.InProgress
+        var retryCount = 0
+
+        var result: Result<RtcSession>
+
+        call.resetLeaveGuard()
```
rahul-lohra: Why not just directly invoke `lifecycle.resetLeaveGuard()`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+                hintHighScaleLivestreamPublisher,
+                JoinAnalyticsModel(retryCount, JoinReason.FirstAttempt),
+            )
+            if (result is Success) {
+                // we initialise the camera, mic and other according to local + backend settings
+                // only when the call is joined to make sure we don't switch and override
+                // the settings during a call.
+                val settings = state.settings.value
+                if (settings != null) {
+                    call.updateMediaManagerFromSettings(settings)
```
rahul-lohra: Just directly invoke - `media.updateMediaManagerFromSettings(callSettings)`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+    ): Result<RtcSession> {
+        logger.d { "[joinAndRing] #ringing; #track; members: $members, video: $video" }
+        state.toggleJoinAndRingProgress(true)
+        return join(
+            ring = false,
+            createOptions = createOptions,
+            callJoinInterceptor = callJoinInterceptor,
+        ).flatMap { rtcSession ->
+            logger.d { "[joinAndRing] Joined #ringing; #track; ring: $members" }
+            call.ring(RingCallRequest(call.isVideoEnabled(), members)).map {
```
rahul-lohra: We can use `apiClient.ring(ringCallRequest)`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+                logger.d { "[joinAndRing] Ringed #ringing; #track; ring: $members" }
+                clientImpl.state._ringingCall.value = call
+                rtcSession
+            }.onError {
+                logger.e { "[joinAndRing] Ring failed #ringing; #track; error: $it" }
+                state.toggleJoinAndRingProgress(false)
+                call.leave(
```
rahul-lohra: We can use  `lifecycle.leave(reason)`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+        call.cancelSfuObservers()
+
+        if (session.value != null) {
+            return Failure(Error.GenericError("Call ${call.cid} has already been joined"))
+        }
+        logger.d {
+            "[joinInternal] #track; create: $create, ring: $ring, notify: $notify, createOptions: $createOptions"
+        }
+
+        call.connectStartTime = System.currentTimeMillis()
```
rahul-lohra: We should avoid updating call related properties from outside of call class. 

Directly use `sessionManager.connectStartTime`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+
+        // step 1. call the join endpoint to get a list of SFUs
+        val locationResult = clientImpl.getCachedLocation()
+        if (locationResult !is Success) {
+            return locationResult as Failure
+        }
+        call.location = locationResult.value
```
rahul-lohra: Directly update - sessionManager.location. Avoid updating/communicating to containing class

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+
+        val options = createOptions
+            ?: if (create) {
+                CreateCallOptions()
+            } else {
+                null
+            }
+        val result =
+            call.joinRequest(
```
rahul-lohra: This method is already available in this class

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -0,0 +1,411 @@
...
+        location: String,
+        migratingFrom: String? = null,
+        migratingFromList: List<String>? = null,
+        ring: Boolean = false,
+        notify: Boolean = false,
+        hintHighScaleLivestreamPublisher: Boolean? = null,
+        joinAnalyticsModel: JoinAnalyticsModel,
+    ): Result<JoinCallResponse> {
+        val migratingFromList =
+            migratingFromList ?: call.getFailedSfuIdsSnapshot().takeIf { it.isNotEmpty() }
```
rahul-lohra: We can directly use `reconnector.getFailedSfuIdsSnapshot()`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallLifecycleManager.kt
```diff
@@ -0,0 +1,146 @@
...
+        isDestroyed = true
+
+        sfuSocketReconnectionTime = null
+
+        /**
+         * TODO Rahul, need to check which call has owned the media at the moment(probably use active call)
+         */
+        call.stopScreenSharing()
+        call.camera.disable()
+        call.microphone.disable()
```
rahul-lohra: Better to use media classes

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallLifecycleManager.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        if (call.id == call.client.state.activeCall.value?.id) {
+            call.client.state.removeActiveCall(call) // Will also stop CallService
+        }
+
+        if (call.id == call.client.state.ringingCall.value?.id) {
```
rahul-lohra: We already have client and state variable. Lets just use that

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallLifecycleManager.kt
```diff
@@ -0,0 +1,146 @@
...
+        return result
+    }
+
+    fun cleanup() {
+        // monitor.stop()
+        state.cleanup()
+        session.value?.cleanup()
+        call.shutDownJobsGracefully()
+        call.stopStatsReporting()
+        call.mediaManager.cleanup() // TODO Rahul, Verify Later: need to check which call has owned the media at the moment(probably use active call)
```
rahul-lohra: directly use `media`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+
+            if (state.connection.value is RealtimeConnection.ReconnectingFailed) {
+                val message = "[reconnect] All recovery attempts exhausted — leaving call ($reason)"
+                logger.w { message }
+                callAnalytics.joinAnalytics.onJoinRequestRetryExhausted(
+                    loopIteration,
+                    AnalyticsCallAbortReason.RETRY_EXHAUSTED.name,
+                    message,
+                )
+                call.leave(
```
rahul-lohra: We can use `CallLifecycleManager` here

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+        reason: String,
+        joinAnalyticsModel: JoinAnalyticsModel,
+    ): ReconnectOutcome {
+        logger.d { "[reconnectRejoin] reconnectAttempts=${call.nonFastReconnectAttempts}" }
+        state._connection.value = RealtimeConnection.Reconnecting
+        val loc = call.location
+            ?: return ReconnectOutcome.PreconditionNotMet("No location available for rejoin")
+        val oldSession = session.value
+            ?: return ReconnectOutcome.PreconditionNotMet("No active session for rejoin")
+        call.reconnectStartTime = System.currentTimeMillis()
```
rahul-lohra: We should avoid updating call related properties from its encapsulated class. Use  `sessionManager.reconnectStartTime`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+            return ReconnectOutcome.Failed(
+                Exception("Failed to get join response: ${joinResponse.errorOrNull()}"),
+            )
+        }
+
+        val cred = joinResponse.value.credentials
+        val currentOptions = oldSession.publisher.value?.currentOptions()
+        logger.i { "Rejoin SFU ${oldSession.sfuUrl} to ${cred.server.url}" }
+
+        call.sessionId = UUID.randomUUID().toString()
```
rahul-lohra: Why not directly use `sessionManager.sessionId`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+                    "[reconnect] loopIteration=$loopIteration strategy=$currentStrategy reason=$reason"
+                }
+
+                val outcome = when (currentStrategy) {
+                    WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_FAST,
+                    WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_UNSPECIFIED,
+                    -> reconnectFast(reason)
+
+                    WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_REJOIN -> {
+                        call.nonFastReconnectAttempts++
```
rahul-lohra: Why not directly use `sessionManager.nonFastReconnectAttempts`

rahul-lohra: Same goes for other call properties that we are writing from this class

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+                            "[reconnect] $currentStrategy (${call.nonFastReconnectAttempts}) failed: ${outcome.error.message}"
+                        }
+
+                        delay(RECONNECT_DELAY_MS)
+                        loopIteration++
+
+                        val wasMigrating = currentStrategy ==
+                            WebsocketReconnectStrategy.WEBSOCKET_RECONNECT_STRATEGY_MIGRATE
+                        val pastFastReconnectDeadline = (System.currentTimeMillis() - loopStartTime) >
+                            call.reconnectDeadlineMillis
```
rahul-lohra: I feel this variable shouldn't be on call class `reconnectDeadlineMillis`. Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallReconnector.kt
```diff
@@ -0,0 +1,542 @@
...
+        state._connection.value = RealtimeConnection.Migrating
+        val loc = call.location
+            ?: return ReconnectOutcome.PreconditionNotMet("No location available for migrate")
+        val oldSession = session.value
+            ?: return ReconnectOutcome.PreconditionNotMet("No active session for migrate")
+        call.reconnectStartTime = System.currentTimeMillis()
+        addFailedSfuId(oldSession.sfuName)
+
+        val joinResponse =
+            call.joinRequest(
```
rahul-lohra: We can use `joinCoordinator.joinRequest`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +129,73 @@ public class Call(
...
+    internal var connectStartTime: Long
+        get() = sessionManager.connectStartTime
+        set(value) {
+            sessionManager.connectStartTime = value
+        }
+    internal var reconnectStartTime: Long
+        get() = sessionManager.reconnectStartTime
+        set(value) {
+            sessionManager.reconnectStartTime = value
+        }
```
rahul-lohra: Can we can move these out? I think it should be safe

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +129,73 @@ public class Call(
...
+        set(value) {
+            sessionManager.sessionId = value
+        }
+    internal val unifiedSessionId: String get() = sessionManager.unifiedSessionId
+
+    internal var location: String?
+        get() = sessionManager.location
+        set(value) {
+            sessionManager.location = value
+        }
```
rahul-lohra: We can move out this one as well, please check

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +129,73 @@ public class Call(
...
+
+    /** Session handles all real time communication for video and audio */
+    internal val session: MutableStateFlow<RtcSession?> get() = sessionManager.session
+
+    var sessionId: String
+        get() = sessionManager.sessionId
+        set(value) {
+            sessionManager.sessionId = value
+        }
+    internal val unifiedSessionId: String get() = sessionManager.unifiedSessionId
```
rahul-lohra: We can move out these 2 as well, please check

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -335,68 +264,38 @@ public class Call(
...
         )
 
+    /** Delegate that wraps all coordinator (REST) API calls for this call. */
+    private val apiClient by lazy { CallApiClient(this) }
+
+    /** Delegate that periodically collects and reports WebRTC stats. */
+    private val statsReporter by lazy { CallStatsReporter(this) }
+
+    /** Delegate that binds video tracks to renderers and handles media-quality overrides. */
+    private val callRenderer by lazy { CallRenderer(this) }
```
rahul-lohra: Do we really need lazy? and lazy with synchronised?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -335,68 +264,38 @@ public class Call(
...
-        if (wasMicrophoneEnabled) {
-            // audioTrack will be recreated on next access, then we enable it
-            microphone.enable(fromUser = false)
-        }
-        if (wasCameraEnabled) {
-            // videoTrack will be recreated on next access, then we enable it
-            camera.enable(fromUser = false)
-        }
-    }
+    internal fun recreateFactoryAndAudioTracks() = media.recreateFactoryAndAudioTracks()
```
rahul-lohra: Maybe we can remove 1 liner internal functions. Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -405,92 +304,52 @@ public class Call(
...
+    private val lifecycle = CallLifecycleManager(this)
+
+    /** Returns whether the device currently has network connectivity. */
+    internal fun isNetworkConnected(): Boolean = connectivityMonitor.isConnected()
+
+    /** Stops the ICE and connectivity monitors (used during teardown). */
+    internal fun stopConnectionMonitors() {
+        iceMonitor.stop()
+        connectivityMonitor.cancelLeaveTimeout()
+        connectivityMonitor.unsubscribe()
```
rahul-lohra: I think these functions should be written below `init {}`. Wdyt?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -707,181 +418,32 @@ public class Call(
...
-    private fun sendJoinErrorAnalytics(failure: SfuConnectionResult.Failure) {
-        callAnalytics.sfuAnalytics.onSfuWsCompleted(
-            success = false,
-            retryCount = session.value?.sfuWsRetryCount?.get() ?: 0,
-            failureReason = failure.error.message,
-            failureCode = (failure.abortReason ?: AnalyticsCallAbortReason.SFU_ERROR).name,
-        )
-    }
+    /** Resets the leave guard so a fresh join can run after a previous leave. */
+    internal fun resetLeaveGuard() = lifecycle.resetLeaveGuard()
```
rahul-lohra: Please see if we can remove these 1 liner delagation style method

rahul-lohra (commented):
Solid PR, please review the comments. They are mostly on code-quality. There is 1 related to usage of client scope

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallApiClient.kt
PratimMallick: Yup these changes we will do in seperate PRs since there many such small things. Want to keep this PR focused on single thing

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Its needed by RtcSession as well

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: This is needed by two components - CallJoinCordinator and CallReconnector

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Yes we can move out, but will do when we cleanup RtcSession

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: removed recreateFactoryAndAudioTracks() and recreatePeerConnectionFactory() (nothing called them through Call; tests hit CallMediaManager directly). Kept ensureFactoryMatchesAudioProfile() since CallJoinCoordinator uses it and inlining wouldn't decouple anything

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Done — moved both teardown functions below init {} so the class follows properties → init → functions ordering

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallConnectivityMonitor.kt
PratimMallick: done

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: I mean those clases `(CallJoinCordinator and CallReconnector)` can directly refer `SessionManager` instead of communicating back to call class

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
rahul-lohra: How about passing `sessionManager` reference to RtcSession, so there is no unnecessary link to call class

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Removed all of them

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallLifecycleManager.kt
PratimMallick: We need the call instance, have moved it out from here though

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +132,171 @@ public class Call(
...
      * Increment this only for REJOIN and MIGRATION strategies
      */
-    internal var nonFastReconnectAttempts = 0
-    internal val clientImpl = client as StreamVideoClient
-    internal val scopeProvider: ScopeProvider = ScopeProviderImpl(clientImpl.scope)
+    internal var nonFastReconnectAttempts: Int
+        get() = sessionManager.nonFastReconnectAttempts
+        set(value) {
+            sessionManager.nonFastReconnectAttempts = value
+        }
```
rahul-lohra: Check if you can remove this, this variable is only used in unit-tests

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +132,171 @@ public class Call(
...
+        callRegistry = callRegistry,
+        // Lets the REST calls queued before leave finish before the scope is torn down.
+        shutDownJobs = {
+            UserScope(ClientScope()).launch {
+                supervisorJob.children.forEach { it.join() }
+                supervisorJob.cancel()
+            }
+            scope.cancel()
+        },
+        stateProvider = { state },
```
rahul-lohra: We don't really need this. 
You can declare `state` variable before this variable

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +132,171 @@ public class Call(
...
+    )
+
+    /** Delegate that owns the unified reconnect state machine (fast / rejoin / migrate). */
+    private val reconnector: CallReconnector = CallReconnector(
+        clientImpl = clientImpl,
+        sessionManager = sessionManager,
+        sessionFactory = sessionFactory,
+        lifecycle = lifecycle,
+        sessionMonitor = { sessionMonitor },
+        stateProvider = { state },
```
rahul-lohra: We don't really need this. 
You can declare `state` variable before this variable

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +132,171 @@ public class Call(
...
+            UserScope(ClientScope()).launch {
+                supervisorJob.children.forEach { it.join() }
+                supervisorJob.cancel()
+            }
+            scope.cancel()
+        },
+        stateProvider = { state },
+        callAnalyticsProvider = { callAnalytics },
+        statsReporter = { statsReporter },
+        media = { media },
```
rahul-lohra: We can also skip this lambda

Declare these 2 above this variables
```kotlin
    /**
     * Creates [MediaManagerImpl] for this call. Captures `this` (and the test hook) so
     * [CallMediaManager] never needs a Call reference.
     */
    private val mediaManagerFactory = MediaManagerFactory { audioUsage, audioUsageProvider ->
        testInstanceProvider.mediaManagerCreator?.invoke()
            ?: MediaManagerImpl(
                clientImpl.context,
                this,
                scope,
                eglBase.eglBaseContext,
                audioUsage,
                audioUsageProvider,
            )
    }


    /** Delegate that owns the peer-connection factory, media manager and audio pipeline. */
    private val media = CallMediaManager(
        type = type,
        id = id,
        clientImpl = clientImpl,
        scope = scope,
        state = state,
        sessionManager = sessionManager,
        eglBase = { eglBase },
        mediaManagerFactory = mediaManagerFactory,
    )
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +132,171 @@ public class Call(
...
+    /** Delegate that owns the unified reconnect state machine (fast / rejoin / migrate). */
+    private val reconnector: CallReconnector = CallReconnector(
+        clientImpl = clientImpl,
+        sessionManager = sessionManager,
+        sessionFactory = sessionFactory,
+        lifecycle = lifecycle,
+        sessionMonitor = { sessionMonitor },
+        stateProvider = { state },
+        callAnalyticsProvider = { callAnalytics },
+        statsReporter = { statsReporter },
```
rahul-lohra: We can also skip this lambda by moving this line up
```kotlin
/** Delegate that periodically collects and reports WebRTC stats. */
    private val statsReporter = CallStatsReporter(type, id, scope, sessionManager, state)
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -185,44 +132,171 @@ public class Call(
...
+        joinCoordinator = { joinCoordinator },
```
rahul-lohra: We introduced a circular dependency where. 
1.  `CallReconnector` depends on `JoinCoordinator`
2.   `JoinCoordinator` depends on  `CallReconnector`

Try to extract the shared logic into separate class

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -405,92 +434,74 @@ public class Call(
...
 
-    private var monitorPublisherPCStateJob: Job? = null
-    private var monitorSubscriberPCStateJob: Job? = null
-    private var sfuListener: Job? = null
-    private var sfuEvents: Job? = null
+    internal var reconnectDeadlineMillis: Int
+        get() = sessionManager.reconnectDeadlineMillis
+        set(value) {
+            sessionManager.reconnectDeadlineMillis = value
+        }
```
rahul-lohra, line 477: We can actually avoid creating this variable, it is referred only at one place
and we directly pass `sessionManager.reconnectDeadlineMillis`

PratimMallick: Yes this one is simplified now

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Done in 543439f293. `state` is now declared above `lifecycle`, so `CallLifecycleManager` takes `state: CallState` directly instead of `stateProvider: () -> CallState`.

The blocker was a cycle rather than ordering: `CallEventManager` took the reconnector eagerly, and `state` is built from `events`, so `reconnector -> state -> events -> eventManager -> reconnector` forced everything downstream to be lazy. Making `CallEventManager` take `reconnector: () -> CallReconnector` breaks it in one place and lets `state` (and the whole middle tier) be declared early.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Done in 543439f293 — same change on `CallReconnector`: `stateProvider: () -> CallState` is now `state: CallState`, since `state` is declared above it.

`callAnalytics` got the same treatment in both components.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Done in 543439f293. `mediaManagerFactory` and `media` are now declared before `lifecycle`, so `CallLifecycleManager` takes `media: CallMediaManager` directly. `iceMonitor` is injected directly for the same reason.

`eglBase` stays a lambda on `CallMediaManager`, but deliberately: it keeps the native EGL context from being created for calls that never render.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Done in 543439f293. `CallStatsReporter(type, id, scope, sessionManager, state)` moved up, and both `CallLifecycleManager` and `CallReconnector` now take `statsReporter: CallStatsReporter` directly.

After this round the only remaining `() -> T` parameters are a genuine cycle (`sessionMonitor` / `connectivityMonitor`), the deferred `eglBase`, and two live reads of mutable state (`callSessionId`, `reconnectDeadlineMillis`). There's a comment above the component graph in `Call.kt` spelling out which is which, so the next reader doesn't have to guess.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Good catch — fixed in 543439f293, and your suggested approach worked, with one wrinkle worth recording.

The two edges were `CallReconnector -> CallJoinCoordinator` for `joinRequest(...)` only, and `CallJoinCoordinator -> CallReconnector` for `reconnect(...)` plus `getFailedSfuIdsSnapshot()`. The wrinkle: that second call was *inside `joinRequest` itself*, so extracting `joinRequest` on its own would have produced `JoinRequester -> Reconnector -> JoinRequester` — the same cycle under a new name.

So two moves instead of one:

1. `joinRequest` moved into `CallApiClient` rather than a new class. It is a coordinator REST call that updates `CallState` from the response, which is exactly what that component's KDoc already claims it owns, so this avoids adding a thirteenth component for one method.
2. The failed-SFU set moved from a private field on `CallReconnector` to `CallSessionManager`. It is shared join/reconnect bookkeeping — written by the reconnect flow, read when building a join request — so it belongs with the other shared session state rather than inside one of the two consumers.

Now both orchestrators depend on `CallApiClient`, `CallApiClient` depends on neither, and `CallJoinCoordinator -> CallReconnector` is one-way for `reconnect(...)`. The `joinCoordinator: () -> CallJoinCoordinator` provider is gone.

Side benefit: `FailedSfuIdsTest` was reaching into the reconnector through four reflection helpers (`getDeclaredField("failedSfuIds")`, `getDeclaredMethod("addFailedSfuId")`, ...). All four are gone, replaced by two one-line test seams, and the behaviour is now covered directly in `CallSessionManagerTest`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: removed

rahul-lohra (approved):
All queries are resolved now. Good work 👍

PR #1752, opened 2026-07-23
Make the Android Video SDK compatible with Android 17.

rahul-lohra:
### Goal

Closes AND-1324

Make the Android Video SDK compatible with Android 17.

Android 17 introduces new restrictions on background media playback [Link](https://developer.android.com/about/versions/17/behavior-changes-17#bg-audio) . Previously, the SDK relied on the `SHORT_SERVICE` foreground service type to play the incoming call ringtone while the app was in the background. This approach is no longer suitable on Android 17.

Solution:
For incoming calls on Android 17, replace `SHORT_SERVICE` with `ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL`.

Additional changes added to compile with compile SDK 37
1. StreamVideoClient.kt
2. PermissionManager.kt

### Implementation

Replace `SHORT_SERVICE` with `ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL` for the incoming call foreground service on Android 17.

### 🎨 UI Changes

None

### Testing

1. Launch Android 17 Emulator
2. Increase all the sound of that emulator to verify audio of incoming call
3. Receive an incoming call on that emulator and verify incoming ringing call is playing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of incoming calls on newer Android versions.
  * Incoming ringing now uses the phone-call foreground service type on Android 17 and later, while retaining short-service behavior on earlier versions.

* **Tests**
  * Added coverage to verify incoming-call service behavior across Android versions.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
```diff
@@ -91,6 +92,24 @@ internal open class ForegroundServicePermissionManager {
...
+     *
+     * On Android 17+, background audio (the ringtone) is muted under [ServiceInfo.FOREGROUND_SERVICE_TYPE_SHORT_SERVICE],
+     * so phoneCall-capable VoIP services use phoneCall instead. Other services keep [noPermissionServiceType].
+     */
+    @SuppressLint("InlinedApi")
+    internal open fun incomingRingingServiceType(): Int =
+        // TODO: replace the hardcoded 37 with Build.VERSION_CODES.<ANDROID_17> once compileSdk /
+        //  targetSdk are raised to 37; the constant does not exist at the current compileSdk.
+        if (requiredForegroundTypes.contains(ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL) &&
+            Build.VERSION.SDK_INT >= 37
```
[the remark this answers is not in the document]
rahul-lohra: The API-level literal is not available at the moment, we will change it once we update to higher SDK

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManagerTest.kt
```diff
@@ -99,6 +100,40 @@ class ForegroundServicePermissionManagerTest {
         assertEquals(0, type)
     }
 
+    @Test
+    @Config(sdk = [Build.VERSION_CODES.UPSIDE_DOWN_CAKE])
+    fun `incoming call keeps short service below android 17`() {
```
aleksandar-apostolov, line 116: Coverage suggestion: this only exercises API 34, but Android 16 (API 36 / `BAKLAVA`) is the current stable release and the most likely spot for an accidental regression. Could we add a case at 36 (and ideally 35) so "SHORT_SERVICE is preserved on Android 15/16" is locked by a test rather than inferred from the shared branch? A parameterized `@Config(sdk = [...])` would cover all three cheaply.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
```diff
@@ -91,6 +92,24 @@ internal open class ForegroundServicePermissionManager {
...
     }
 
+    /**
+     * Foreground-service type for the incoming-ringing window.
+     *
+     * On Android 17+, background audio (the ringtone) is muted under [ServiceInfo.FOREGROUND_SERVICE_TYPE_SHORT_SERVICE],
+     * so phoneCall-capable VoIP services use phoneCall instead. Other services keep [noPermissionServiceType].
+     */
+    @SuppressLint("InlinedApi")
+    internal open fun incomingRingingServiceType(): Int =
```
aleksandar-apostolov, line 103: `requiredForegroundTypes` is `open`, so a subclass could override it without `PHONE_CALL`. On 37+ that should fall through to `noPermissionServiceType()`, but there's no test for that path. Narrow case, low priority — just flagging since the extension point is exposed to subclasses.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
aleksandar-apostolov: The hardcoded `37` appears here and in the test. The TODO already explains why the constant isn't available yet — could we extract a single `private const val ANDROID_17_SDK = 37` and reference it from both? One place to update when `Build.VERSION_CODES` catches up.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManagerTest.kt
```diff
@@ -99,6 +100,40 @@ class ForegroundServicePermissionManagerTest {
...
+            type,
+        )
+    }
+
+    @Test
+    fun `incoming call uses phone call service type on android 17 and above`() {
+        // Android 17 background-audio hardening mutes the ringtone under SHORT_SERVICE; use the
+        // while-in-use phoneCall type instead. Robolectric 4.11.1 caps @Config at API 34, so the
+        // >= 37 branch is exercised by forcing SDK_INT.
+        ReflectionHelpers.setStaticField(Build.VERSION::class.java, "SDK_INT", 37)
```
aleksandar-apostolov: Nit: `SDK_INT` isn't restored after this. Robolectric's per-method sandbox means it won't leak today, so this is fine — just noting a `@Config`-based approach or a try/finally reset would be more robust if that isolation ever changes.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
rahul-lohra: Sure

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManagerTest.kt
rahul-lohra: Nice suggestion.

Our current Robolectric version (4.11.1) only supports @Config up to API 34, so we can’t include API 35/36 in the parameterized test.

I’ll add a separate test to explicitly verify the behavior on newer Android versions and prevent regressions.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
rahul-lohra: Right

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManagerTest.kt
rahul-lohra: Right

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManager.kt
rahul-lohra: Added a new test-case

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/notifications/internal/service/permissions/ForegroundServicePermissionManagerTest.kt
rahul-lohra: Added 2 new tests

aleksandar-apostolov (approved):

PR #1758, opened 2026-07-31
Safeguard all call class's public methods against NPE

rahul-lohra:
### Goal
Closes #AND-1350

We got a crash with stacktrace

```kotlin
Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
io.getstream.video.android.core.Call._join$stream_video_android_core_release (Call.kt:788)
io.getstream.video.android.core.Call$_join$1.invokeSuspend (Call.kt:17)
kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:34)
```

Since we were unable to reproduce the crash with the exact stack trace, we decided to defensively guard all public methods of the Call class against potential `NullPointerExceptions`.

### Implementation

Wrap all `Call` class's public method which returns `Result`  with `try-catch` 

```kotlin
@OptIn(ExperimentalContracts::class)
internal inline fun <T : Any> runResultCatchingCancellable(block: () -> Result<T>): Result<T> {
    contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
    return try {
        block()
    } catch (ce: CancellationException) {
        throw ce
    } catch (npe: NullPointerException) {
        Result.Failure(Error.ThrowableError(npe.message ?: "NullPointerException caught", npe))
    }
}
```
### 🎨 UI Changes

None

### Testing

Smoke test video/audio call


PratimMallick (approved):

rahul-lohra:
Closing this PR — the NPE crash is already resolved in #1740, released in **v1.29.0**.

**Note on reproducing the exact stack trace**

The reported message only appears on a release build with **AGP 9+**:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
```

**Why:** `session.value!!` is a null-assertion that R8 compiles differently per version. On **AGP 9+**, R8 rewrites it into an `Object.getClass()` call, so a null value produces the `getClass()` message above. On **AGP 8**, R8 kept it as `Intrinsics.checkNotNull`, which throws a plain `NullPointerException` without that text — same crash, different wording. This is why it couldn't be reproduced with the exact message on our AGP 8 builds.

PR #1763, opened 2026-08-06
Fix peer connection previous-connect analytics scope

rahul-lohra:
### Goal

Closes #AND-1368

Ensure `wasPreviouslyConnected` is reported only for the relevant peer connection role within the current call analytics lifecycle, instead of leaking cached state from previous calls.

### Implementation
Moved peer connection “ever connected” tracking out of the client-level reporter state and into `PeerConnectionAnalyticsStateHolder`, so the value is scoped to peer connection analytics for the current call. `PeerConnectionAnalytics` now updates publisher/subscriber connected state when a `CONNECTED` event is observed and passes the role-specific `wasPrevConnected` value into `ClientEventReporter`.

### 🎨 UI Changes

None

### Testing

- first connection attempts report `wasPrevConnected = false`
- reconnects after a successful connection report `wasPrevConnected = true`
- publisher and subscriber previous-connect state are tracked independently
- reporter tests use the explicit `wasPrevConnected` input instead of relying on cached client-level state
- Verify manually in backend logs

## Summary by CodeRabbit

* **Analytics**
  * Improved peer-connection analytics to accurately identify whether publisher and subscriber connections have previously connected.
  * Reconnection events now report the correct prior-connection status for clearer connection lifecycle insights.

* **Tests**
  * Added coverage for initial connections, failures, reconnects, and publisher/subscriber-specific connection states.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/PeerConnectionAnalytics.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/PeerConnectionAnalytics.kt
@@ -162,16 +162,7 @@ internal class PeerConnectionAnalytics(
         iceState: VideoAnalyticsIceState,
         peerConnectionState: PeerConnection.PeerConnectionState?,
     ) {
-        when (peerConnectionState) {
-            PeerConnection.PeerConnectionState.CONNECTED -> {
-                if (role == PeerConnectionRole.PUBLISH) {
-                    stateHolder.updatePublisherEverConnected(true)
-                } else {
-                    stateHolder.updateSubscriberEverConnected(true)
-                }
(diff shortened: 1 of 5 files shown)


aleksandar-apostolov (approved):
LGTM on the fix — clean re-scope of `wasPreviouslyConnected` into the per-call state holder. Please add a test asserting a new call starts `wasPrevConnected = false` after a prior call connected — that's the actual AND-1368 regression and it's currently uncovered.

rahul-lohra:
> Please add a test asserting a new call starts wasPrevConnected = false after a prior call connected — that's the actual AND-1368 regression and it's currently uncovered.


Added 2 tests to cover this case
1. a new call starts with wasPrevConnected set to false
2. an existing call whose Publisher was previously connected sends wasPrevConnected as true

PR #1764, opened 2026-08-10
Single-flight Call.join to stop concurrent-join race

PratimMallick:
### Goal

Fixes [AND-1376](https://linear.app/stream/issue/AND-1376/fix-concurrent-join-race)

Prevent overlapping `Call.join()` calls from creating multiple `RtcSession`s that share the same `sessionId`. The SFU keeps only the latest participant and evicts the others, which leaves zombie publishers that cannot publish A/V and often fail subsequent RPCs with `PARTICIPANT_NOT_FOUND`, triggering reconnect/rejoin loops.

Also fix a related footgun: calling `join()` again while already joined used to return `Failure` **and** clear the live session / set `RealtimeConnection.Failed`, which tore down a healthy call (easy to hit with accidental double-join).

### Implementation

- Extract `StreamRefCountedSingleFlightProcessor`: keyed single-flight that runs shared work on the **call scope**, tracks waiters, and cancels the shared job only when the **last** waiter is cancelled (one UI cancel does not kill other waiters / auto-join). Last waiter always detaches the map entry even if the deferred is already dead.
- Wire `CallJoinCoordinator.join()` through that processor so concurrent callers share one join attempt and once-only setup (telemetry, interceptor, leave guard, `InProgress`).
- Subsequent `join()` / `joinInternal()` while a session already exists returns `Success(existing)` (idempotent) instead of failing and tearing down the call.
- SFU traces + `logger.w` for already-joined (`join-already-joined`) and coalesced concurrent joins (`join-coalesced`), including a warning when a coalesced caller’s interceptor differs from the in-flight one.
- Keep `discardFailedSession()` cleanup on SFU connect failure during join so failed sessions do not keep issuing RPCs after eviction.
- Unit tests for the processor (coalesce, `onCoalesced`, cancel-one / cancel-last, cancelled-scope detach) and join coordinator (concurrent join, already-joined Success, traces, cleanup).

**Behavior notes for reviewers**
- Cancelling one of several concurrent `join()` waiters no longer cancels the shared join; cancelling the last waiter does.
- `join()` while already joined: `Failure("already been joined")` → `Success(existing session)` (intentional API softening; avoids destroying a live call).
- Porting the refcounted processor into `stream-android-core` is follow-up, not this PR.

### Testing

- [x] `./gradlew :stream-video-android-core:spotlessApply`
- [x] `./gradlew :stream-video-android-core:testDebugUnitTest --tests 'io.getstream.video.android.core.call.components.CallJoinCoordinatorTest'` — passed
- [x] `./gradlew :stream-video-android-core:testDebugUnitTest --tests 'io.getstream.video.android.core.utils.StreamRefCountedSingleFlightProcessorTest'` — passed
- [x] Manual dogfood (debug UI reverted before ship): concurrent lobby/activity joins coalesce to one SFU join; in-call re-join returns existing session without finishing the activity

<details>
<summary>Failure modes this mitigates</summary>

```
- Duplicate SFU joinRequests for the same session_id / unified_session_id
- Zombie RtcSessions after SFU participant eviction
- PARTICIPANT_NOT_FOUND on SetPublisher / UpdateMuteStates / IceTrickle / sendAnswer
- Publisher PC thrash (NEW→CLOSED for losers; survivor stuck CHECKING)
- Inability to publish audio/video after a “successful” join UI
- Cascading full-rejoin / reconnect loops driven by those RPC failures
- Accidental second join() tearing down an already-connected call
```

</details>

#### General

#### Code & documentation

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

N/A — core join orchestration fix, no UI changes.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -312,6 +387,7 @@ internal class CallJoinCoordinator(
                     if (!didReconnectSucceed()) {
                         logger.e { "[_join] Could not recover. Error : $sfuConnectionResult" }
                         sendJoinErrorAnalytics(sfuConnectionResult)
+                        discardFailedSession(localSession)
```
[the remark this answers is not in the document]
PratimMallick: Agreed the interaction was worth looking at, but we fixed it the other way.

Once join is returning Failure after !didReconnectSucceed(), there’s no reason to keep any session active — neither the join’s localSession nor a reconnect replacement. Preserving the replacement would leave live sockets/PCs after a failed join.

Follow-up commit removes the ownership early-return in discardFailedSession: it always clears the active slot and cleanup()s both the join session and any replacement. Added a unit test for the recovery-failed + replacement case.

aleksandar-apostolov:
I think porting the SingleFlight mechanism from core and then using it here would be easier for migration than this in-line implementation. WDYT?

PratimMallick:
> I think porting the SingleFlight mechanism from core and then using it here would be easier for migration than this in-line implementation. WDYT?

The one from core runs on its own scope(which is the call scope), whereas for join we want to run in the caller's scope(UI/viewmodel). Hence used a newer way

aleksandar-apostolov:
> > I think porting the SingleFlight mechanism from core and then using it here would be easier for migration than this in-line implementation. WDYT?
> 
> The one from core runs on its own scope(which is the call scope), whereas for join we want to run in the caller's scope(UI/viewmodel). Hence used a newer way

But you can create a `val callFlights = SingleFlight(whateverScope)` no?, My point is, we can re-use that implementation and when we merge this to v2, we remove the ported impl and the `Call` part remains the same. 

Or do you mean to use the caller scope, like the UI scope for example to rely on the scope cancellation for join cancellation also?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,149 @@
...
+        if (cancelRunning) {
+            flights.values.forEach { it.deferred.cancel() }
+        }
+        flights.clear()
+    }
+
+    fun stop(): Result<Unit> = runCatching {
+        if (closed.compareAndSet(false, true)) {
+            clear(cancelRunning = true).getOrThrow()
+        }
```
rahul-lohra, line 277: `flights` are not thread-safe on these method(s). Is this intentional?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -77,6 +103,42 @@ internal class CallJoinCoordinator(
         hintHighScaleLivestreamPublisher: Boolean? = null,
         callJoinInterceptor: CallJoinInterceptor? = null,
     ): Result<RtcSession> {
+        return joinFlight.run(JOIN_FLIGHT_KEY) {
```
rahul-lohra: Using same key has a risk here.
What if the arguments are different 
like - 
 `create = false` and `create = true`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,149 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.video.android.core.utils
+
+// package io.getstream.android.core.internal.processing
```
rahul-lohra: we can remove this comment

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,149 @@
...
+
+        var released = false
+        suspend fun releaseWaiter(cancelIfLast: Boolean) {
+            if (released) return
+            released = true
+            val shouldCancelShared = mutex.withLock {
+                flight.waiters = (flight.waiters - 1).coerceAtLeast(0)
+                cancelIfLast && flight.waiters == 0 && flight.deferred.isActive
+            }
+            if (shouldCancelShared) {
```
rahul-lohra: This block is not protected by mutex, so a new flight with same key can enter right at this point. Please double check and we should add a unit-test to protect this

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Runs [block] once for [key] while concurrent callers await the same result.
+     *
+     * Returns [Result.failure] with a [ClosedSendChannelException] if [stop] has already
+     * been called. [CancellationException] is still rethrown when this waiter (or the shared
+     * job, including last-waiter cancel) is cancelled.
+     */
+    @Suppress("UNCHECKED_CAST")
+    suspend fun <T> run(key: String, block: suspend () -> T): Result<T> {
```
rahul-lohra: Can we simplify this function? Because it is very complex state machine inside 1 function
The outer public should ideally expose the high level the algorithm

Use this for reference - 

```text
run
├── acquireWaiter
│   └── selectFlightLocked
│       ├── createFlightLocked
│       └── removeFlightIfCurrentLocked
└── awaitSharedResult
    └── releaseWaiter
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: agreed

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: although these helper functions were not getting used by the Join case, but we should make them thread safe. Doing it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: okay, let me refactor this a bit

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: sure

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: this was kept intentionally , since having different keys based on args will recreate the same zoombie call issues. I dont think any regular use case will demand different join calls within milliseconds. It will always be due to a bug that we end up with these type of cases where one type of join is in flight multiple other joins of the same type comes in

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,149 @@
...
+        if (closed.get()) {
+            return Result.failure(ClosedSendChannelException("RefCountedSingleFlight is closed"))
+        }
+
+        val flight = mutex.withLock {
+            val running = flights[key]?.takeUnless { it.deferred.isCompleted } as Flight<T>?
```
rahul-lohra: There appears to be an uncovered race when a new caller arrives while the previous flight is still cancelling.

Consider this sequence:

1. `run("join", blockA)` starts and has one waiter.
2. The waiter is cancelled, reducing `waiters` to `0` and cancelling the shared deferred.
3. The deferred enters `Cancelling` but has not completed its `finally` cleanup.
4. Before cancellation completes, another caller invokes `run("join", blockB)`.

During `Cancelling`, a deferred can have:

```kotlin
isActive == false
isCompleted == false
```

The current lookup only excludes completed flights:

```kotlin
flights[key]?.takeUnless { it.deferred.isCompleted }
```

Therefore, the second caller can attach to the cancelling flight because `isCompleted` is still `false`. `blockB` is never executed, and the second caller eventually receives the previous flight’s `CancellationException`.

It can happen whenever ownership moves quickly between callers—for example, an Activity/ViewModel waiter is cancelled and another UI or auto-join caller invokes `join()` before the old flight finishes cancellation.

Could we add a test where the first block deliberately pauses inside non-cancellable cancellation cleanup, then start a second `run()` during that window? The expected behaviour should be:

1. The second caller does not attach to the cancelling flight.
2. It waits for the old flight to finish cleanup.
3. It then starts a fresh flight and executes `blockB`.

Checking only `deferred.isActive` would prevent attaching to the cancelling flight, but immediately starting a replacement could briefly overlap the old flight’s cleanup.

Please double check

PratimMallick: Good catch on the `Cancelling` window (`isActive == false`, `isCompleted == false`).

That attach path is closed in the latest commit by keeping map removal and `deferred.cancel()` in the **same** `mutex.withLock` as the last-waiter decrement:

1. `waiters → 0`
2. remove this flight from `flights`
3. `cancel()`
4. unlock

So when a new `run("join")` takes the mutex next, the key is already gone and it starts a fresh flight — it never joins the dying deferred via `!isCompleted`.

Mutex alone would not be enough if we left the entry in the map; removal under that same lock is what prevents the sequence you described.

Added regression coverage for last-waiter cancel + a concurrent new `run()` (must not fail with the old flight’s `CancellationException`).

Happy to tighten further with an `isActive` guard as defense-in-depth if you still want that.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Resolved in `dec065eacf`.

`flights` is now a `ConcurrentHashMap` (same approach as `StreamSingleFlightProcessorImpl`), so `has` / `cancel` / `clear` / `stop` are safe without taking the coroutine mutex. Waiter counting and select/create/remove for `run()` still go through `mutex.withLock`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Resolved in `dec065eacf` — removed the leftover `// package io.getstream...` comment.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Resolved in `dec065eacf`.

Last-waiter release now does decrement → remove from `flights` → `deferred.cancel()` under a single `mutex.withLock`, so a new `run()` cannot attach between unlock and cancel.

Added regression tests for last-waiter cancel + a concurrent new `run()` (must not fail with the dying flight’s `CancellationException`).

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Resolved in `dec065eacf`.

`run` is split along the structure you suggested:

```
run
├── acquireWaiter
│   └── selectFlightLocked
│       ├── createFlightLocked
│       └── removeFlightIfCurrentLocked
└── awaitSharedResult
    └── releaseWaiter
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Keeping the single `JOIN_FLIGHT_KEY` as discussed — no code change here.

Keying by args (`create` / `ring` / etc.) would reintroduce concurrent joins for the same `Call` and the zombie / `PARTICIPANT_NOT_FOUND` race this PR is fixing. First in-flight caller’s args win; overlapping joins with different flags in the same millisecond aren’t a supported product path.

rahul-lohra: okay

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -366,4 +486,8 @@ internal class CallJoinCoordinator(
         logger.d { "[_join] Reconnect after recoverable connection failure settled on $terminal" }
         return terminal is RealtimeConnection.Connected
     }
+
+    private companion object {
+        const val JOIN_FLIGHT_KEY = "join"
+    }
```
rahul-lohra: Its better to declare companion class at the top of the class.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,187 @@
...
+    ): Flight<T> = mutex.withLock {
+        selectFlightLocked(key, block)
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun <T> selectFlightLocked(
+        key: String,
+        block: suspend () -> T,
+    ): Flight<T> {
+        val running = flights[key]?.takeUnless { it.deferred.isCompleted } as Flight<T>?
```
rahul-lohra: Do not reuse a cancelling flight

`Deferred.isCompleted` remains false while the job is in the `Cancelling` state. If `cancel(key)`, `clear(true)`, or the parent scope cancels this flight before its `finally` removes the map entry, a new caller can attach to the dying flight and receive `CancellationException` without executing its block. Please only reuse flights whose deferred is `isActive`, and remove a flight from the map before cancelling it under the same synchronization boundary. Add a deterministic regression test that holds the old block in `NonCancellable` cleanup, cancels it, starts another `run` for the same key, and verifies that the new block executes as a fresh flight.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: okay ..moving it

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * Returns [Result.failure] with a [ClosedSendChannelException] if [stop] has already
+     * been called. [CancellationException] is still rethrown when this waiter (or the shared
+     * job, including last-waiter cancel) is cancelled.
+     */
+    suspend fun <T> run(key: String, block: suspend () -> T): Result<T> {
+        if (closed.get()) {
+            return Result.failure(ClosedSendChannelException("RefCountedSingleFlight is closed"))
+        }
+        val flight = acquireWaiter(key, block)
```
rahul-lohra: Serialize the closed check with flight creation

The `closed` check happens before acquiring `mutex`, so `run()` can observe `false`, lose the race to `stop()`, and then create a new flight after the processor has been closed and cleared. Please check `closed` inside the same critical section that selects or creates the flight, and make `stop()` close and detach existing flights through that synchronization boundary.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Fixed in `e0094292f7`.

Lookup now reuses a flight only while `deferred.isActive` — a `Cancelling` job (`isCompleted == false`) is no longer joinable. `cancel(key)`, `clear(true)`, last-waiter cancel, and `stop()` all remove from the map then cancel under the same mutex.

Added a regression test that holds the old block in `NonCancellable` cleanup, calls `cancel(key)`, then starts another `run` for the same key and asserts the new block executes as a fresh flight.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Fixed in `e0094292f7`.

The `closed` check now sits inside the same `mutex.withLock` that selects/creates the flight (`acquireWaiter`). `stop()` is suspend and uses that lock: set `closed`, snapshot, clear the map, then cancel. A `run()` that loses the race to `stop()` sees `closed` and returns `ClosedSendChannelException` instead of installing a new flight.

Test: `stop` while an old flight is still running, then a later `run` is rejected.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,207 @@
...
+ * - Cancelling the **last** waiter **does** cancel the shared job (sole-caller cancel still
+ *   aborts the operation).
+ * - [CancellationException] is rethrown to the cancelled waiter instead of being wrapped in
+ *   [Result.failure].
+ *
+ * Use this when the operation should survive Activity/ViewModel teardown of some waiters
+ * (e.g. screen handoff, UI join + call-scoped auto-join) but should not keep running after
+ * nobody is waiting — unless [scope] itself is cancelled (leave / call cleanup).
+ *
+ * Candidate for Stream Android Core v2 alongside [StreamSingleFlightProcessorImpl].
```
aleksandar-apostolov, line 50: Re-ran this at HEAD after `e0094292` — the cancelling-attach and `stop()` races are both closed, and the two new tests are the deterministic shape. Verified rather than taken on the diff.

On the KDoc though: `stream-android-core` already ships `StreamSingleFlightProcessor` with this exact `run`/`has`/`cancel`/`clear`/`stop` surface, and video already carries a copy of `StreamSingleFlightProcessorImpl`. This makes three. The ref-counting is the genuinely new idea and core is where it belongs — worth a follow-up rather than growing a third copy here.

Two things that point the same way: `cancel`/`clear`/`stop` just became `suspend` to serve the mutex, which makes them awkward from a non-suspending teardown path, and nothing calls them. Core avoids the mutex entirely by evicting on `invokeOnCompletion` with a conditional `remove(key, value)` (core#73) — non-suspending, and the Cancelling window cannot exist.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -77,6 +111,44 @@ internal class CallJoinCoordinator(
...
+        createOptions: CreateCallOptions?,
+        ring: Boolean,
+        notify: Boolean,
+        hintHighScaleLivestreamPublisher: Boolean?,
+        callJoinInterceptor: CallJoinInterceptor?,
+    ): Result<RtcSession> {
+        // Single already-joined gate for the whole join flow: subsequent join() calls while a
+        // session is live return that session instead of building a second one. Every retry
+        // below clears the session first, so [joinInternal] always starts without one.
+        sessionManager.session.value?.let { existing ->
```
aleksandar-apostolov, line 222: Telemetry deltas worth a deliberate call, since join reporting matters:

- This gate returns before `onJoinFunctionStart()`, so `join()` on a live call now emits nothing and returns Success. Previously it emitted the join start and reported a Failure. Accidental double-join — the footgun this PR fixes — goes invisible, so we lose the ability to measure how often it happens in the field.
- N concurrent joins now emit one `onJoinFunctionStart()` instead of N. Right as a count of real joins, but a step change in the metric.
- A coalesced caller's `callJoinInterceptor` is dropped silently — `state.callJoinInterceptor` is a single slot assigned inside the flight.

Suggest a distinct signal on the already-joined and coalesced paths instead of going quiet, plus a `logger.w` when a coalescing caller passes a different interceptor than the in-flight one.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -202,6 +274,14 @@ internal class CallJoinCoordinator(
...
 
+    /**
+     * Performs one join attempt: coordinator round-trip, [RtcSession] creation and SFU connect.
+     *
+     * Assumes no session is active — [executeJoin] owns the already-joined check and clears the
+     * session before every retry. Calling this with a live session would build a second
+     * [RtcSession] for the same `sessionId` and produce the SFU-evicted zombie this coordinator
+     * exists to prevent.
+     */
     suspend fun joinInternal(
```
aleksandar-apostolov, line 360: This drops the guard and replaces it with a comment. `joinInternal` is still module-visible with default args and is called directly by `CallJoinCoordinatorTest` and `JoinRecoverableFailureTest`, so the only thing stopping a second `RtcSession` on the same `sessionId` is now KDoc. `executeJoin` clears the session before every retry, so putting the early return back is free and keeps the invariant enforced in code.

aleksandar-apostolov:
Goal section: `closes [AND-1379](https://linear.app/stream/issue/AND-1376/fix-concurrent-join-race)` — the text and the link point at different tickets. AND-1379 is a separate issue (ICE restart / rejoin escalation). Both AND-1376 and AND-1379 currently have this PR attached and both sit in In Review, so AND-1379 reads as in-progress when nothing here touches it. Should be `Fixes AND-1376`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
```diff
@@ -0,0 +1,207 @@
...
+        var released = false
+        suspend fun releaseWaiter(cancelIfLast: Boolean) {
+            if (released) return
+            released = true
+            // Decrement, map removal, and cancel must stay under one lock so a new run() cannot
+            // attach to a flight that is about to be cancelled (waiters already at 0).
+            mutex.withLock {
+                flight.waiters = (flight.waiters - 1).coerceAtLeast(0)
+                if (cancelIfLast && flight.waiters == 0 && flight.deferred.isActive) {
+                    cancelAndDetachLocked(flight)
```
rahul-lohra: ### [P2] Detach cancelled flights even when the deferred is no longer active

**Issue**

A deferred created from an already-cancelled scope may be cancelled before its coroutine body starts. In that case, the body’s `finally` block may never run.

The current waiter cleanup removes the flight only when the deferred is still active:

```kotlin
if (cancelIfLast && flight.waiters == 0 && flight.deferred.isActive) {
    cancelAndDetachLocked(flight)
}
```

Because an already-cancelled deferred has `isActive == false`, its entry can remain in `flights` after the final waiter leaves.

**Risk**

This is particularly risky when the processor will support replacing or reusing its scope:

- `has(key)` can incorrectly report that an inactive flight exists.
- Flights created by the old scope remain associated with that scope.
- Supplying a new scope cannot restart or migrate old deferreds.
- Calls using unique keys can accumulate stale entries.
- Old and new scope work could overlap unless replacement clears existing flights first.

**Solution**

Always remove a flight when its final waiter leaves. Cancellation should remain conditional because completed or already-cancelled deferreds do not need to be cancelled again.

```kotlin
flight.waiters--

if (flight.waiters == 0) {
    if (cancelIfLast && flight.deferred.isActive) {
        cancelAndDetachLocked(flight)
    } else {
        removeFlightIfCurrentLocked(
            key = flight.key,
            deferred = flight.deferred,
        )
    }
}
```

When scope replacement is added, we will clear and cancel all existing flights before exposing the new scope. The replacement must use the same mutex so it cannot race with `run()`:

```kotlin
mutex.withLock {
    val oldFlights = flights.values.toList()

    flights.clear()
    oldFlights.forEach { it.deferred.cancel() }

    scope = newScope
}
```

A `Deferred` permanently belongs to the scope that created it, so old flights must never be carried over to the new scope.

rahul-lohra (approved):
Nice work, left [P2] comment about clearing stale flights before introducing a reusable coroutine scope. It does not block this PR, so I’m approving it.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Agreed this is a third copy of the single-flight surface.

We will keep the video-local `StreamRefCountedSingleFlightProcessor` for this PR so join coalescing can land. Porting the refcounted variant into `stream-android-core` (next to `StreamSingleFlightProcessor`, and ideally onto `invokeOnCompletion` instead of a mutex) is follow-up — not rewriting that here.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Done in the latest commit.

- Already-joined `join()` / `joinInternal()`: `logger.w` and `sfuTracer.trace("join-already-joined", …)` and still no `onJoinFunctionStart()`.
- N concurrent callers: one real join start; coalesced waiters log a warning and `sfuTracer.trace("join-coalesced", …)` after the shared join.
- Coalesced caller whose `callJoinInterceptor` is not the in-flight one: extra `logger.w`.

Left these as traces + logs rather than new join-start analytics events so they stay distinguishable from a real join.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Restored the already-joined Success guard on `joinInternal` (still `internal`; tests call it directly) plus a unit test. `executeJoin` keeps the same gate for the public `join()` path.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/utils/StreamRefCountedSingleFlightProcessor.kt
PratimMallick: Fixed: when `waiters` hits 0 we always remove the flight. `cancel()` only runs if the deferred is still `isActive`. Added a regression test for a processor whose scope is already cancelled (`has(key) == false` after the last waiter leaves).

rahul-lohra: Nice

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -77,6 +114,70 @@ internal class CallJoinCoordinator(
         hintHighScaleLivestreamPublisher: Boolean? = null,
         callJoinInterceptor: CallJoinInterceptor? = null,
     ): Result<RtcSession> {
+        var coalesced = false
```
rahul-lohra, line 137: IMO, the first line should be `callAnalytics.joinAnalytics.onJoinFunctionStart()` this analytics indicate how many times this join api is invoked. 

So can we keep it that way?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -77,6 +114,70 @@ internal class CallJoinCoordinator(
...
+    ): Result<RtcSession> {
+        // Subsequent join() calls while a session is live return that session instead of
+        // building a second one. [joinInternal] repeats the same check for direct callers.
+        sessionManager.session.value?.let { existing ->
+            logger.w { "[join] Call already joined — returning existing session" }
+            existing.sfuTracer.trace("join-already-joined", "join() while session already live")
+            return Success(existing)
+        }
+
         callAnalytics.joinAnalytics.onJoinFunctionStart()
```
rahul-lohra: We should move this line to the beginning of the function; otherwise, it could result in incorrect analytics.

This analytics event should be recorded as soon as we enter the join function, regardless of any internal optimizations.

Let me know if you have any concern

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: modified it to support multiple join_initiated events but with same attempt-id

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: modified it to support multiple join_initiated events but with same attempt-id

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Done in `37789336db`.

`onJoinFunctionStart()` is now the first line of `join()`. Every public `Call.join()` — including coalesced waiters and already-joined no-ops — emits JoinInitiated with a **new** attempt id. We’ll tighten attempt-id correlation later.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Follow-up in `37789336db`.

We now emit `onJoinFunctionStart()` for every public `Call.join()`, including already-joined and coalesced waiters, each with a fresh attempt id. That restores field visibility of accidental / duplicate join() calls.

Coalesced interceptor mismatch still logs a warning; already-joined / coalesced still get the SFU traces. Attempt-id correlation (reuse the in-flight id) is follow-up.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Correction: we dropped same-attempt-id reuse. Each `join()` now mints a new attempt id (`37789336db`). Correlation cleanup later.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -537,20 +538,33 @@ public class RtcSession internal constructor(
 
         setMuteState(isEnabled = true, TrackType.TRACK_TYPE_AUDIO)
         val streamId = buildTrackId(TrackType.TRACK_TYPE_AUDIO)
-        val track = publisher.value?.publishStream(
+        val audio = publisher.value?.publishStream(
             streamId,
             TrackType.TRACK_TYPE_AUDIO,
-        )
+        ).asPublishedOrNull<org.webrtc.AudioTrack>(TrackType.TRACK_TYPE_AUDIO) ?: return
```
aleksandar-apostolov, line 543: Hasn't `setMuteState(isEnabled = true, ...)` already gone to the SFU by the time we return here? We'd look unmuted to everyone while sending nothing. Move it below the publish, or flip it back on this path? Same in the video and screen-share collectors.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/JoinAnalytics.kt
```diff
@@ -32,6 +32,9 @@ internal class JoinAnalytics(
 
     /**
      * Records the invocation of the public `Call.join()` API.
+     *
+     * Each invocation mints a new [JoinTelemetryState.joinStageAttemptId], including
```
aleksandar-apostolov, line 36: Non-blocker. Fresh id per invocation means N concurrent joins give N JoinInitiated but one CoordinatorJoinInitiated — reads as failures that never happened. Could we flag the coalesced and already-joined events so the funnel can filter them, until the correlation follow-up lands?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -278,6 +396,29 @@ internal class CallJoinCoordinator(
 
         state._connection.value = RealtimeConnection.Joined(localSession)
 
+        // Last-/sole-waiter cancel aborts this call-scoped job. If that happens after the
```
aleksandar-apostolov: Non-blocker. Still accurate? With `cancelIfLastWaiter = false` I think only scope cancellation on leave gets here now. Worth saying in the `join()` KDoc that callers can't abort a join anymore, only `leave()`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -90,6 +129,70 @@ internal class CallJoinCoordinator(
         callJoinInterceptor: CallJoinInterceptor? = null,
     ): Result<RtcSession> {
         callAnalytics.joinAnalytics.onJoinFunctionStart()
+        var coalesced = false
+        return joinFlight.run(
+            JOIN_FLIGHT_KEY,
+            onCoalesced = {
```
aleksandar-apostolov, line 141: Non-blocker. Is `state.callJoinInterceptor` set this early? The leader assigns it a few suspension points into `executeJoin`, so an early coalescer would warn about a drop that didn't happen. Capture the in-flight one when the flight is created and pass it in?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Good catch — `setMuteState(true)` already launches `UpdateMuteStates` (`muted = false`) before we get to `asPublishedOrNull`, so a null `publishStream` still looked live on the SFU with no track.

Moved unmute to after a successful publish in the audio, video, and screen-share paths so it matches the disable order (mute, then unpublish). Added a regression test that a null / missing publisher does not call `setMuteState`, and that a real track still unmutes after publish.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/JoinAnalytics.kt
PratimMallick: Its been taking care in another PR stacked on this where @rahul-lohra is sending new JoinInitiated per every join call with metadata suggesting if they are regular or concurrent ones

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Agreed — with `cancelIfLastWaiter = false` that catch is only call-scope cancellation (`leave()` / cleanup), not a waiter dropping.

Updated the `join()` KDocs (`Call` + coordinator) to say cancelling the caller does not abort the in-flight join, only `leave()` does, and rewrote the comment on the post-install `CancellationException` path.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Yes — `state.callJoinInterceptor` was assigned a few suspends into `executeJoin` (`guestUserJob.await()`, factory setup), so an early coalescer could warn about a drop against null/stale state.

The leader interceptor is now snapshotted in `onLeader` under the same lock as flight creation, and coalescers compare against that. `state.callJoinInterceptor` is also assigned immediately after the already-joined check, before any suspend.

Added a test that the interceptor is installed while the guest token is still pending, and that `onLeader` runs once before coalescers attach.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -65,21 +70,61 @@ internal class CallJoinCoordinator(
...
+     */
+    private val joinFlight = StreamRefCountedSingleFlightProcessor(scope)
+
+    /**
+     * Interceptor from the waiter that created the current join flight. Snapshotted in
+     * [onLeader] under the flight mutex so coalescers can compare against it before
+     * [executeJoin] assigns [CallState.callJoinInterceptor].
+     */
+    @Volatile
+    private var inFlightJoinInterceptor: CallJoinInterceptor? = null
```
rahul-lohra: [P2] - I think this property shouldn't be here. Instead, it should be scoped to Flight.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -90,6 +135,75 @@ internal class CallJoinCoordinator(
...
+        // building a second one. [joinInternal] repeats the same check for direct callers.
+        sessionManager.session.value?.let { existing ->
+            logger.w { "[join] Call already joined — returning existing session" }
+            existing.sfuTracer.trace("join-already-joined", "join() while session already live")
+            return Success(existing)
+        }
+
+        // Before any suspend so coalesced waiters and SFU observers see this join's
+        // interceptor rather than a stale/null [CallState.callJoinInterceptor].
+        state.callJoinInterceptor = callJoinInterceptor
```
rahul-lohra: I think we will receive stale `callJoinInterceptor` here
Consider the case when the Caller A's Activity is destroyed and we move to activity B with call.join(interceptor) again
Then in that case a dead activity's interceptor will be invoked. 

Please think around this. What behaviour should we follow?

PratimMallick: yes this was intentional, are taking the first interceptor that was sent and knowingly discarding the interceptor of next join calls, This is the same pattern used for other join parameters

rahul-lohra: I have listed 4 concerns below. Please check
 
### 1 `CallJoinInterceptor` is lifecycle-sensitive

`CallJoinInterceptor` is different from the other join parameters. The other parameters are request configuration, while the interceptor is integrator-provided executable code. Its implementation may reference an Activity, UI components, navigation, or lifecycle-bound state.
Consider this scenario:

> 1. Activity A calls `join(interceptorA)`.
> 2. Activity A is destroyed and recreated as Activity B.
> 3. Activity B calls `join(interceptorB)`.

With the current implementation, the second call coalesces into the existing flight. `interceptorA` remains selected and `interceptorB` is discarded. The SDK may later invoke code associated with the destroyed Activity A. _**This could update stale UI, navigate using the old Activity, throw an exception, or skip the intended user interaction**_.

### 2 This changes integrator-visible behaviour
Before join coalescing, concurrent join attempts could each assign `CallState.callJoinInterceptor`, so a later execution could replace the previous interceptor. With single-flight, only the first join block executes; subsequent interceptors are silently discarded.
This removes the previous race, but it also establishes first-interceptor-wins as new deterministic behaviour.
Since this PR explicitly resolves concurrent join() behaviour, interceptor ownership still seems like a loose end. 

### 3 Interceptor ownership trade-offs

Since `CallJoinInterceptor` is integrator-provided executable code. The main risk is invoking an interceptor supplied by a caller whose coroutine has already been cancelled. The interceptor itself runs in the SDK scope, so caller cancellation does not automatically stop it.
I have listed down Interceptor ownership trade-offs
| Policy | Advantages | Disadvantages | Risk |
| -- | -- | -- | -- |
| First interceptor wins — current | Simple; consistent with other first-flight arguments | Caller A’s interceptor remains selected even after A’s Job is cancelled. Caller B’s interceptor is ignored. | High: SDK can invoke code belonging to a cancelled caller. That code may use cancelled work, stale state, or throw CancellationException, potentially preventing the active-state transition. |
| Latest non-null wins | Simple replacement rule; join(null) does not erase an interceptor | The latest interceptor may also be cancelled before invocation, and the previous interceptor is no longer available as fallback. | Medium: Reduces stale-first-caller risk but does not verify whether the latest caller is still valid. |
| First non-cancelled wins | Preserves first-wins when the original caller remains valid; falls back to the next candidate after cancellation | Must store interceptor candidates with their caller Jobs and select them atomically. | Low: A cancellation can still race with selection. Once invocation starts, the selected interceptor must be frozen and allowed to finish. |
| Latest non-cancelled wins | Uses the most recent eligible interceptor and retains older candidates as fallback | Changes the existing first-wins policy and requires synchronized ordering and selection. | Low–Medium: Avoids known cancelled callers, but concurrent arrival order determines which valid interceptor wins.|

> **My Recommendation**: **_first non-cancelled wins_**. It preserves the current policy unless the original caller is cancelled, while avoiding invocation of an interceptor already known to be associated with a cancelled caller.

 ### 4. Concurrent interceptor behaviour is not documented

 `Call.join()` KDoc states that concurrent callers share one in-flight attempt, but it does not explain what happens when those callers provide different arguments or different `CallJoinInterceptor` instances.

 `CallJoinInterceptor` KDoc also does not state that the first interceptor is retained and later interceptors are ignored. Therefore, an integrator cannot predict this behaviour from the public API contract.

Once the ownership policy is agreed, it should be documented explicitly. For example, if first-wins remains:

 Concurrent `join()` calls share the same in-flight operation. The arguments and `CallJoinInterceptor` from the caller that creates the operation are used. Arguments supplied by coalesced callers are ignored.

 If latest-non-null wins, that replacement behaviour should be documented instead.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Interceptor candidates now live on the flight: each waiter registers `(Job, interceptor)` under the same lock as attach. `firstNonCancelledAttachment` reads that list. The coordinator field is gone.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
PratimMallick: Policy now:
- Each waiter registers its interceptor on the flight with that waiter’s `Job`.
- Selection is the first non-null interceptor whose waiter job is **not cancelled** (completed successfully still counts; we do not treat `join()` returning as cancellation).
- `join(null)` does not erase an earlier interceptor.
- `ActiveStateGate` resolves the interceptor at `callReadyToJoin` start and freezes that instance for the invocation.
- After the flight is removed, `CallState.callJoinInterceptor` holds the last live selection.

Documented on `Call.join()` and `CallJoinInterceptor`. Added tests for cancelled-first → second wins, active-first keeps its interceptor, and provider resolution at invoke time.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -812,6 +812,14 @@ public class CallState(
     @Volatile
     internal var callJoinInterceptor: CallJoinInterceptor? = null
 
+    /**
+     * Live interceptor selection while a join flight is in progress. Falls back to
+     * [callJoinInterceptor] after the flight is removed. Resolved at interceptor
+     * invocation, not at ringing-Active entry.
+     */
+    @Volatile
+    internal var callJoinInterceptorProvider: (() -> CallJoinInterceptor?)? = null
```
rahul-lohra, line 821: [P2] We have 2 objects related to `callJoinInterceptor`. The `callJoinInterceptorProvider` should be enough. We can remove the older one which is: `internal var callJoinInterceptor: CallJoinInterceptor? = null`. 

This is not a blocker and does not affect the core responsibility of this PR

PR #59, opened 2022-11-03
Task/documentation prep

filbabic:
Added WIP tutorial

PR #1767, opened 2026-08-14
Update build conventions to 0.14.0

gpunto:
### Goal

Bump `streamConventions` to 0.14.0, which renames `coverage.sonarCoverageExclusions` to `coverage.sonarExclusions` and wires it to `sonar.exclusions` instead of `sonar.coverage.exclusions`. Part of [AND-1362](https://linear.app/stream/issue/AND-1362/broaden-conventions-sonarcoverageexclusions-to-a-full-sonarexclusions).

### Implementation

- Bump `streamConventions` 0.13.1 → 0.14.0.
- Rename the `streamProject.coverage` property to `sonarExclusions`.
- Re-record the `CallLobbyTest` Paparazzi snapshot (see below).

The listed paths are now excluded from all Sonar analysis (coverage, duplication, issues), not just coverage. That is the intent for the binaries and generated code; it also drops `core/model` from issue analysis, which is acceptable since those are plain model classes already excluded from coverage.

### Paparazzi verification was silently skipped

0.14.0 also contains conventions #59, which makes the Paparazzi plugin check lazy. Before, `pluginManager.hasPlugin("app.cash.paparazzi")` was evaluated before the module applied the plugin, so it was always false and `testCoverage` only depended on `testDebugUnitTest`. Now `withPlugin` wires `verifyPaparazziDebug` in properly, and since CI runs `./gradlew :testCoverage`, snapshot verification actually runs for the first time.

It immediately caught one stale snapshot: `CallLobbyTest` differed by 23.26%. The committed image predates #1721, which made `CallLobby` overlay `DefaultMediaTrackFallbackContent` until the first frame renders. Under Paparazzi no frame ever renders, so the loading spinner permanently covers the inspection-mode avatar placeholder the old image captured. The new snapshot is the correct output of current code, not a regression from this bump: re-recording on 0.13.1 produces a byte-identical image.

### Testing

- `recordPaparazziDebug` on 0.13.1 and 0.14.0 produce identical output, confirming the snapshot drift is independent of this bump.
- `verifyPaparazziDebug` passes with the re-recorded snapshot; 83 tests, only the stale one was failing before.
- `testCoverage --dry-run` shows `verifyPaparazziDebug` present on 0.14.0 and absent on 0.13.1.

#### General
- [ ] Assigned a person / code owner group
- [ ] Thread with the PR link started in a respective Slack channel

#### Code & documentation
- [x] No client-facing changes


stream-video-android-ui-compose/src/test/snapshots/images/io.getstream.video.android.compose_CallLobbyTest_snapshot CallLobby composable.png
gpunto, line 1: Re-recorded, not a regression from this bump.

The light avatar placeholder in the old image is the `stream_video_call_sample` vector, drawn by `ParticipantVideoRenderer`'s `LocalInspectionMode` branch. The new image is `DefaultMediaTrackFallbackContent`: the `baseSheetTertiary` background plus a `CircularProgressIndicator` frozen at animation frame 0, which is the faint speck in the centre.

Cause is #1721, which made `CallLobby` pass `fallbackContent = {}` and overlay the loading content while `!isVideoRendered`. Under Paparazzi no frame ever renders, so `onRendered` never fires and the spinner permanently covers the placeholder. The old image was last recorded in #1534, well before that.

It surfaced here because conventions 0.14.0 makes the Paparazzi plugin check lazy, so `verifyPaparazziDebug` is finally wired into `:testCoverage` (previously `hasPlugin` ran before the module applied the plugin, so it was always false and verification was silently skipped). Re-recording on 0.13.1 yields a byte-identical image, so the drift predates this PR.

PR #1768, opened 2026-08-14
Avoid leaking pending broadcast in GenericCallActionBroadcastReceiver when call id is missing

rahul-lohra:
### Goal
Closes #AND-1390
`GenericCallActionBroadcastReceiver` called `goAsync()` unconditionally, but only the `streamCallId != null` path ever finished the pending result. When a broadcast arrived without a call id, the receiver logged a warning and returned — leaving the `BroadcastReceiver.PendingResult` open. Android then kept the broadcast (and its process priority) alive until the ~10s ANR timeout instead of completing immediately.

This PR ensures the pending result is only created when there is a coroutine that will finish it, eliminating the leak.

### Implementation

Fix — in `GenericCallActionBroadcastReceiver.onReceive`, moved goAsync() inside the `streamCallId != null` branch:

Before: goAsync() ran before the null check; the else branch only logged and returned, never calling pendingResult.finish().
After: the pending result is created only in the branch whose launched coroutine finishes it. The missing-call-id branch just logs — nothing to leak.

### 🎨 UI Changes

None

### Testing

Smoke test the app

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of call action broadcasts when calls are cancelled or encounter errors.
  - Ensured asynchronous broadcast operations always complete reliably.
  - Prevented unnecessary background processing when call details or actions are invalid.
- **Tests**
  - Added coverage for valid, invalid, cancelled, and uninitialized call scenarios.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/receivers/GenericCallActionBroadcastReceiver.kt
PratimMallick, line 75: Should we handle cancellation Exception? Will this scope get cancelled? If we handle cancellationException then we should keep `pendingResult.finish()` in `finally` block so that it runs when scope is cancelled

PratimMallick: but overall ok

PratimMallick (approved):
looks fine .. except for a minor question about the cancellation Exception

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/receivers/GenericCallActionBroadcastReceiver.kt
rahul-lohra: Code is updated now. Thanks for pointing it out

PR #1770, opened 2026-08-17
Signal noise cancellation state to the SFU

aleksandar-apostolov:
### Goal

Fixes [AND-1399](https://linear.app/stream/issue/AND-1399/noise-cancellation-state-is-never-signalled-to-the-sfu)

`Call.setAudioProcessingEnabled` / `toggleAudioProcessing` only flipped the local WebRTC audio-processing module. The SFU `StartNoiseCancellation` / `StopNoiseCancellation` RPCs are generated from `signal.proto` and wrapped by both signalling decorators (trace + retry), but had no caller — so the server received no noise-cancellation signal from Android sessions. JS (`Call.notifyNoiseCancellationStarting`) and Swift (`callController.startNoiseCancellation`) both send it.

First of a three-part stack. AND-1400 and AND-1401 follow and touch the same call path.

### Implementation

- `RtcSession.startNoiseCancellation()` / `stopNoiseCancellation()`, following the existing `updateMuteState` shape — `safeApiCall` around the signalling service, request carries `session_id`. Both `internal`.
- `Call.setAudioProcessingEnabled` / `toggleAudioProcessing` notify the SFU after applying the local change. The signal follows the state **actually applied**, not the requested one, so a no-op toggle (no `ManagedAudioProcessingFactory` supplied) reports the resolved value.
- Fire-and-forget on the call scope; failures logged at warn, never surfaced. No-ops when there is no active session.
- No public API change — `apiCheck` passes with the API dump byte-unchanged.

### 🎨 UI Changes

None — no UI surface is affected.

### Testing

```
./gradlew :stream-video-android-core:testDebugUnitTest   # 997 tests, 0 failures, 50 pre-existing skips
./gradlew :stream-video-android-core:apiCheck            # passes, api dump unchanged
./gradlew :stream-video-android-core:spotlessCheck
```

New coverage:

- `NoiseCancellationSignalTest` — 7 cases at the `Call` level: enable signals start, disable signals stop, a request that cannot be applied locally signals stop rather than start, toggle signals the resolved state, state changed before a session exists is sent once one is installed, a replacement session is told the current state, and rapid changes leave the SFU holding the final state.
- `RtcSessionTest2` — 2 cases asserting each RPC reaches the signalling service with the correct `session_id`.

## Summary by CodeRabbit

* **New Features**
  * Noise-cancellation controls now synchronize their enabled or disabled state with the SFU.
  * Toggling audio processing reports the resulting noise-cancellation state.
  * State updates safely run in the background and do not interrupt audio controls if signaling fails.

* **Bug Fixes**
  * Improved behavior when no active call session is available.

* **Tests**
  * Added coverage for noise-cancellation signaling, toggling, and session handling.


PratimMallick:
## Review notes

Overall the RPC wiring looks good — `RtcSession` mirrors `updateMuteState`, and the second commit’s `SerialProcessor` + `@Volatile desired…` correctly addresses out-of-order / stale-session races. A couple of correctness gaps worth addressing (or explicitly deferring to the stack):

### 1. High — NC not re-synced when a session appears

`notifyNoiseCancellationState` only runs from `setAudioProcessingEnabled` / `toggleAudioProcessing`. If `session` is null at that moment, the RPC is skipped and nothing re-sends when a session later appears.

That hits the real product path today: the demo lobby calls `setAudioProcessingEnabled(isAutoOn)` **before** join. Same for REJOIN/MIGRATE — local NC can stay on while the new SFU session never gets `StartNoiseCancellation`.

`desiredNoiseCancellationEnabled` is remembered, but nothing observes `session` to flush it. Even with AND-1401’s auto-on in core, settings can resolve before the session is up and hit the same hole.

**Suggestion:** on session install (join / rejoin / migrate), if the desired/local NC state is enabled, call `notifyNoiseCancellationState` again — or have `notify` enqueue and drain when `session` becomes non-null.

### 2. Medium — `setAudioProcessingEnabled` signals requested state, not applied

```kotlin
fun setAudioProcessingEnabled(enabled: Boolean) {
    media.setAudioProcessingEnabled(enabled)
    notifyNoiseCancellationState(enabled) // requested
}
```

`toggleAudioProcessing` correctly notifies with the **resolved** value. `set` does not. When no `ManagedAudioProcessingFactory` is configured, the local set is a no-op but the SFU can still receive `startNoiseCancellation`.

**Suggestion:** notify with `media.isAudioProcessingEnabled()` after the set (or only notify when applied state actually changed). This also stays consistent with AND-1400’s MUSIC_HIGH_QUALITY no-op path.

### Test nits

- Rapid-change test asserts `stop` was called, not that it was the **last** call (`coVerifyOrder` / last-invocation would be stronger).
- No coverage for session replacement mid-queue (called out in the serialize commit).
- PR body still says “4 cases”; the second commit adds a 5th.

### Stack note

Fine to treat this PR as “RPC plumbing when a session already exists,” but worth confirming AND-1399 isn’t considered fully closed for lobby auto-on / rejoin until the session-flush lands here or in #1771 / #1773.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -869,9 +885,48 @@ public class Call(
...
+     * Signals run one at a time and each sends the latest requested state, so rapid toggles
+     * converge on the right value instead of racing. The session is resolved as the signal runs,
+     * not when it was requested — a rejoin in between must not send the request at a session that
+     * has already been replaced.
+     */
+    private fun notifyNoiseCancellationState(enabled: Boolean) {
+        desiredNoiseCancellationEnabled = enabled
+        scope.launch {
+            noiseCancellationSignals.submit("noiseCancellation") {
+                val session = session.value
```
PratimMallick, line 922: **High — NC not re-synced when a session appears**

If `session` is null here, the RPC is skipped and nothing re-sends when a session later appears.

This still bites once the stack lands: #1773 moves auto-on into core via `observeNoiseCancellationPolicy` → `setAudioProcessingEnabled(true)` when settings resolve, which is often **before** join. Same for REJOIN/MIGRATE — local NC can stay on while the new SFU session never gets `StartNoiseCancellation`.

`desiredNoiseCancellationEnabled` is remembered, but nothing observes `session` to flush it.

**Suggestion:** on session install (join / rejoin / migrate), if desired/local NC is enabled, call `notifyNoiseCancellationState` again — or drain when `session` becomes non-null.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -869,9 +885,48 @@ public class Call(
 
     fun isAudioProcessingEnabled(): Boolean = media.isAudioProcessingEnabled()
 
-    fun setAudioProcessingEnabled(enabled: Boolean) = media.setAudioProcessingEnabled(enabled)
+    fun setAudioProcessingEnabled(enabled: Boolean) {
+        media.setAudioProcessingEnabled(enabled)
+        notifyNoiseCancellationState(enabled)
```
PratimMallick: **Medium — signals requested state, not applied**

`toggleAudioProcessing` correctly notifies with the **resolved** value. `set` passes the requested `enabled` flag.

When no `ManagedAudioProcessingFactory` is configured, the local set is a no-op but the SFU can still receive `startNoiseCancellation`. This gets worse in #1771 (MUSIC_HIGH_QUALITY no-op) and #1773 (deferred factory apply).

**Suggestion:** notify with the applied state after the set (or only notify when applied state actually changed).

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        call.injectSession(session)
+
+        // Signals are queued and each sends whatever was requested most recently, so whichever
+        // order the queued coroutines start in, the last one carries "off". Intermediate signals
+        // are legitimate — the state really was on for a moment — so this asserts convergence,
+        // not suppression.
+        call.setAudioProcessingEnabled(true)
+        call.setAudioProcessingEnabled(false)
+
+        coVerify(timeout = SIGNAL_TIMEOUT_MS) { session.stopNoiseCancellation() }
```
PratimMallick, line 172: **Test nit:** this asserts `stop` was called, not that it was the **last** call. `coVerifyOrder` / last-invocation would be stronger for the convergence claim.

Also: no coverage yet for session replacement mid-queue (called out in the serialize commit), and the PR body still says “4 cases” while this is a 5th.

PratimMallick (commented):
Moved the earlier conversation comment onto the relevant lines. Overall the RPC wiring and serialization look good — the notes below are the gaps worth addressing (or explicitly deferring in the stack).

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Addressed — `observeNoiseCancellationSignalTarget` now flushes the **applied** state when a session is installed (including replacements). The follow-up to read `isAudioProcessingEnabledIfCreated()` instead of replaying the last signal is the right call for the deferred-factory case in #1771/#1773.

Join still builds the factory in `RtcSession` init **before** `setActiveSession`, so the observer sees the applied value. Looks good.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Addressed — `setAudioProcessingEnabled` now notifies `media.isAudioProcessingEnabledIfCreated()`, matching toggle. The new test that a request with no processor signals stop (not start) covers it.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
PratimMallick: Addressed — the rapid-change test now asserts `signalled.last()`, and there is coverage for pre-join install plus replacement session. PR body is updated to 7 cases.

PratimMallick (commented):
Re-reviewed after the follow-up commits. Previous findings are addressed:

- Session flush via `observeNoiseCancellationSignalTarget` (applied state, including replacements)
- `set` notifies applied-if-created, not the request
- Tests cover pre-join, replacement, last-signal convergence, and no-processor → stop

No remaining blockers on this PR. Join still constructs the factory in `RtcSession` init before `setActiveSession`, so the observer sees the applied value.

PratimMallick (approved):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -869,9 +869,37 @@ public class Call(
...
+                session.stopNoiseCancellation()
+            }
+            if (result is Result.Failure) {
+                logger.w {
+                    "[notifyNoiseCancellationState] #sfu; enabled: $enabled, " +
+                        "failed: ${result.value.message}"
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
aleksandar-apostolov: Already addressed — this was fixed on the same day it was posted, and the thread was just never resolved.

Signals go through a `SerialProcessor` so they run one at a time, and the session is resolved *inside* the serialized task rather than captured before it, so a rejoin in between cannot send the request at a replaced session. Each queued signal sends the most recently requested state, so rapid changes converge instead of racing.

Both requested cases are covered: `rapid changes leave the SFU holding the final state` (asserts the last state the SFU is left holding, not just that a stop happened) and `a replacement session is told the current state`.

PR #1771, opened 2026-08-17
Noise cancellation state no longer leaks between calls or misreports

aleksandar-apostolov:
Stacked on #1770. Merge after parent.

### Goal

Fixes [AND-1400](https://linear.app/stream/issue/AND-1400/noise-cancellation-state-leaks-across-calls-and)

The audio processor is owned by the client and shared by every call. Two defects followed from that:

- **Misreport.** Under the `MUSIC_HIGH_QUALITY` audio bitrate profile the processor is deliberately never attached to the native factory, but `isAudioProcessingEnabled()` still read the shared module's own `isEnabled` flag — so it returned `true` while nothing was processing, and UI bound to it highlighted an inactive feature.
- **Cross-call leak.** The flag was never cleared on leave, so enabling noise cancellation in one call silently enabled it in the next, regardless of that call's type or settings.

JS scopes this per-call in `MicrophoneManager`; Swift per-call via `setAudioFilter`.

Second of a three-part stack. AND-1401 follows.

### Implementation

- `StreamPeerConnectionFactory` tracks whether `audioProcessing` was actually wired into the native factory. `isAudioProcessingEnabled()` reports that; `setAudioProcessingEnabled` / `toggleAudioProcessing` no-op rather than mutating a module this call cannot use.
- Before the factory is built the attachment check falls back to what it *will* be built with, so the pre-join lobby path (`setAudioProcessingEnabled(isAutoOn)`) keeps working.
- `resetAudioProcessing()` clears the shared flag; `CallMediaManager.cleanup()` calls it on leave, reading the backing field so cleanup never lazily builds a factory just to reset one.
- No public API change — `apiCheck` passes with the API dump unchanged.

### 🎨 UI Changes

None in the SDK. Fixes stale highlighting in UI that binds to `isAudioProcessingEnabled()` — the demo app's settings-menu item is one such consumer.

### Testing

```
./gradlew :stream-video-android-core:testDebugUnitTest   # 1000 tests, 0 failures, 50 pre-existing skips
./gradlew :stream-video-android-core:apiCheck            # passes, api dump unchanged
./gradlew :stream-video-android-core:spotlessApply
```

New coverage — `StreamPeerConnectionFactoryTest` +4, `CallMediaManagerTest` +2:

- Getter returns false under `MUSIC_HIGH_QUALITY` even with the module flagged on; setter and toggle leave the shared module untouched there.
- `resetAudioProcessing` clears the shared module.
- `cleanup()` resets it, and does not build a factory when the call never used one.

The three guard tests were verified red with the fix reverted, so they aren't vacuous.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio-processing state reporting for calls where processing is unavailable.
  * Prevented noise-cancellation settings from leaking between calls.
  * Ensured call cleanup resets audio processing without creating unnecessary call resources.
  * Disabled audio processing for high-quality music calls, preventing unsupported state changes or reporting.

* **Documentation**
  * Updated the Android SDK changelog with the latest audio-processing and call behavior fixes.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
@@ -286,8 +286,17 @@ internal class CallMediaManager(
         mediaManager.screenShare.disable(fromUser = true)
     }
 
-    fun isAudioProcessingEnabled(): Boolean {
-        return peerConnectionFactory.isAudioProcessingEnabled()
+    /**
+     * Whether noise cancellation is on for this call: what is actually processing once a factory
+     * exists, and what the call asked for before then.
+     *
+     * Never builds a factory. Once one exists its answer wins, so a call under MUSIC_HIGH_QUALITY
+     * reports off even though it asked for on — nothing is attached to process its audio.
+     */
+    fun isAudioProcessingEnabled(): Boolean = if (_peerConnectionFactory != null) {
+        isAudioProcessingEnabledIfCreated()
+    } else {
+        desiredAudioProcessingEnabled
     }
 
     /**
@@ -312,10 +321,17 @@ internal class CallMediaManager(
         _peerConnectionFactory?.setAudioProcessingEnabled(enabled)
     }
 
+    /**
+     * Flips the wanted audio-processing state, applying it if a factory exists.
+     *
+     * Goes through [setAudioProcessingEnabled] rather than the factory so a toggle before joining
+     * cannot build one: a factory created there captures the pre-join audio bitrate profile, and
+     * one holding the shared processor is kept rather than released, so the profile would be
+     * pinned for the rest of the call.
+     */
     fun toggleAudioProcessing(): Boolean {
-        return peerConnectionFactory.toggleAudioProcessing().also { enabled ->
-            desiredAudioProcessingEnabled = enabled
-        }
+        setAudioProcessingEnabled(!desiredAudioProcessingEnabled)
+        return isAudioProcessingEnabled()
     }
 
     /** Disables all local capture devices. Used when leaving the call. */
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
@@ -296,6 +296,10 @@ internal class CallMediaManager(
     }
 
     fun cleanup() {
+        // The audio processor is owned by the client and shared by every call, so this call's
+        // choice has to be cleared here or it carries into the next one. Reads the backing
+        // field directly: cleanup must never lazily build a factory just to reset it.
+        _peerConnectionFactory?.resetAudioProcessing()
         mediaManager.cleanup()
     }
 }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
@@ -74,6 +74,17 @@ internal class CallMediaManager(
     // peerConnectionFactory is nullable and recreated when audioBitrateProfile changes (before joining)
     private var _peerConnectionFactory: StreamPeerConnectionFactory? = null
 
+    /**
+     * Audio-processing state this call wants, remembered separately from the factory so it
+     * survives one not existing yet or being recreated.
+     *
+     * The processor itself is owned by the client and shared by every call, and its enabled flag
+     * lives on that one instance. Rather than clearing the flag when a call ends — which reaches
+     * across into whatever call runs next — every factory applies the state its own call asked
+     * for as it is built. Defaults to off so a call that never asks does not inherit.
+     */
+    private var desiredAudioProcessingEnabled: Boolean = false
+
     var peerConnectionFactory: StreamPeerConnectionFactory
         get() {
             if (_peerConnectionFactory == null) {
@@ -85,7 +96,9 @@ internal class CallMediaManager(
                     audioBitrateProfileProvider = { mediaManager.microphone.audioBitrateProfile.value },
                     sharedEglBaseProvider = { eglBase() },
                     webRtcLoggingLevel = clientImpl.loggingLevel.webRtcLoggingLevel,
-                )
+                ).also { factory ->
+                    factory.setAudioProcessingEnabled(desiredAudioProcessingEnabled)
+                }
             }
             return _peerConnectionFactory!!
         }
@@ -171,9 +184,17 @@ internal class CallMediaManager(
      * This should only be called before the call is joined.
      */
     fun recreatePeerConnectionFactory() {
-        _peerConnectionFactory?.dispose()
-        _peerConnectionFactory = null
+        val previous = _peerConnectionFactory
         // Next access to peerConnectionFactory will recreate it with current profile
+        _peerConnectionFactory = null
+        if (previous?.hasAudioProcessingAttached() == true) {
+            // The shared audio processor is torn down when the native factory holding it is
+            // released, which would leave the client without one for every later call. Dropping
+            // the factory instead leaks it, the same way a factory is dropped on call cleanup.
+            logger.i { "Keeping the previous factory alive: it holds the shared audio processor" }
+            return
+        }
+        previous?.dispose()
     }
 
     /** Applies server-provided call settings to the local media manager. */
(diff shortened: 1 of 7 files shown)


PratimMallick:
## Review notes

Solid, focused fix for the two AND-1400 defects. Attachment tracking + pre-factory prediction look correct, and cleanup via the backing field is the right way to avoid lazily building a factory just to reset it. Tests covering MUSIC_HIGH_QUALITY guards and cleanup are in good shape.

### Medium — `set` still signals requested state to the SFU (interaction with #1770)

With the new attachment gating, `StreamPeerConnectionFactory.setAudioProcessingEnabled` is a **no-op** under `MUSIC_HIGH_QUALITY`, but `Call.setAudioProcessingEnabled` (from parent #1770) still calls:

```kotlin
notifyNoiseCancellationState(enabled) // requested flag
```

`toggleAudioProcessing` already notifies the resolved applied value; `set` does not. On a Hi-Fi call the SFU can receive `startNoiseCancellation` while local processing is unattached and `isAudioProcessingEnabled()` reports `false`.

**Suggestion:** have `Call.setAudioProcessingEnabled` notify the applied state (e.g. `media.isAudioProcessingEnabled()` / if-created), ideally in #1770 or as a small follow-up in this stack before merge.

No other material findings on this PR in isolation — happy to approve once the signaling mismatch is acknowledged or fixed in the stack.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -680,28 +704,48 @@ public class StreamPeerConnectionFactory(
...
     }
 
     /**
      * Sets the audio processing on or off.
+     *
+     * No-op when the processor is not wired into the native factory: the module is shared
+     * across calls, and a call that cannot use it must not change what other calls observe.
      */
     public fun setAudioProcessingEnabled(enabled: Boolean) {
+        if (!isAudioProcessingAttached()) return
```
PratimMallick, line 740: **Medium — interaction with #1770 SFU notify**

This no-op under `MUSIC_HIGH_QUALITY` is correct for the shared module. But `Call.setAudioProcessingEnabled` (from parent #1770) still does `notifyNoiseCancellationState(enabled)` with the **requested** flag after calling into here.

`toggleAudioProcessing` already notifies the resolved applied value; `set` does not. On a Hi-Fi call the SFU can receive `startNoiseCancellation` while local processing is unattached and `isAudioProcessingEnabled()` reports `false`.

**Suggestion:** have `Call.setAudioProcessingEnabled` notify the applied state (ideally in #1770, or as a small follow-up in this stack before merge).

PratimMallick (commented):
Moved the earlier conversation comment onto the relevant lines. Solid, focused fix for the AND-1400 defects — one signaling interaction with #1770 worth fixing in the stack.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
```diff
@@ -285,6 +285,10 @@ internal class CallMediaManager(
     }
 
     fun cleanup() {
+        // The audio processor is owned by the client and shared by every call, so this call's
```
PratimMallick: Nice — reading the backing field so cleanup never lazily builds a factory just to reset it. This is the right fix for the cross-call leak.

aleksandar-apostolov: This code is gone now, so the approval landed on something that no longer exists.

Clearing the shared flag on cleanup is itself the cross-call bug: a call's cleanup runs behind stats collection and the leave event, so it lands after the next call has started and turns *that* call's noise cancellation off. The mechanism you liked — reading the backing field so cleanup never builds a factory — was right; what it was doing was not.

Instead, each call remembers the state it asked for and applies it as its own factory is built, defaulting to off. Nothing writes the shared flag on the way out, so a call can neither inherit a previous choice nor overwrite a later one.

Also stopped releasing a factory that was built with the shared processor — releasing it tears the processor down for the whole process, leaving every later call without noise cancellation. Reachable by changing the audio bitrate profile before joining.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
PratimMallick: Addressed in parent #1770 — `Call.setAudioProcessingEnabled` now signals the applied state, so a MUSIC_HIGH_QUALITY no-op will not send `startNoiseCancellation`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
PratimMallick: Agreed — clearing the shared flag on cleanup was the wrong tool. Remembering per-call desired state and applying it on factory create is the right model. Thanks for the correction.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
```diff
@@ -171,9 +184,17 @@ internal class CallMediaManager(
      * This should only be called before the call is joined.
      */
     fun recreatePeerConnectionFactory() {
-        _peerConnectionFactory?.dispose()
-        _peerConnectionFactory = null
+        val previous = _peerConnectionFactory
         // Next access to peerConnectionFactory will recreate it with current profile
+        _peerConnectionFactory = null
+        if (previous?.hasAudioProcessingAttached() == true) {
```
PratimMallick, line 190: **Medium — leaking the old factory, then attaching the same processor to a new one**

Understood: disposing a native factory that called `setAudioProcessingFactory` tears down the client-owned processor. Dropping instead of `dispose()` avoids that.

The next getter still builds a **second** factory and (for non-MUSIC_HIGH_QUALITY) calls `setAudioProcessingFactory` on the **same** `ManagedAudioProcessingFactory`. Two native factories then claim one processor; `createNative` may run twice.

This path only runs when a factory already exists before join *and* the audio bitrate profile changed (`ensureFactoryMatchesAudioProfile`). After this stack, `set` no longer builds a factory, so the common auto-on path is fine. It still fires if anything pre-join hits the lazy getter (`isAudioProcessingEnabled()` / `toggleAudioProcessing()`) and the profile then changes.

Worth confirming Krisp/WebRTC is safe with two live factories sharing one processor — or skipping the second `setAudioProcessingFactory` when the leaked factory already owns it.

PratimMallick (commented):
Re-reviewed. The previous SFU requested-vs-applied issue is fixed in #1770. Per-call desired state + apply-on-factory-create is the right replacement for resetting the shared flag on cleanup.

One remaining note on leaking a factory that still holds the shared processor (see inline).

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
PratimMallick: Not blocking. The common join path no longer builds a factory before `ensureFactoryMatchesAudioProfile`, so this only hits if something else (toggle / `isAudioProcessingEnabled()`) created one in the lobby and the profile then changed. Fine to land as a known tradeoff; worth a follow-up if Krisp misbehaves with two native factories sharing the processor.

PratimMallick (approved):

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/StreamPeerConnectionFactory.kt
```diff
@@ -680,28 +704,48 @@ public class StreamPeerConnectionFactory(
...
 
+    /**
+     * Clears the shared audio processor's enabled state.
+     *
+     * The [ManagedAudioProcessingFactory] is owned by the client and reused by every call, so a
+     * call must not leave its choice behind for whichever call comes next.
+     */
+    internal fun resetAudioProcessing() {
+        audioProcessing?.isEnabled = false
+    }
```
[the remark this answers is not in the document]
aleksandar-apostolov: Finding was right, fix went a different way.

`resetAudioProcessing()` is gone rather than moved. Clearing the shared flag on the way out is the problem itself, and it does not need two concurrent calls to bite: a call's cleanup runs behind stats collection and the leave event, so it lands after the next call has already started and turns *that* call's noise cancellation off.

Instead, nothing writes the shared flag when a call ends. Each call records the state it asked for and applies it as its own peer-connection factory is built, defaulting to off — so a call can neither inherit a previous choice nor overwrite a later one, and there is no reset to own.

No client-level refcounting, and no two-call regression test: the processor's enabled state is a single flag shared by the whole client, so two calls running at once cannot hold different values however the reset is scoped. A test asserting otherwise would be asserting something the design cannot provide. What the tests cover instead is that a leaving call cannot affect a later one, and that a factory built later applies its own call's state.

Related, from the same area: releasing a factory built with the shared processor tears that processor down for the rest of the process, so such a factory is now kept rather than released.

PR #1773, opened 2026-08-18
Enforce the noise-cancellation capability and call settings in core

aleksandar-apostolov:
Stacked on #1771. Merge after parent.

### Goal

Fixes [AND-1401](https://linear.app/stream/issue/AND-1401/noise-cancellation-capability-and-call-settings-are-not-enforced-in)

Core checked neither `OwnCapability.EnableNoiseCancellation` nor the call type's `noise_cancellation` mode, so noise cancellation ran regardless of what the server granted. Android enforced neither check — and shipped the mode helpers (`NoiseCancellationUtils`) in core while only the demo app consumed them.

Last of a three-part stack (#1770, #1771).

### Implementation

- `NoiseCancellationPolicy` owns both checks. `Call` reconciles against it: enabling is refused when either is withheld, and withdrawn if either is taken away mid-call. Disabling is always permitted so a withdrawal can never be blocked.
- The reconciler waits for settings to **resolve** rather than reading "not told yet" as a refusal — the latter switches noise cancellation off and back on during join.
- The call type's `auto-on` default moves out of the demo app's lobby view model into core, so every app gets it instead of whoever remembered to wire it.
- Applying a wanted state no longer forces the peer-connection factory into existence — `CallMediaManager` remembers it and a factory created later picks it up on construction. This matters beyond tidiness: a factory built before join captures the pre-join audio bitrate profile and defeats `ensureFactoryMatchesAudioProfile`.
- **New public API:** `CallState.audioProcessingEnabled: StateFlow<Boolean>`, so consumers can observe SDK-driven changes. Additive only — the API dump gains exactly one getter, nothing removed or altered. Version constant untouched.
- demo-app collects that flow instead of snapshotting `isAudioProcessingEnabled()`, and now reads settings only to decide whether to show the menu item.

Behaviour note for review: an app running on a call type where the capability is not granted loses noise cancellation, silently. That is the intended fix, and it is the line worth reading closely. The check is also a soft one — the noise-cancellation artifact is publicly resolvable and `ManagedAudioProcessingFactory` is public, so this closes the default path rather than making it airtight; definitive enforcement is server-side.

### 🎨 UI Changes

None in the SDK. In the demo app the noise-cancellation menu item now reflects SDK-driven changes rather than only local taps.

### Testing

```
./gradlew :stream-video-android-core:testDebugUnitTest   # 1014 tests, 0 failures, 50 pre-existing skips
./gradlew :stream-video-android-core:apiCheck            # passes against the regenerated dump
./gradlew :demo-app:compileDevelopmentDebugKotlin
```

New coverage, 18 cases:

- `NoiseCancellationPolicyTest` (9) — capability x mode matrix, including an unrecognised mode and the settings-not-yet-resolved case.
- `NoiseCancellationPolicyGateTest` (5) — enabling refused without the capability, with the mode disabled, and before settings resolve; toggling-on refused; disabling always allowed.
- `NoiseCancellationSignalTest` (4) — updated to seed the server state the gate now requires.

Three gate tests were verified red with the gate removed. Server state is seeded reflectively in `CallTestSeams` rather than by adding setters to `CallState` that would exist only for tests.

## Summary by CodeRabbit

* **New Features**
  * Added noise cancellation controls with call-level state visibility.
  * Noise cancellation now respects call settings and permissions, including automatic activation where configured.
  * Added customizable participant labels and video preview styling for the call lobby.
  * Improved audio-processing status reporting and per-call state handling.

* **Bug Fixes**
  * Fixed race conditions affecting participant state initialization.
  * Ensured noise cancellation is disabled when no longer permitted.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -68,6 +68,7 @@ import io.getstream.video.android.core.call.components.CallSessionManager
 import io.getstream.video.android.core.call.components.CallStatsReporter
 import io.getstream.video.android.core.call.components.ClientCallRegistry
 import io.getstream.video.android.core.call.components.MediaManagerFactory
+import io.getstream.video.android.core.call.components.NoiseCancellationPolicy
 import io.getstream.video.android.core.call.components.RtcSessionFactory
 import io.getstream.video.android.core.call.components.SessionMonitor
 import io.getstream.video.android.core.call.connection.StreamPeerConnectionFactory
@@ -97,6 +98,8 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 import org.threeten.bp.OffsetDateTime
@@ -499,12 +502,18 @@ public class Call(
     @Volatile
     private var noiseCancellationSignalPending: Boolean = false
 
+    private val noiseCancellationPolicy = NoiseCancellationPolicy()
+
+    /** Guards the call type's auto-on default so it is applied at most once per call. */
+    private var autoOnApplied: Boolean = false
+
     init {
         media.startAudioLevelMonitoring()
         powerManager = safeCallWithDefault(null) {
             clientImpl.context.getSystemService(POWER_SERVICE) as? PowerManager
         }
         observeNoiseCancellationSignalTarget()
+        observeNoiseCancellationPolicy()
     }
 
     /** Basic crud operations */
@@ -907,16 +916,38 @@ public class Call(
      * cannot differ between two calls running at the same time.
      */
     fun setAudioProcessingEnabled(enabled: Boolean) {
+        if (enabled && !isNoiseCancellationAllowed()) {
+            logger.w {
+                "[setAudioProcessingEnabled] #sfu; rejected, noise cancellation is not " +
+                    "available for this call"
+            }
+            return
+        }
         media.setAudioProcessingEnabled(enabled)
-        // Signals what was actually applied, not what was asked for: with no audio processor
-        // configured the request is a no-op locally, and the SFU must not be told otherwise.
+        // Publishes and signals what was actually applied, not what was asked for: with no audio
+        // processor configured the request is a no-op locally, and neither the state flow nor the
+        // SFU must be told otherwise.
+        publishAudioProcessingState()
         notifyNoiseCancellationState(media.isAudioProcessingEnabledIfCreated())
     }
 
-    fun toggleAudioProcessing(): Boolean = media.toggleAudioProcessing().also {
-        // Signals what is running, not what the toggle now reports: before a factory exists the
-        // toggle reflects the wanted state, and the SFU must only hear about real processing.
-        notifyNoiseCancellationState(media.isAudioProcessingEnabledIfCreated())
+    fun toggleAudioProcessing(): Boolean {
+        // Reads without building a factory: the gate runs before join, and a factory created
+        // there would capture the pre-join audio bitrate profile.
+        if (!media.isAudioProcessingEnabled() && !isNoiseCancellationAllowed()) {
+            logger.w {
+                "[toggleAudioProcessing] #sfu; rejected, noise cancellation is not " +
+                    "available for this call"
+            }
+            return false
+        }
+        return media.toggleAudioProcessing().also {
+            publishAudioProcessingState()
+            // Signals what is running, not what the toggle now reports: before a factory exists
+            // the toggle reflects the wanted state, and the SFU must only hear real processing.
+            notifyNoiseCancellationState(media.isAudioProcessingEnabledIfCreated())
+        }
+    }
     }
 
     /**
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
@@ -948,6 +948,64 @@ public class Call(
             notifyNoiseCancellationState(media.isAudioProcessingEnabledIfCreated())
         }
     }
+
+    /**
+     * Whether the server permits noise cancellation for this call, per the capability granted to
+     * the user and the call type's noise-cancellation mode.
+     */
+    private fun isNoiseCancellationAllowed(): Boolean = noiseCancellationPolicy.isAllowed(
+        capabilities = state.ownCapabilities.value,
+        settings = state.settings.value,
+    )
+
+    /**
+     * Publishes the state observers should see: what the call asked for while no factory exists,
+     * and what is actually processing once one does. Never builds a factory.
+     */
+    private fun publishAudioProcessingState() {
+        state.setAudioProcessingEnabled(media.isAudioProcessingEnabled())
+    }
+
+    /**
+     * Keeps noise cancellation inside what the server allows.
+     *
+     * Deliberately waits for settings to resolve before deciding anything: treating "not told
+     * yet" as "not allowed" would switch noise cancellation off and back on during join. Once
+     * resolved this applies the call type's auto-on default, and withdraws noise cancellation if
+     * the capability or the mode is taken away mid-call.
+     */
+    private fun observeNoiseCancellationPolicy() {
+        scope.launch {
+            state.settings
+                .filterNotNull()
+                .combine(
+                    state.ownCapabilities,
+                ) { settings, capabilities -> settings to capabilities }
+                .collect { (settings, capabilities) ->
+                    if (!noiseCancellationPolicy.isAllowed(capabilities, settings)) {
+                        // Withdraws unconditionally rather than only when something is already
+                        // processing: a state wanted before the factory existed is remembered and
+                        // would otherwise be applied the moment one is built, after the server
+                        // had withheld it.
+                        val wasProcessing = media.isAudioProcessingEnabledIfCreated()
+                        if (wasProcessing || media.isAudioProcessingWanted()) {
+                            logger.i {
+                                "[noiseCancellation] withdrawing, no longer allowed for this call"
+                            }
+                            media.setAudioProcessingEnabled(false)
+                            publishAudioProcessingState()
+                            notifyNoiseCancellationState(false)
+                        }
+                        return@collect
+                    }
+                    if (!autoOnApplied) {
+                        autoOnApplied = true
+                        if (noiseCancellationPolicy.isAutoOn(capabilities, settings)) {
+                            setAudioProcessingEnabled(true)
+                        }
+                    }
+                }
+        }
     }
 
     /**
@@ -1009,6 +1067,7 @@ public class Call(
                 // while joining and after the signal that found no session.
                 val applied = media.isAudioProcessingEnabledIfCreated()
                 if (noiseCancellationSignalPending || desiredNoiseCancellationEnabled || applied) {
+                    publishAudioProcessingState()
                     notifyNoiseCancellationState(applied)
                 }
             }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
@@ -314,6 +314,19 @@ internal class CallMediaManager(
         _peerConnectionFactory?.isAudioProcessingEnabled() ?: false
 
     /**
+     * Whether this call wants audio processing, whether or not a factory exists to run it yet.
+     *
+     * The wanted state outlives the factory, so a policy that withholds noise cancellation has to
+     * clear it — otherwise the next factory applies it after the server said no.
+     */
+    fun isAudioProcessingWanted(): Boolean = desiredAudioProcessingEnabled
+
+    /** Forgets the wanted audio-processing state, so nothing is re-applied after the call ends. */
+    fun resetDesiredAudioProcessing() {
+        desiredAudioProcessingEnabled = false
+    }
+
+    /**
      * Records the wanted audio-processing state and applies it to the factory if one exists.
      * A factory built later picks the value up on creation, so this never has to build one.
      */
@@ -343,6 +356,9 @@ internal class CallMediaManager(
     }
 
     fun cleanup() {
+        // The wanted state must not outlive the call: a reused Call would otherwise re-apply it
+        // to the factory built for the next session.
+        resetDesiredAudioProcessing()
         mediaManager.cleanup()
     }
 }
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
@@ -76,7 +76,9 @@ internal class CallMediaManager(
 
     /**
      * Audio-processing state this call wants, remembered separately from the factory so it
-     * survives one not existing yet or being recreated.
+     * survives one not existing yet or being recreated. Applying it must never itself build a
+     * factory: one created before join captures the pre-join audio bitrate profile and defeats
+     * [ensureFactoryMatchesAudioProfile].
      *
      * The processor itself is owned by the client and shared by every call, and its enabled flag
      * lives on that one instance. Rather than clearing the flag when a call ends — which reaches
@@ -303,18 +305,17 @@ internal class CallMediaManager(
      * Audio-processing state as far as it is known, without forcing the peer-connection factory
      * to be built. False before the factory exists — nothing can be processing yet.
      *
-     * Used where the state is reported rather than changed, which must never bring a factory into
-     * existence: one created before join would be built with the pre-join audio bitrate profile
-     * and defeat [ensureFactoryMatchesAudioProfile].
+     * Used where the state is reported rather than changed — state publishing, the policy gate
+     * and signalling — which must never bring a factory into existence: one created before join
+     * would be built with the pre-join audio bitrate profile and defeat
+     * [ensureFactoryMatchesAudioProfile].
      */
     fun isAudioProcessingEnabledIfCreated(): Boolean =
         _peerConnectionFactory?.isAudioProcessingEnabled() ?: false
 
     /**
      * Records the wanted audio-processing state and applies it to the factory if one exists.
-     * A factory built later picks the value up on creation, so this never has to build one:
-     * a factory created before join captures the pre-join audio bitrate profile and defeats
-     * [ensureFactoryMatchesAudioProfile].
+     * A factory built later picks the value up on creation, so this never has to build one.
      */
     fun setAudioProcessingEnabled(enabled: Boolean) {
         desiredAudioProcessingEnabled = enabled
--- a/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
+++ b/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
@@ -16,6 +16,8 @@
 
 package io.getstream.video.android.core
 
+import io.getstream.android.video.generated.models.NoiseCancellationSettings
+import io.getstream.android.video.generated.models.OwnCapability
 import io.getstream.result.Result
 import io.getstream.video.android.core.base.IntegrationTestBase
 import io.getstream.video.android.core.call.RtcSession
@@ -55,9 +57,22 @@ class NoiseCancellationSignalTest : IntegrationTestBase(connectCoordinatorWS = f
         injectPeerConnectionFactory(factory)
     }
 
-    /** A joined call whose audio processing can genuinely be turned on. */
+    /**
+     * Seeds the server state noise cancellation needs: capability granted and the call type's mode
+     * set to available. Without both the policy gate refuses and nothing is signalled — see
+     * [NoiseCancellationPolicyGateTest].
+     */
+    private fun Call.allowNoiseCancellation() {
+        state.injectServerState(
+            capabilities = listOf(OwnCapability.EnableNoiseCancellation),
+            settings = noiseCancellationSettings(NoiseCancellationSettings.Mode.Available),
+        )
+    }
+
+    /** A joined, allowed call whose audio processing can genuinely be turned on. */
     private fun callWithProcessor(): Pair<Call, RtcSession> {
         val call = client.call("default", randomUUID())
+        call.allowNoiseCancellation()
         call.injectWorkingProcessor()
         val session = mockk<RtcSession>(relaxed = true)
         call.injectSession(session)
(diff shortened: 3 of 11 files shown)


PratimMallick:
## Review notes

Policy ownership, waiting for settings to resolve (instead of treating “not told yet” as refuse), always allowing disable, and moving auto-on into core are the right shape. A few lifecycle gaps around deferred factory + publish/notify look like real merge blockers.

### 1. High — factory later applies desired NC silently

`CallMediaManager.setAudioProcessingEnabled` only stores `desiredAudioProcessingEnabled` when no factory exists. The factory getter later applies it:

```kotlin
).also { factory ->
    desiredAudioProcessingEnabled?.let(factory::setAudioProcessingEnabled)
}
```

…but never triggers `publishAudioProcessingState()` or `notifyNoiseCancellationState()`. So after auto-on / early `set(true)` before join, local NC can turn on when the factory is built while `CallState.audioProcessingEnabled` stays `false` and the SFU never gets `StartNoiseCancellation`.

### 2. High — SFU signal still not flushed when a session appears

Same hole as #1770: `notifyNoiseCancellationState` no-ops when `session == null`. Core auto-on now calls `setAudioProcessingEnabled(true)` when settings resolve (often pre-join), so the skipped start is never retried on join / rejoin / migrate.

### 3. High — withdrawal ignores desired state while factory is absent

```kotlin
if (media.isAudioProcessingEnabledIfCreated()) {
    media.setAudioProcessingEnabled(false)
    ...
}
```

If the user/auto-on set desired `true` before the factory existed, capability/mode withdrawal does nothing. The next factory create can still apply desired `true` and enable NC while disallowed. Withdrawal should clear `desiredAudioProcessingEnabled` (and notify stop if a session exists) regardless of whether the factory is up yet.

### 4. Medium — `set` notifies requested; publish uses applied-if-created

`setAudioProcessingEnabled` still does `notifyNoiseCancellationState(enabled)` with the request, while `publishAudioProcessingState()` uses `isAudioProcessingEnabledIfCreated()`. `toggle` notifies the resolved value. Requested vs applied can diverge when the factory is absent or processing is unattached (also amplified by #1771’s MUSIC_HIGH_QUALITY no-op).

### 5. Medium — `toggle` / `isAudioProcessingEnabled()` still lazy-build the factory

`set` carefully avoids building a factory, but:

```kotlin
fun isAudioProcessingEnabled() = peerConnectionFactory.isAudioProcessingEnabled()
fun toggleAudioProcessing() = peerConnectionFactory.toggleAudioProcessing()
```

`Call.toggleAudioProcessing` calls `isAudioProcessingEnabled()` first for the gate, so opening settings / toggling can build the factory before join and pin the pre-join audio bitrate profile — the exact problem deferred-set was meant to avoid. Prefer the if-created path (or equivalent) for the gate.

### 6. Medium — cleanup / Call teardown leave desire flags

`CallMediaManager.cleanup` resets the shared module on an existing factory but does not clear `desiredAudioProcessingEnabled`. `Call` never resets `autoOnApplied` / `desiredNoiseCancellationEnabled`. If the same `Call` object is reused after leave, remembered NC intent can leak into the next session.

---

**Suggested fix direction:** when desired NC is applied on factory create (and on session install), publish + notify from one place; on policy withdraw, always clear desired; align `set` notify with applied state; keep toggle/isEnabled off the lazy factory getter pre-join.

PratimMallick (commented):
Moved the earlier conversation comment onto the relevant lines. Policy ownership and waiting for settings to resolve look right — the notes below are lifecycle gaps around deferred factory + publish/notify.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
```diff
@@ -85,7 +93,9 @@ internal class CallMediaManager(
                     audioBitrateProfileProvider = { mediaManager.microphone.audioBitrateProfile.value },
                     sharedEglBaseProvider = { eglBase() },
                     webRtcLoggingLevel = clientImpl.loggingLevel.webRtcLoggingLevel,
-                )
+                ).also { factory ->
+                    desiredAudioProcessingEnabled?.let(factory::setAudioProcessingEnabled)
```
PratimMallick: **High — factory later applies desired NC silently**

When `setAudioProcessingEnabled` / auto-on runs before a factory exists, only `desiredAudioProcessingEnabled` is stored. This apply on create never triggers `publishAudioProcessingState()` or `notifyNoiseCancellationState()`.

After auto-on / early `set(true)` before join, local NC can turn on here while `CallState.audioProcessingEnabled` stays `false` and the SFU never gets `StartNoiseCancellation`.

**Suggestion:** when desired is applied on factory create (and on session install), publish + notify from one place.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -886,12 +895,80 @@ public class Call(
...
+                            media.setAudioProcessingEnabled(false)
+                            publishAudioProcessingState()
+                            notifyNoiseCancellationState(false)
+                        }
+                        return@collect
+                    }
+                    if (!autoOnApplied) {
+                        autoOnApplied = true
+                        if (noiseCancellationPolicy.isAutoOn(capabilities, settings)) {
+                            setAudioProcessingEnabled(true)
```
PratimMallick, line 1004: **High — SFU signal still not flushed when a session appears**

Auto-on calls `setAudioProcessingEnabled(true)` when settings resolve, often **before** join. `notifyNoiseCancellationState` no-ops when `session == null`, and nothing re-sends on join / rejoin / migrate (same hole as #1770).

**Suggestion:** flush desired/local NC state when a session is installed.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -886,12 +895,80 @@ public class Call(
...
+    private fun observeNoiseCancellationPolicy() {
+        scope.launch {
+            state.settings
+                .filterNotNull()
+                .combine(
+                    state.ownCapabilities,
+                ) { settings, capabilities -> settings to capabilities }
+                .collect { (settings, capabilities) ->
+                    if (!noiseCancellationPolicy.isAllowed(capabilities, settings)) {
+                        if (media.isAudioProcessingEnabledIfCreated()) {
```
PratimMallick: **High — withdrawal ignores desired state while factory is absent**

If desired was set `true` before the factory existed, this check is false and withdrawal does nothing. The next factory create can still apply desired `true` and enable NC while disallowed.

**Suggestion:** always clear `desiredAudioProcessingEnabled` on withdraw (and notify stop if a session exists), regardless of whether the factory is up yet.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -886,12 +895,80 @@ public class Call(
...
+        if (enabled && !isNoiseCancellationAllowed()) {
+            logger.w {
+                "[setAudioProcessingEnabled] #sfu; rejected, noise cancellation is not " +
+                    "available for this call"
+            }
+            return
+        }
         media.setAudioProcessingEnabled(enabled)
+        publishAudioProcessingState()
         notifyNoiseCancellationState(enabled)
```
PratimMallick: **Medium — `set` notifies requested; publish uses applied-if-created**

`notifyNoiseCancellationState(enabled)` uses the request, while `publishAudioProcessingState()` uses `isAudioProcessingEnabledIfCreated()`. `toggle` notifies the resolved value. Requested vs applied can diverge when the factory is absent or processing is unattached (also amplified by #1771’s MUSIC_HIGH_QUALITY no-op).

**Suggestion:** notify the applied state, consistent with toggle.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -886,12 +895,80 @@ public class Call(
...
     }
 
-    fun toggleAudioProcessing(): Boolean = media.toggleAudioProcessing().also { enabled ->
-        notifyNoiseCancellationState(enabled)
+    fun toggleAudioProcessing(): Boolean {
+        if (!isAudioProcessingEnabled() && !isNoiseCancellationAllowed()) {
```
PratimMallick: **Medium — gate still lazy-builds the factory**

`CallMediaManager.setAudioProcessingEnabled` carefully avoids building a factory, but `isAudioProcessingEnabled()` (used here) and `toggleAudioProcessing()` still go through the lazy `peerConnectionFactory` getter.

Opening settings / toggling can therefore build the factory before join and pin the pre-join audio bitrate profile — the exact problem deferred-set was meant to avoid.

**Suggestion:** use the if-created path (or equivalent) for the gate / pre-join reads.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
```diff
@@ -269,8 +279,24 @@ internal class CallMediaManager(
...
+    fun isAudioProcessingEnabledIfCreated(): Boolean =
+        _peerConnectionFactory?.isAudioProcessingEnabled() ?: false
+
+    /**
+     * Records the wanted audio-processing state and applies it to the factory if one exists.
+     * A factory built later picks the value up on creation, so this never has to build one.
+     */
     fun setAudioProcessingEnabled(enabled: Boolean) {
-        return peerConnectionFactory.setAudioProcessingEnabled(enabled)
+        desiredAudioProcessingEnabled = enabled
```
PratimMallick, line 334: **Medium — cleanup / teardown leave desire flags**

`desiredAudioProcessingEnabled` is recorded here, but `cleanup` (from #1771) only resets the shared module on an existing factory — it does not clear this field. `Call` also never resets `autoOnApplied` / `desiredNoiseCancellationEnabled`.

If the same `Call` object is reused after leave, remembered NC intent can leak into the next session.

aleksandar-apostolov: Half taken, half deferred.

The media-level wanted state is now cleared on cleanup — that's the one that would be re-applied to the next factory.

`autoOnApplied` and `desiredNoiseCancellationEnabled` on `Call` are not. Both need a teardown hook in `CallLifecycleManager` that doesn't exist, and reachability is limited to reusing a `Call` instance after leave, which the client drops from its map with `isDestroyed` set. Happy to add the hook if you'd rather not carry it — I didn't want to grow the component surface for a lifecycle we don't support.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
PratimMallick: Addressed in the stack — factory apply on create + session observer flushing **applied** state (and publishing the StateFlow) covers the join path, where `RtcSession` builds the factory before `setActiveSession`. The new test `a state applied only once the factory exists is signalled on join` locks it in.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Addressed in #1770 — session collect flushes applied NC state on install / replacement.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Addressed — withdrawal now clears wanted state even when nothing is processing yet (`isAudioProcessingWanted()`), with a gate test for that path.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Addressed — `set` now publishes and notifies applied-if-created, same as toggle.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: The **gate** is fixed (`isAudioProcessingEnabledIfCreated()`), so a refused toggle no longer builds a factory. One residual: once the gate allows the toggle, `media.toggleAudioProcessing()` still goes through the lazy getter — commenting on that separately.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallMediaManager.kt
PratimMallick: Agreed on the defer. Clearing `desiredAudioProcessingEnabled` on cleanup is the one that would be re-applied; `autoOnApplied` / `desiredNoiseCancellationEnabled` only matter if a `Call` is reused after leave, which this client does not support. Fine to leave.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -907,14 +916,90 @@ public class Call(
...
+        // Reads without building a factory: the gate runs before join, and a factory created
+        // there would capture the pre-join audio bitrate profile.
+        if (!media.isAudioProcessingEnabledIfCreated() && !isNoiseCancellationAllowed()) {
+            logger.w {
+                "[toggleAudioProcessing] #sfu; rejected, noise cancellation is not " +
+                    "available for this call"
+            }
+            return false
+        }
+        return media.toggleAudioProcessing().also { enabled ->
```
PratimMallick: **Medium — allowed toggle still lazy-builds the factory**

The gate correctly uses `isAudioProcessingEnabledIfCreated()`, so a *refused* toggle no longer builds a factory (test covers that).

Once the gate allows it, `media.toggleAudioProcessing()` still goes through `peerConnectionFactory`, which constructs one. A pre-join toggle after settings have resolved (custom lobby, `get()` then toggle) can still pin the pre-join audio bitrate profile and then hit the #1771 leak-on-recreate path.

`setAudioProcessingEnabled` already records desired without building. Toggle could flip `desiredAudioProcessingEnabled` the same way when no factory exists, and only call into the factory when one is already there.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -907,14 +916,90 @@ public class Call(
      * cannot differ between two calls running at the same time.
      */
     fun setAudioProcessingEnabled(enabled: Boolean) {
+        if (enabled && !isNoiseCancellationAllowed()) {
```
PratimMallick, line 919: **Low — `set(true)` before settings resolve is dropped**

`isAllowed(null settings)` is false, so an enable before `get()`/`join` returns is silently ignored. Auto-on later covers `Mode.AutoOn`; it does not cover an app that called `setAudioProcessingEnabled(true)` on a call type whose mode is `Available`.

Disable still goes through, which is right. If this is intentional (don’t enable until we know the server allows it), worth a one-line KDoc so callers don’t expect the request to be remembered.

PratimMallick (commented):
Re-reviewed. The previous highs (silent factory apply, session flush, withdrawal of wanted state, set vs applied, refused-toggle building a factory, cleanup of desired) are addressed. Two residuals below — neither is a merge blocker if the join-order assumption holds.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Not blocking. Demo toggles in-call (factory already exists). Pre-join `set` already records desired without building; this is only an issue for apps that call `toggleAudioProcessing()` in the lobby after settings have loaded. Fine to follow up separately.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: Not blocking. Intentional: don’t enable until we know the server allows it. Auto-on covers `Mode.AutoOn`; a silent drop for `Available` before settings resolve is acceptable. Optional KDoc can wait.

PratimMallick (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 ## stream-video-android
 ### 🐞 Fixed
+Noise cancellation is now gated in the SDK by the `enable-noise-cancellation` capability granted to the user and by the call type's noise-cancellation mode. Enabling it is refused when either withholds it, and it is switched off if either is withdrawn mid-call. The call type's `auto-on` mode is also applied by the SDK now, so apps no longer need to read the setting and call `setAudioProcessingEnabled` themselves during pre-join.
+Add `CallState.audioProcessingEnabled`, a `StateFlow` reflecting whether noise cancellation is running. Observe it instead of snapshotting `Call.isAudioProcessingEnabled()` — the SDK can now change the state on its own, so a snapshot goes stale.
```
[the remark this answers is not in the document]
aleksandar-apostolov: Removed rather than relocated — the changelog is no longer part of the release process here, so this PR no longer touches it. Release notes come from PR titles.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -886,12 +895,80 @@ public class Call(
...
+            logger.w {
+                "[toggleAudioProcessing] #sfu; rejected, noise cancellation is not " +
+                    "available for this call"
+            }
+            return false
+        }
+        return media.toggleAudioProcessing().also { enabled ->
+            publishAudioProcessingState()
+            notifyNoiseCancellationState(enabled)
+        }
```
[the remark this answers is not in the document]
aleksandar-apostolov: Both parts are addressed.

The gate now reads through a non-creating accessor, so a refused toggle cannot bring a factory into existence. There is a test asserting no factory exists after one.

The media toggle is expressed as a set against the wanted state, so it records `desiredAudioProcessingEnabled` and never reaches the lazily-creating accessor. That half landed in the parent PR, which has since merged.

Worth noting the consequence you were pointing at is sharper than "loses its selected state after recreation": a factory holding the client's shared audio processor is deliberately kept rather than released, so a factory built too early could not be corrected at all — the pre-join profile would be pinned for the rest of the call.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
```diff
@@ -33,11 +35,25 @@ class NoiseCancellationSignalTest : IntegrationTestBase(connectCoordinatorWS = f
...
+    private fun allowedCall(): Pair<Call, RtcSession> {
         val call = client.call("default", randomUUID())
+        call.state.injectServerState(
+            capabilities = listOf(OwnCapability.EnableNoiseCancellation),
+            settings = noiseCancellationSettings(NoiseCancellationSettings.Mode.Available),
+        )
         val session = mockk<RtcSession>(relaxed = true)
         call.injectSession(session)
+        return call to session
+    }
```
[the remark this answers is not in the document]
aleksandar-apostolov: Already the case — the rapid-changes test builds its call through a helper that grants the capability and sets the mode, so it exercises a real on-to-off transition. It also asserts the last state the SFU is left holding rather than that a stop happened at some point.

CHANGELOG.md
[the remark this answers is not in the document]
aleksandar-apostolov: Removed rather than relocated — the changelog is no longer part of the release process here, so this PR no longer touches it. Release notes come from PR titles.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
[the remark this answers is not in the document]
aleksandar-apostolov: Both parts are addressed.

The gate now reads through a non-creating accessor, so a refused toggle cannot bring a factory into existence. There is a test asserting no factory exists after one.

The media toggle is expressed as a set against the wanted state, so it records `desiredAudioProcessingEnabled` and never reaches the lazily-creating accessor. That half landed in the parent PR, which has since merged.

Worth noting the consequence you were pointing at is sharper than "loses its selected state after recreation": a factory holding the client's shared audio processor is deliberately kept rather than released, so a factory built too early could not be corrected at all — the pre-join profile would be pinned for the rest of the call.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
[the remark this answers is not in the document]
aleksandar-apostolov: Already the case — the rapid-changes test builds its call through a helper that grants the capability and sets the mode, so it exercises a real on-to-off transition. It also asserts the last state the SFU is left holding rather than that a stop happened at some point.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -886,12 +895,80 @@ public class Call(
     fun isAudioProcessingEnabled(): Boolean = media.isAudioProcessingEnabled()
 
     fun setAudioProcessingEnabled(enabled: Boolean) {
+        if (enabled && !isNoiseCancellationAllowed()) {
+            logger.w {
+                "[setAudioProcessingEnabled] #sfu; rejected, noise cancellation is not " +
+                    "available for this call"
+            }
+            return
+        }
```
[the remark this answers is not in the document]
aleksandar-apostolov: Declining this one — the drop is the enforcement working, not a gap.

Enabling before the server has said noise cancellation is permitted is exactly what this change exists to prevent. Retaining the request and applying it once settings resolve would mean the SDK switching noise cancellation on at some later moment the caller did not ask for; that behaviour already exists and is scoped to the call type's `auto-on` mode, where the server is the one asking for it.

It is also not silent: the refusal logs a warning, and `CallState.audioProcessingEnabled` reflects what is actually in effect, so a caller can observe the outcome rather than infer it.

Disabling stays available immediately and unconditionally, which is the part that has to hold.

In practice a UI reads the settings to decide whether to show the control at all, so the pre-settings window is not one a user can act in.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/call/components/NoiseCancellationPolicyTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import io.getstream.android.video.generated.models.CallSettingsResponse
+import io.getstream.android.video.generated.models.NoiseCancellationSettings
+import io.getstream.android.video.generated.models.OwnCapability
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class NoiseCancellationPolicyTest {
```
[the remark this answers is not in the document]
aleksandar-apostolov: Declining — `TestBase` would make this test worse.

Its `init` block runs `StreamService.instance.getAuthData(...)` inside `runBlocking`, so every subclass performs a live network call against a real environment before any test executes. It also brings a dispatcher rule and an Android context.

`NoiseCancellationPolicy` is a pure decision object over a capability list and a settings response. This test hands it mocks and asserts return values — it needs no context, no dispatcher, and certainly no auth round-trip. Extending `TestBase` would turn a hermetic, millisecond-scale unit test into one that depends on the network being up.

PR #1775, opened 2026-08-19
[AND-1410] Correctly register Telecom video and streaming capabilities

rahul-lohra:
### Goal

Closes #AND-1410

Correct the app-wide Jetpack Telecom capability registration. The existing implementation combines the video-calling and call-streaming flags using bitwise `and`, which evaluates to `0` because the flags occupy different bits.
Use bitwise `or` so Android Telecom is correctly informed that the SDK supports both capabilities.

Basic Telecom integration and audio routing worked because AndroidX still registered a self-managed account and the per-call attributes were correct. However, app-wide video and call-streaming capabilities were missing because the capability mask evaluated to zero.

### Implementation

- Combine Telecom capabilities using bitwise `or`.
- Add a regression test verifying that video calling and call streaming are
  both passed to `registerAppWithTelecom`.

### 🎨 UI Changes

None

### Testing
Smoke test with telecom enabled


rahul-lohra:
Will be part of Android 17 telecom change

PR #1730, opened 2026-06-19
Move @Preview composables out of published artifact to fix R8 builds

PratimMallick:
### Goal
Fixes [1244](https://linear.app/stream/issue/AND-1244/fix-r8-build-failure-with-streampreviewdatautils-classes-referenced)

R8/minified release builds of apps consuming `stream-video-android-ui-compose` fail with `Missing class io.getstream.video.android.mock.StreamPreviewDataUtils` (and `StreamPreviewDataUtilsKt`). The compose module depends on `stream-video-android-previewdata` as `compileOnly`, but the Compose compiler emits `ComposableSingletons$*` lambda holders for the inline `@Preview` functions into production classes, so the preview-data references leak into the published release artifact even though that module is never on a consumer's runtime classpath.

Fixes #1448

### Implementation

- Moved every `@Preview` composable out of `src/main` into a new **debug-only** source set (`src/debug/kotlin`, one `<Name>Preview.kt` per original file, same package). Previews still render in Android Studio (debug variant) but are no longer compiled into the release artifact.
- Changed the `stream-video-android-previewdata` dependency from `compileOnly` to `debugImplementation`.
- Two production composables referenced by previews were widened `private` → `internal` (`CalLCenterContent`, `ParticipantActionsWithoutState`). One preview that was also used by production code (`ModerationWarningUiContentDemo` via `LocalInspectionMode`) stays in `main` with the `@Preview` moved to a thin debug wrapper.
- Declared `kotlinx-datetime` (0.6.1) explicitly as an `implementation` dependency of the compose module — it is used by the livestream countdown UI in production but was only arriving transitively, which also caused `Missing class kotlinx.datetime.*` R8 errors (reported in the same issue thread).
- Regenerated the public API dump: 50 compiler-generated `ComposableSingletons$*` preview classes are removed from the published API surface.

### Testing

- [x] `./gradlew :stream-video-android-ui-compose:compileReleaseKotlin` — published variant compiles without preview-data.
- [x] `./gradlew :stream-video-android-ui-compose:compileDebugKotlin` — previews still compile (IDE rendering intact).
- [x] `./gradlew :stream-video-android-ui-compose:spotlessCheck` + `:apiCheck` — pass (API baseline updated via `apiDump`).
- [x] `./gradlew :stream-video-android-ui-compose:testDebugUnitTest` — pass.
[...]

Issue #1448, opened 2025-06-16
R8 build failure with StreamPreviewDataUtils classes referenced from ComposableSingletons


matteinn:
**Describe the bug**
When building with R8 enabled, I encounter build failures due to missing preview data utilities. The library includes `stream-video-android-previewdata` as `compileOnly`, but Compose generates `ComposableSingletons` classes that reference these preview utilities in production code.

```
ERROR: R8: Missing class io.getstream.video.android.mock.StreamPreviewDataUtils (referenced from: void io.getstream.video.android.compose.ui.components.call.renderer.internal.ComposableSingletons$ScreenShareVideoRendererKt$lambda-2$1.invoke(androidx.compose.runtime.Composer, int))

Missing class io.getstream.video.android.mock.StreamPreviewDataUtilsKt (referenced from: void io.getstream.video.android.compose.ui.components.call.ComposableSingletons$CallAppBarKt$lambda-7$1.invoke(androidx.compose.runtime.Composer, int) and 35 other contexts)
```

The automatically generated missing_rules.txt file doesn't work:
```
# Please add these rules to your existing keep rules in order to suppress warnings.
# This is generated automatically by the Android Gradle plugin.
-dontwarn io.getstream.video.android.mock.StreamPreviewDataUtils
-dontwarn io.getstream.video.android.mock.StreamPreviewDataUtilsKt
```

The only workaround I've found so far is including the preview module in runtime dependencies of the consumer app, even though it adds unnecessary code:
```
implementation("io.getstream:stream-video-android-previewdata:same-version")
```

Kotlin generates `ComposableSingletons` classes that contain lambda references to all Composable functions in a file, including those marked with `@Preview`. Even though preview functions like `ScreenShareVideoRendererPreview()` are marked `private`, the compiler still generates references to them that are accessible from production code.

Suggested Solutions
1. Move all preview-specific code to dedicated files in a preview source set that can be excluded from release builds
2. Change compileOnly dependency on `stream-video-android-previewdata` to `api` to ensure it's available in production
3. Modify the library to avoid using Preview Data Utils directly in Compose Previews, or provide comprehensive ProGuard rules

👉  The first approach is recommended to avoid shipping unnecessary code in production apps.

**SDK version**
- 1.6.3

**To Reproduce**
1. Include the Stream Video Compose UI library in an Android app
2. Enable R8
3. Try to build the app

**Expected behavior**
App builds fine.

**Device:**
Any

**Screenshots**
None

[...]

PR #1776, opened 2026-08-20
Snapshot test baseline for the Compose Video SDK

andremion:
### Goal

Establish a green, stable Paparazzi snapshot baseline for the Compose Video SDK before the design refresh work starts, so the follow-up refactors can prove they cause zero visual change. Part of [AND-1417](https://linear.app/stream/issue/AND-1417/snapshot-test-baseline-for-the-compose-video-sdk).

The CI gate itself already landed in #1767 (conventions 0.14.0 wires `verifyPaparazziDebug` into the `testCoverage` task that CI runs on every PR).

### Implementation

- Adopt the chat SDK test conventions: a `PaparazziComposeTest` interface with `snapshot`, `snapshotWithDarkMode`, and `snapshotWithDarkModeRow` helpers and shared hdpi device configs (`PIXEL_2_HDPI`, `PIXEL_4A_HDPI`, `PIXEL_2_LANDSCAPE_HDPI`). It replaces the `BaseComposeTest` abstract class. All 12 existing test classes migrated.
- Fix the flakiness behind the 10 `@Ignore`d tests (they pointed at AND-786, which is canceled) and enable all of them. The root cause was nondeterministic preview state in `stream-video-android-previewdata`: participant session ids were random UUIDs, `Call.sessionId` itself is a random UUID per JVM, and `previewParticipantsList` cleared and mutated the shared `previewCall` state on every access. Now `previewCall.sessionId` is pinned, session ids are deterministic (`"session-N-userId"`), all six participants are populated once at init, and the clear side effects are removed. No public API change (`apiCheck` passes).
- Extend coverage from 74 to 143 goldens: all individual control actions, dialogs (StreamDialog, invite users, participant actions, moderation warning), spotlight renderers, livestream backstage, 7-participant grids, lobby states. Every snapshot test (not only new coverage) follows the shared preview convention: an `internal fun XxxPreview()` composable rendered by both the `@Preview` wrapper and the snapshot test, so IDE previews and goldens cannot drift apart. The unification kept every golden byte identical (verified by re-recording), with the test content as the source of truth. One deviation from chat: the shared previews live in `src/debug` instead of `src/main`, because #1730 moved previews out of the published artifact (R8 issue #1448); the snapshot tests therefore live in `src/testDebug`.
- Apply the chat naming and coverage conventions: test methods are lowercase phrases (they become the golden file names), components use the combined `snapshotWithDarkMode` golden (dark half on top, light half below), and full screens get an explicit `... in dark mode` sibling test. Screen sharing tests gained the light variant they were missing, and tests that duplicated the same composition were removed.
- Use `RenderingMode.SHRINK` so component goldens crop to their content like in the chat repo, with two exceptions: the dialog tests (window level `Dialog` overlays do not shrink correctly) and the grid renderer tests (`PortraitVideoRendererTest`, `LandscapeVideoRendererTest`), which size tiles from `LocalConfiguration` and need the full device canvas. The grid tests were split into their own classes so the other participant component tests can shrink.
- The participant actions menu is snapshotted through `ParticipantActionsDialogContent` (extracted, internal, no API change) because its production wrapper renders through a window level `Popup`, which Paparazzi and Android Studio previews cannot capture.
- Goldens use `maxPercentDifference = 0.5`: CI measured 0.10% to 0.30% font antialiasing drift between macOS recorded goldens and Linux verification on Paparazzi 1.3.4, the newest version usable with Kotlin 1.9 (1.3.5 requires Kotlin 2.0). The chat repo runs 1.3.5 with the default threshold; revisit when this repo upgrades.
- Give the grid renderer tests a dedicated call per participant count (`previewGridCall`). The renderers read remote participants from call state for the small layouts, so the shared six-participant `previewCall` could not represent a one, two, or three participant call. This also makes the `N participants` goldens true to their names.
- Raise the test worker heap to 4g in the ui-compose module, same as the chat repo, since Paparazzi renders in the forked test JVM.
- All goldens re-recorded because names, device configs (hdpi, smaller files), and coverage changed.

### Note on the combined dark/light goldens

The `snapshotWithDarkMode` helpers render the content twice, under `VideoTheme(isInDarkMode = true)` and `false`, same as chat. In the recorded goldens both halves look identical. That is expected and not a bug in the tests: `VideoTheme` currently ignores its `isInDarkMode` parameter (colors always come from `StreamColors.defaultColors()`, resolved from single-value color resources), so the SDK is effectively dark-only today. When the StreamDesign token layer lands (phase 3 of the design refresh) and the theme gets a real light palette, the light halves of these goldens will change and will be re-recorded in that PR as visible proof of the theme change. The harness is ready for it now.

### Testing

- `verifyPaparazziDebug` green twice in a row with `--rerun-tasks`, plus filtered runs (`--tests`) to prove the goldens do not depend on test execution order.
- `apiCheck` and `spotlessCheck` green repo-wide.

## Summary by CodeRabbit

* **New Features**
  * Added reusable previews for call, audio, participant, livestream, avatar, dialog, and control components.
  * Added configurable participant scenarios, including portrait, landscape, screen sharing, and spotlight layouts.
  * Added separate previews for audio/video and enabled/disabled states.

* **Bug Fixes**
  * Preview participant data is now stable, ordered, and consistent across renders.

* **Tests**
  * Expanded light and dark mode screenshot coverage across major UI components and layouts.
  * Improved snapshot test reliability with standardized device profiles, rendering settings, and increased memory capacity.


andremion:
@coderabbitai full review

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/AudioCallContentPreview.kt
```diff
@@ -25,27 +25,72 @@ import io.getstream.video.android.mock.previewCall
...
+    )
+}
+
+@Composable
+internal fun AudioCallContentWithoutHeaderPreview() {
+    AudioCallContent(
+        call = previewCall,
+        isMicrophoneEnabled = false,
+        isShowingHeader = false,
+    )
```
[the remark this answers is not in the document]
andremion: This one is intentional. AudioCallContent is deprecated but still public API on develop, so it keeps its own snapshot coverage until it is removed. The AudioOnlyCallContent replacement is covered separately by AudioOnlyCallContentPreview and AudioOnlyCallContentWithoutHeaderPreview in the same file. Switching these two previews would drop the deprecated API coverage and duplicate the other two.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/base/StreamButtonPreview.kt
```diff
@@ -24,12 +24,12 @@ import androidx.compose.foundation.layout.height
...
 import androidx.compose.foundation.layout.width
 import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.automirrored.filled.ExitToApp
-import androidx.compose.material.icons.automirrored.filled.PhoneMissed
 import androidx.compose.material.icons.filled.AccessAlarm
 import androidx.compose.material.icons.filled.Camera
+import androidx.compose.material.icons.filled.ExitToApp
 import androidx.compose.material.icons.filled.GridView
 import androidx.compose.material.icons.filled.GroupAdd
+import androidx.compose.material.icons.filled.PhoneMissed
```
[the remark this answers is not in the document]
andremion: Fixed in c8d18f31d0: the previews now use the AutoMirrored icon variants. Rendering is identical in LTR, so no golden changed.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/LandscapeScreenSharingVideoRendererPreview.kt
```diff
@@ -25,24 +25,51 @@ import io.getstream.video.android.compose.theme.VideoTheme
...
+@Composable
+internal fun LandscapeScreenSharingMyContentPreview() {
+    LandscapeScreenSharingVideoRenderer(
+        call = previewCall,
+        session = ScreenSharingSession(participant = previewParticipantsList[0]),
+        participants = previewParticipantsList,
+        dominantSpeaker = previewParticipant,
+        modifier = Modifier.fillMaxSize(),
+    )
+}
```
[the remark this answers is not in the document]
andremion: Fixed in c8d18f31d0: the 'for other participant' fixtures (landscape and portrait) now share a remote participant's screen, and the four affected goldens were re-recorded to show the '<name> is Presenting' state.

stream-video-android-previewdata/src/main/kotlin/io/getstream/video/android/mock/StreamPreviewDataUtils.kt
```diff
@@ -54,11 +54,13 @@ public val previewCall: Call = Call(
     id = "123",
     user = previewUsers[0],
 ).apply {
-    val participants = previewUsers.take(2).map { user ->
-        val sessionId = UUID.randomUUID().toString()
+    // Deterministic session ids keep the participant order stable across runs, so previews
+    // and snapshot goldens always render the same participants in the same positions.
+    sessionId = "session-0-${previewUsers[0].id}"
+    val participants = previewUsers.mapIndexed { index, user ->
```
gpunto, line 63: The `upsertParticipants` below only writes `_participants`; `call.state.participants` is published later by `SortedParticipantsState`'s collectors on the call scope (`_sortedParticipants` starts as `emptyList()` and is filled from `scope.launch`). `previewGridCall` `runBlocking`-waits for exactly this, and its KDoc says the renderers otherwise snapshot the transient empty state. `previewCall` has no such wait, yet `CallContent.kt:401` and `AudioRoomContent.kt:94,208` read that flow and their goldens do render all six participants.

The filtered `--tests` runs prove order-independence, not timing-independence, so I think the same wait belongs here.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/PortraitVideoRendererPreview.kt
```diff
@@ -27,173 +27,122 @@ import androidx.compose.ui.tooling.preview.Preview
...
+}
+
+@Composable
+internal fun PortraitParticipantsPreview7() {
+    PortraitParticipants(participantCount = 7)
+}
+
+@Composable
+private fun PortraitParticipants(participantCount: Int) {
+    val gridCall = previewGridCall(participantCount)
```
gpunto: Non-blocking: this runs in the composable body without `remember`, so every composition builds a fresh `Call` (whose `CallState` launches collectors that are never cancelled) and blocks on `runBlocking { withTimeout(5_000) }`. Harmless for a single Paparazzi frame, but in an IDE preview it blocks the render thread per composition. `remember(participantCount) { previewGridCall(participantCount) }` would cover it. Same at `LandscapeVideoRendererPreview.kt:133`.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/call/ringing/outgoingcall/OutgoingCallDetailsPreview.kt
```diff
@@ -25,12 +25,31 @@ import io.getstream.video.android.mock.previewMemberListState
...
+    StreamPreviewDataUtils.initializeStreamVideo(LocalContext.current)
+    VideoTheme {
+        OutgoingCallDetailsAudioPreview()
+    }
+}
+
+@Composable
+internal fun OutgoingCallDetailsAudioPreview() {
+    OutgoingCallDetails(
+        isVideoType = false,
```
gpunto: Non-blocking: `outgoing call details video.png` and `outgoing call details audio.png` are byte-identical (same md5). `isVideoType` only reaches `ParticipantInformation`, whose `CallStatus.Outgoing` branch renders `stream_video_call_status_outgoing` without interpolating the call type — the `Incoming` branch does interpolate it, and that golden pair does differ. So one of these two pins nothing as it stands.

gpunto (commented):
Description covers the dark-mode-inert goldens and the preview determinism rework well, so nothing from me on those. Three things left inline.

stream-video-android-previewdata/src/main/kotlin/io/getstream/video/android/mock/StreamPreviewDataUtils.kt
andremion: Good catch. Added the same wait to previewCall in 4cfabd4b7e: after upsertParticipants it now blocks until call.state.participants reports all six, like previewGridCall does. All goldens stayed byte identical, which confirms class load timing was covering it until now.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/PortraitVideoRendererPreview.kt
andremion: Fixed in 4cfabd4b7e: both helpers now use remember(participantCount) { previewGridCall(participantCount) }, so IDE previews build the call once per count. No golden changed.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/call/ringing/outgoingcall/OutgoingCallDetailsPreview.kt
andremion: Confirmed, same md5 for the outgoing pair while the incoming pair differs, and only the Incoming branch of ParticipantInformation uses the call type. Removed the audio outgoing details test, its preview, and the golden in 4cfabd4b7e. If a redesign later differentiates the outgoing layout by call type, the variant comes back with a golden that actually pins something.

gpunto (approved):

PR #1777, opened 2026-08-24
Outgoing ring timeout no longer restarts on every call state update

aleksandar-apostolov:
### Goal

Fixes [AND-1412](https://linear.app/stream/issue/AND-1412/outgoing-ring-timer-restarts-on-every-state-recompute)

A caller could ring far past the configured `autoCancelTimeoutMs` without either side timing out. `RingingState.Outgoing` was the only member of the hierarchy without structural equality, and `updateRingingState()` allocates a fresh instance on every run — so each recomputation republished the state on `CallState.ringingState`. Every coordinator API response and eight event handlers reach that path, so a chatty update stream reset the ring timeout from zero, indefinitely.

`RingingState.Incoming` is already a `data class`, so only the caller side was affected.

Found in the Video team's ringing-reliability audit.

### Implementation

`RingingState.Outgoing` becomes a `data class`. That restores `StateFlow` conflation, which four behaviours depended on:

- `startRingingTimer()` no longer cancels and restarts the full `autoCancelTimeoutMs` delay on every update.
- `CallServiceRingingStateObserver` no longer restarts the outgoing ringtone. On API 28+ `playWithRingtoneManager` stops the current `Ringtone`, builds a new one, and plays it — so the caller heard the tone restart from the beginning on every update. The API <28 `MediaPlayer` path was already guarded by `isPlaying`.
- `DefaultNotificationHandler`'s `distinctUntilChanged()` now filters, so the ongoing-call notification is no longer rebuilt and re-posted per update.
- `previousRingingStates` is a hash set; with no `hashCode` every instance was a distinct member and it grew for the lifetime of the ring.

A genuine `acceptedByCallee` transition still passes the change check, so accept handling is unchanged. All 45 `RingingState.Outgoing` usages are type checks or read `acceptedByCallee` — nothing relied on identity.

**API dump:** additive only — the generated `component1`, `copy`, `copy$default`, `equals`, `hashCode`, `toString`. Constructors unchanged, so consumers stay binary-compatible. Version constant untouched.

The ticket also notes an optional hardening at the timer's fire condition (`Outgoing || (Incoming && activeCall == null)` by precedence). Separate concern, left for its own PR.

### 🎨 UI Changes

None.

### Testing

```
./gradlew :stream-video-android-core:testDebugUnitTest   # 1038 tests, 0 failures, 50 pre-existing skips
./gradlew :stream-video-android-core:apiCheck            # passes against the regenerated dump
./gradlew spotlessCheck
```

New coverage, 8 cases:

- `RingingStateTest` (6) — equality and `hashCode` for `Outgoing` and `Incoming`, identical states collapsing in a hash set, and the `StateFlow` not republishing identical outgoing states while still publishing a real acceptance.
- `CallServiceRingingStateObserverTest` (+2) — repeated identical outgoing states play the ringtone once; acceptance is still handled after them.

Five were verified red with the one-word fix reverted, so they aren't vacuous: `state flow does not republish identical outgoing states`, `outgoing states with the same acceptance are equal`, `identical outgoing states collapse in a hash set`, `repeated identical outgoing states do not restart the outgoing sound`, `outgoing acceptance still handled after repeated identical states`. The other three pass either way — contract guards, not regression proofs.

Not covered: the timer restart itself. `CallState.updateRingingState()` is only reachable through `IntegrationTestBase`, which connects a live coordinator WS, so there is no isolated harness for it. The equality contract is the shared root cause of all four symptoms and is pinned, but this is coverage of the cause rather than the timer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented repeated identical outgoing ringing states from triggering duplicate ringing sounds.
  * Ensured accepted calls still stop the outgoing ringing sound correctly.

* **Improvements**
  * Improved ringing-state comparisons and duplicate-state handling for more consistent call behavior.


aleksandar-apostolov:
The SonarCloud gate is red on `new_maintainability_rating`, flagging two smells in `setActiveCall` — cognitive complexity at `ClientState.kt:174` and a nested `if` at `:199`. Neither comes from this PR: the diff is a single one-word hunk at line 55.

The cause is the new-code baseline. Sonar last analysed `develop` on 2025-12-19, while `develop` is currently at `ff290bd68b` (2026-08-21), so everything merged in between reads as "new code" to any PR touching the same file — `setActiveCall`'s complexity lands on whoever edits `ClientState.kt` next. `develop` already carries 36 open violations of these same two rules. Filing the baseline refresh separately; it affects every PR in the repo, not just this one.

Not fixing the two smells here — refactoring `setActiveCall` inside a ring-timer bugfix would mix concerns.

Separately, `Test compose (1)` failed on the first run (`testReconnectionDuringCallRecording` + `testUserRejectsTheOutgoingAudioCall`, both `findObject(...) must not be null`) and passed on re-run with no code change. The same shard failed the same morning on an unrelated branch, with the same reconnection test and an *incoming*-ring case, so it was the emulator window rather than this change.

aleksandar-apostolov:
Thanks — one of the two lands, the other I'd push back on.

**Isolating the `stopCallSound` assertion — fixed in 2c06f6b.** Correct, and worse than described: the assertion was fully vacuous. The initial `Idle` emission routes to `else -> stopCallSound()`, so I verified it by deleting the acceptance transition from the test entirely and it still passed. Now the repeated-outgoing phase is asserted first, recorded calls are dropped, and only then does the acceptance run — verified failing with the acceptance removed, so it actually bites.

**`RingingStateTest` extending `TestBase` — not taking this one.** It isn't the convention here: 5 of 133 test classes in `stream-video-android-core` extend it, and the neighbours this sits among (`ActiveStateGateTest`, `StreamVideoClientTest`, `MicrophoneManagerTest`, `SpeakerManagerTest`) don't. `TestBase` also brings an Android context via `IntegrationTestHelper`, a `StreamPeerConnectionFactory` mock and a dispatcher rule — `RingingStateTest` asserts `equals`/`hashCode` on a sealed interface and needs none of it. Extending it would add setup cost and an Android dependency to a pure-Kotlin test.

PratimMallick (approved):

PR #1778, opened 2026-08-24
Introduce VideoComponentFactory as the component override mechanism

andremion:
### Goal

Resolves [AND-762](https://linear.app/stream/issue/AND-762/introduce-videocomponentfactory-compose-component-factory).

Introduce `VideoComponentFactory` as the single component override mechanism for the Compose Video SDK, mirroring `ChatComponentFactory` from the Chat SDK so the customization model is the same across both products. Today customization is spread across per-composable lambda slots, the style layer and the activity delegate; the factory gives one place to override components globally.

Stacked on #1776 (AND-1417). The base will be retargeted to `develop` after that PR merges, so only the last commit is relevant for review.

### Implementation

- `VideoComponentFactory`: a public interface with 29 `@Composable` methods, each with a default implementation that renders the built-in component. It covers the components behind the existing lambda slots on `CallContent`, `ParticipantVideo`, `ControlActions`, `CallLobby`, `CallAppBar` and the ringing screens.
- `VideoComponentFactoryParams.kt`: one params holder class per method (29), so methods can grow without breaking overrides. This follows the chat convention, including the params classes from day one (chat had to retrofit them in AND-1119).
- `CompoundComponentFactory`: layers overrides on top of the current factory for a subtree. Note: chat's version uses `remember(keys)`, which treats the vararg array as a single identity-compared key. This PR uses `remember(currentComponentFactory, *keys)` instead; the same fix probably applies to the chat SDK.
- Wiring: `componentFactory` parameter on `VideoTheme`, a public `LocalComponentFactory`, and a `VideoTheme.componentFactory` accessor.
- `StreamCallActivityComposeDelegate` exposes an overridable `componentFactory` used by the screens it renders. Screen-level overrides stay on the delegate, component-level overrides live in the factory.
- Demo app sample: `DemoComponentFactory` replaces the inline reaction and participant-action lambdas that `CallScreen` passed in two places, and is installed on the demo activity delegate.

Compatibility:

- All existing lambda slots keep their signatures and now delegate to the factory by default, so the change is additive at the source level and the PR can land on `develop`.
- The new `VideoTheme` parameter changes its JVM signature, so consumers need a recompile against this version. The rest of the api dump diff is additions only.

Decisions that deviate from a literal 1:1 slot mapping:

- The `CallAppBar` leading/center/trailing slots are separate factory methods instead of nullable fields on `CallAppBarParams`, because a null field could not distinguish "hide" from "use the default".
- `ParticipantLabel`'s `soundIndicatorContent` slot is not covered, because its two overloads have different default paddings, so one factory method could not serve both without a visual change.
- The lobby controls method does not delegate to the overridable `ControlActions` method on purpose, so overriding the in-call bar does not change the pre-join lobby bar.

Out of scope, per the ticket ("Initial method coverage: the slots that already exist as lambdas"): internal render paths that never had lambda slots still call components directly, so a factory override of `ParticipantVideo` applies to the main grid but not inside the default floating self tile, the default picture-in-picture content or the screen share renderer, and `AudioCallContent` has no factory methods yet. Tracked in the follow-up ticket [AND-1441](https://linear.app/stream/issue/AND-1441/route-the-remaining-internal-render-paths-through).

### 🎨 UI Changes

No visual changes. This is a refactor of how the default slot content is resolved; the snapshot suite must stay identical and it does (no golden regenerated).

### Testing

- `:stream-video-android-ui-compose:verifyPaparazziDebug` passes with no golden regenerated, which is the acceptance criterion for identical rendering of all default paths.
- `apiDump` updated and `apiCheck` passes.
- Debug and release compilation of the SDK module, plus the demo app.
- Manual check in the demo app: `DemoComponentFactory` drives the custom reaction and the Kick participant action through the factory instead of inline lambdas, on the grid tiles and now also on the floating self tile (Kick excludes the local participant).

To test an override without the demo app, pass a factory to the theme:

```kotlin
VideoTheme(
    componentFactory = object : VideoComponentFactory {
        @Composable
        override fun RowScope.CallAppBarTrailingContent(params: CallAppBarTrailingContentParams) {
            // Custom trailing content
        }
    },
) {
    CallContent(call = call)
}
```

## Summary by CodeRabbit

* **New Features**
  * Added customizable call UI components across app bars, active calls, lobbies, participant video, and incoming/outgoing call screens.
  * Added theme-level controls for replacing or extending default video-call interface elements.
  * Added participant reaction customization, including placement and display duration.
  * Added a capability-gated “Kick” participant action in the demo app.
  * Added customizable participant labels, connection indicators, overlays, captions, moderation UI, and call controls.
* **Improvements**
  * Updated the demo call screen to use the new customizable components.


andremion:
@coderabbitai review

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactory.kt
```diff
@@ -0,0 +1,553 @@
...
+
+    /**
+     * Content drawn over the video content of the active call, excluding the controls.
+     * Empty by default.
+     *
+     * @param params Parameters for this component.
+     */
+    @Composable
+    public fun CallContentVideoOverlayContent(params: CallContentVideoOverlayContentParams) {
+    }
```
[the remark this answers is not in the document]
andremion: The KDoc on each of these methods already documents that they are empty by default and exist as override points, so an inline comment would duplicate it. The chat SDK ships the same empty extension points in ChatComponentFactory without inline comments, and SonarCloud has the same open S1186 findings there (for example ChannelListHelperContent and MessageSpacer) with no gate impact. Keeping the bodies empty on purpose; if we want the report clean, the better fix is marking the findings as won't fix in SonarCloud, in both repos.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/CompoundComponentFactory.kt
```diff
@@ -0,0 +1,57 @@
...
+ * @param content The content that uses the compound factory.
+ */
+@Composable
+public fun CompoundComponentFactory(
+    vararg keys: Any?,
+    factory: (currentComponentFactory: VideoComponentFactory) -> VideoComponentFactory,
+    content: @Composable () -> Unit,
+) {
+    val currentComponentFactory = LocalComponentFactory.current
+    val compoundComponentFactory = remember(currentComponentFactory, *keys) {
```
gpunto: `factory` isn't in the key list, so a factory lambda that captures a composable parameter is remembered past the change: `currentComponentFactory` never changes, `keys` is empty by default, and the anonymous object keeps whatever it captured on first composition.

Adding `factory` as a key closes that without giving up memoisation: the Compose compiler memoises a non-composable lambda argument on its captures, so the instance stays stable across recompositions and changes exactly when a capture does.

For reference, the chat equivalent keys on the `keys` array itself (`remember(keys) { ... }`), so its memoisation hangs on that array's identity rather than on the factory. This version already tightened that by keying on `currentComponentFactory` and spreading `keys` — `factory` is the one key still missing.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactory.kt
```diff
@@ -0,0 +1,553 @@
...
+
+    /**
+     * A single participant video with a label, connection indicator, reactions and actions.
+     * The default implementation renders
+     * [io.getstream.video.android.compose.ui.components.call.renderer.ParticipantVideo].
+     *
+     * @param params Parameters for this component.
+     */
+    @Composable
+    public fun ParticipantVideo(params: ParticipantVideoParams) {
```
gpunto: Non-blocking, but worth closing before release: two tiles on the same screen ignore this override. `FloatingParticipantVideo`'s `videoRenderer` default and `DefaultPictureInPictureContent` both call the top-level `ParticipantVideo` directly rather than `VideoTheme.componentFactory.ParticipantVideo`, so overriding this restyles the grid tiles but not the local floating tile or the PiP tile. The inner slots (label/reaction/actions) *do* route through the factory, so the result is a partial override rather than a clean no-op.

Same pattern in the `videoRenderer` defaults of `ParticipantsLayout` and `ParticipantsScreenSharing`, though `CallContent` forwards past those.

stream-video-android-ui-compose/src/testDebug/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactoryTest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    fun `factory incoming call content`() {
+        snapshot {
+            VideoComponentFactoryIncomingCallPreview()
+        }
+    }
+
+    @Test
+    fun `factory incoming call content in dark mode`() {
```
gpunto, line 84: Non-blocking: this golden is byte-identical to `factory incoming call content` (both md5 `6ef5603c…`), and the outgoing pair below is the same story (`7a6788e0…`). Both pairs also match the existing `CallContentTest_… with minimum parameters` goldens.

Across the commit, 8 of the 14 new PNGs duplicate another image byte-for-byte. The null-branch coverage these previews add is real, but a pixel assertion can't see it — worth dropping the redundant ones or saying in the comment what they actually pin.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactoryParams.kt
```diff
@@ -0,0 +1,464 @@
...
+    val participant: ParticipantState,
+    val labelPosition: Alignment = Alignment.BottomStart,
+)
+
+/**
+ * Parameters for [VideoComponentFactory.ParticipantVideoConnectionIndicatorContent].
+ *
+ * @param networkQuality The network quality of the participant connection.
+ */
+public data class ParticipantVideoConnectionIndicatorContentParams(
```
gpunto: Non-blocking: only `networkQuality` here, so an override can't tell which participant the indicator belongs to.

Related gaps nearby: `ParticipantVideoLabelContentParams` exposes only the derived `labelPosition` and not `style`, so `isShowingParticipantLabel`/`isScreenSharing` are unreachable; `ParticipantVideoReactionContentParams` has no `call`; and `ParticipantVideoActionsContentParams` is the one params class that doesn't lead with `call`.

These are public `data class`es, so adding a parameter later breaks `componentN`/`copy$default` — cheaper to settle the field sets now.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactory.kt
```diff
@@ -0,0 +1,553 @@
...
+import io.getstream.video.android.core.StreamVideo
+import io.getstream.video.android.core.call.state.LeaveCall
+import io.getstream.video.android.core.notifications.internal.service.CallServiceConfig
+import io.getstream.video.android.ui.common.R
+
+/**
+ * Default implementation of [VideoComponentFactory]. A singleton, so providing it as a default
+ * never invalidates the composition locals that carry the factory.
+ */
+internal object DefaultVideoComponentFactory : VideoComponentFactory
```
gpunto: Non-blocking: this is `internal` while `LocalComponentFactory` is `public`, so an app can provide the local but can't name the SDK default, and can't pass an explicit "use the defaults" value to `VideoTheme` or `StreamCallActivityComposeDelegate.componentFactory`. The workaround `object : VideoComponentFactory {}` allocates a fresh non-singleton, which loses exactly the property this KDoc advertises. A public `VideoComponentFactory.Default` would close it.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactory.kt
```diff
@@ -0,0 +1,553 @@
...
+/**
+ * A factory for creating the components used by the Video Compose SDK.
+ *
+ * Each method has a default implementation that renders the built-in component, so you only need
+ * to override the components you want to customize. The factory is provided through
+ * [VideoTheme], and the built-in components read it via [VideoTheme.componentFactory].
+ *
+ * ```
+ * VideoTheme(
+ *     componentFactory = object : VideoComponentFactory {
```
gpunto: Non-blocking: an anonymous object allocated at the call site is a new instance on every recomposition of the caller, and `LocalComponentFactory` is a `compositionLocalOf` with structural equality — so this example invalidates every factory reader in the subtree. `DemoComponentFactory` gets it right; suggest the example hoist to a top-level `object` (or a `remember`ed instance).

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactoryParams.kt
```diff
@@ -0,0 +1,464 @@
...
+    val onBackPressed: () -> Unit = {},
+)
+
+/**
+ * Parameters for [VideoComponentFactory.CallAppBarCenterContent].
+ *
+ * @param call The call that contains all the participants state and tracks.
+ * @param title The title shown in the app bar. When `null`, the default title is used.
+ */
+public data class CallAppBarCenterContentParams(
```
gpunto, line 74: Non-blocking: `title` is nullable and documented as "when `null`, the default title is used", but no SDK path can pass null — `CallAppBar`'s own `title` defaults to a non-null `stringResource` and forwards it, and the factory's `CallAppBar` already resolves `params.title ?: stringResource(...)` before calling it. Could be non-null, dropping one of the two fallbacks.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/CompoundComponentFactory.kt
andremion: Good catch, and the memoization detail makes it the right key set. Fixed in 47716e665fac2748102c24d13470e8247f2d0019.

stream-video-android-ui-compose/src/testDebug/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactoryTest.kt
andremion: That identity is actually what these pin: the defaults delegate to the same built-ins with the same preview data, so the images have to match, and the value is the coverage of the default bodies. The dark siblings match everywhere in the suite until the light palette lands with AND-1419, so I kept the pairs to let them diverge with no test changes. Added a note on the test class in 47716e665fac2748102c24d13470e8247f2d0019.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactoryParams.kt
andremion: The nullable is for direct callers of the factory method, like CallAppBarParams(call) with no title: a data class default can't be stringResource(...) since default expressions aren't composable, so null meaning default is the only way they get the localized title. The built-in CallAppBar keeps its own default because it's a separate entry point. Happy to revisit if you see a better shape.

gpunto (approved):

PR #1781, opened 2026-08-25
Fix the flaky MockK Context setup in PictureInPictureTest

andremion:
### Goal

Fix the flaky `PictureInPictureTest > should enter pip mode with correct params` test. It fails intermittently on CI during `setup` with:

```
io.mockk.MockKException: Can't instantiate proxy for class android.content.Context
Caused by: io.mockk.proxy.MockKAgentException: Value for this result is not assigned
```

The failure is a race in MockK's inline agent when it proxies platform classes. The test shares its JVM with the Paparazzi snapshot tests, which append to the bootstrap classpath, and that combination makes the agent attach unreliable on the Linux runners. Seen on PR #1776, where the diff was test-only.

Resolves [AND-1446](https://linear.app/stream/issue/AND-1446/fix-the-mockk-context-flake-in-pictureinpicturetest).

### Implementation

The test class already runs with Robolectric, so the platform classes do not need to be mocked at all:

- Replaced `mockk<Context>()` with a real Robolectric `Activity` (`Robolectric.buildActivity(Activity::class.java).create().get()`). Since the context is itself an `Activity`, `findActivity()` resolves it naturally, so the `mockkStatic("...PictureInPictureKt")` call is gone too.
- The PiP system feature is now controlled with `shadowOf(activity.packageManager).setSystemFeature(...)` instead of mocked `PackageManager` and `Resources`.
- The enter-PiP assertions now check `activity.isInPictureInPictureMode`. Robolectric's `ShadowActivity` sets that flag in both `enterPictureInPictureMode` overloads (verified in the Robolectric 4.11.1 shadow source).
- MockK is kept only for the domain classes (`Call`, `ScreenSharingSession`, `ParticipantState`). `tearDown` now only calls `clearAllMocks()`; the old `unmockkStatic(ContextCompat::class)` was unmocking something that was never mocked.

With no inline-agent proxying of `android.content.Context`, the race that caused the flake cannot happen anymore.

After review, the assertions were strengthened:

- The `getPictureInPictureParams` tests assert the built params values where the runtime allows it: `isAutoEnterEnabled` on S and above, plus aspect ratio and title on TIRAMISU (`getAspectRatio()` returned `float` until API 32, so it can only be asserted from 33).
- The configured auto-enter test covers both the enabled and the disabled case, so removing the `setAutoEnterEnabled` call in production fails the tests.
- The `getAspect` tests pass `Configuration.ORIENTATION_*`, the values production actually feeds, instead of `ActivityInfo.SCREEN_ORIENTATION_*`.

### 🎨 UI Changes

No UI changes, this is a test-only change.

### Testing

Run the test class locally:

```
./gradlew :stream-video-android-ui-compose:testDebugUnitTest --tests "io.getstream.video.android.compose.pip.PictureInPictureTest"
```

All 10 tests pass (0 failures, 0 skipped).

## Summary by CodeRabbit

* **Tests**
  * Improved Picture-in-Picture coverage using realistic device and activity behavior.
  * Added validation for supported, pre-Oreo, and unsupported-device scenarios.
  * Tests now verify the activity’s actual Picture-in-Picture state.

### DIFF
--- a/stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
+++ b/stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
@@ -137,6 +137,8 @@ public class PictureInPictureTest {
         val builder = getPictureInPictureParams(Rational(16, 9), pipConfig)
         val params = builder.build()
 
+        // The PictureInPictureParams getters are only public from API 33,
+        // so the params values can only be asserted in the TIRAMISU tests.
         assertNotNull(params)
     }
 
@@ -145,7 +147,21 @@ public class PictureInPictureTest {
     public fun `should set title and seamless resize for TIRAMISU`() {
         val builder = getPictureInPictureParams(Rational(9, 16), pipConfig)
         val params = builder.build()
-        assertNotNull(params)
+
+        assertEquals(Rational(9, 16), params.aspectRatio)
+        assertEquals("Video Player", params.title.toString())
+        assertTrue(params.isSeamlessResizeEnabled)
+        assertTrue(params.isAutoEnterEnabled)
+    }
+
+    @Test
+    @Config(sdk = [Build.VERSION_CODES.TIRAMISU])
+    public fun `should disable auto-enter when configuration disables it for TIRAMISU`() {
+        val config = PictureInPictureConfiguration(enable = true, autoEnterEnabled = false)
+
+        val params = getPictureInPictureParams(Rational(9, 16), config).build()
+
+        assertFalse(params.isAutoEnterEnabled)
     }
 
     @Test
--- a/stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
+++ b/stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
@@ -17,8 +17,8 @@
 package io.getstream.video.android.compose.pip
 
 import android.app.Activity
-import android.content.pm.ActivityInfo
 import android.content.pm.PackageManager
+import android.content.res.Configuration
 import android.os.Build
 import android.util.Rational
 import io.getstream.video.android.core.Call
@@ -66,7 +66,7 @@ public class PictureInPictureTest {
     // enterPictureInPicture Test
     @Test
     @Config(sdk = [Build.VERSION_CODES.TIRAMISU])
-    public fun `should enter pip mode with correct params`() {
+    public fun `should enter pip mode when O and above`() {
         val screenSharing = mockk<ScreenSharingSession>(relaxed = true)
         every { call.state.screenSharingSession.value } returns screenSharing
 
@@ -104,8 +104,8 @@ public class PictureInPictureTest {
         val screenSharing = mockk<ScreenSharingSession>()
         every { screenSharing.participant } returns localParticipant
 
-        val aspect1 = getAspect(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT, null)
-        val aspect2 = getAspect(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT, screenSharing)
+        val aspect1 = getAspect(Configuration.ORIENTATION_PORTRAIT, null)
+        val aspect2 = getAspect(Configuration.ORIENTATION_PORTRAIT, screenSharing)
 
         assertEquals(Rational(9, 16), aspect1)
         assertEquals(Rational(9, 16), aspect2)
@@ -113,7 +113,7 @@ public class PictureInPictureTest {
 
     @Test
     public fun `should return 16x9 when landscape`() {
-        val aspect = getAspect(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE, null)
+        val aspect = getAspect(Configuration.ORIENTATION_LANDSCAPE, null)
         assertEquals(Rational(16, 9), aspect)
     }
 
@@ -125,7 +125,7 @@ public class PictureInPictureTest {
         val screenSharing = mockk<ScreenSharingSession>()
         every { screenSharing.participant } returns remoteParticipant
 
-        val aspect = getAspect(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT, screenSharing)
+        val aspect = getAspect(Configuration.ORIENTATION_PORTRAIT, screenSharing)
         assertEquals(Rational(16, 9), aspect)
     }
 
@@ -133,35 +133,43 @@ public class PictureInPictureTest {
 
     @Test
     @Config(sdk = [Build.VERSION_CODES.S]) // Android 12
-    public fun `should set aspect ratio and auto-enter when SDK S`() {
+    public fun `should enable auto-enter by default when SDK S`() {
         val builder = getPictureInPictureParams(Rational(16, 9), pipConfig)
         val params = builder.build()
 
-        // The PictureInPictureParams getters are only public from API 33,
-        // so the params values can only be asserted in the TIRAMISU tests.
-        assertNotNull(params)
-    }
-
-    @Test
-    @Config(sdk = [Build.VERSION_CODES.TIRAMISU])
-    public fun `should set title and seamless resize for TIRAMISU`() {
-        val builder = getPictureInPictureParams(Rational(9, 16), pipConfig)
-        val params = builder.build()
-
-        assertEquals(Rational(9, 16), params.aspectRatio)
-        assertEquals("Video Player", params.title.toString())
-        assertTrue(params.isSeamlessResizeEnabled)
+        // getAspectRatio() returns float until API 32 and Rational from 33, so calling it
+        // against the current SDK stubs throws NoSuchMethodError on the S runtime.
+        // Only the boolean getters can be asserted here.
         assertTrue(params.isAutoEnterEnabled)
     }
 
     @Test
     @Config(sdk = [Build.VERSION_CODES.TIRAMISU])
-    public fun `should disable auto-enter when configuration disables it for TIRAMISU`() {
-        val config = PictureInPictureConfiguration(enable = true, autoEnterEnabled = false)
+    public fun `should set aspect ratio and title for TIRAMISU`() {
+        val builder = getPictureInPictureParams(Rational(9, 16), pipConfig)
+        val params = builder.build()
 
-        val params = getPictureInPictureParams(Rational(9, 16), config).build()
+        assertEquals(Rational(9, 16), params.aspectRatio)
+        assertEquals("Video Player", params.title.toString())
+        // isSeamlessResizeEnabled() defaults to true when never set, so this only guards
+        // a regression to setSeamlessResizeEnabled(false); set-true is not observable.
+        assertTrue(params.isSeamlessResizeEnabled)
+    }
 
-        assertFalse(params.isAutoEnterEnabled)
+    @Test
+    @Config(sdk = [Build.VERSION_CODES.TIRAMISU])
+    public fun `should follow the configured auto-enter for TIRAMISU`() {
+        val enabled = getPictureInPictureParams(
+            Rational(9, 16),
+            PictureInPictureConfiguration(enable = true, autoEnterEnabled = true),
+        ).build()
+        val disabled = getPictureInPictureParams(
+            Rational(9, 16),
+            PictureInPictureConfiguration(enable = true, autoEnterEnabled = false),
+        ).build()
+
+        assertTrue(enabled.isAutoEnterEnabled)
+        assertFalse(disabled.isAutoEnterEnabled)
     }
 
     @Test
--- a/stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
+++ b/stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
@@ -17,13 +17,10 @@
 package io.getstream.video.android.compose.pip
 
 import android.app.Activity
-import android.content.Context
 import android.content.pm.ActivityInfo
 import android.content.pm.PackageManager
-import android.content.res.Configuration
 import android.os.Build
 import android.util.Rational
-import androidx.core.content.ContextCompat
 import io.getstream.video.android.core.Call
 import io.getstream.video.android.core.ParticipantState
 import io.getstream.video.android.core.model.ScreenSharingSession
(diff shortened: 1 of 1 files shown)


andremion:
@coderabbitai review

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
```diff
@@ -17,64 +17,49 @@
...
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.robolectric.Robolectric
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.Shadows.shadowOf
 import org.robolectric.annotation.Config
 
 @RunWith(RobolectricTestRunner::class)
 public class PictureInPictureTest {
```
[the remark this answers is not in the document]
andremion: TestBase is not reachable from this module: it lives in the core module's own test source set (stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/base/TestBase.kt) and test sources are not shared between Gradle modules. It would also be a bad fit here: its init block makes a network call (StreamService.instance.getAuthData against the pronto environment) and it mocks StreamPeerConnectionFactory, so extending it would add heavy setup and a new source of flakiness to the exact test this PR is stabilizing. No test in the compose module uses it. Keeping this test standalone.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
```diff
@@ -84,35 +69,29 @@ public class PictureInPictureTest {
     public fun `should enter pip mode with correct params`() {
         val screenSharing = mockk<ScreenSharingSession>(relaxed = true)
         every { call.state.screenSharingSession.value } returns screenSharing
-        val activity = context.findActivity()!!
 
-        enterPictureInPicture(context, call, pipConfig)
+        enterPictureInPicture(activity, call, pipConfig)
 
-        verify(exactly = 1) { activity.enterPictureInPictureMode(any()) }
+        assertTrue(activity.isInPictureInPictureMode)
```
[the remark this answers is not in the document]
andremion: Applied where the platform allows it. Activity.getPictureInPictureParams() does not exist in the Activity API, and Robolectric's ShadowActivity does not store the params passed to enterPictureInPictureMode, so the entry test keeps the isInPictureInPictureMode assertion. For the builder helper tests: the PictureInPictureParams getters (getAspectRatio, isAutoEnterEnabled, getTitle, isSeamlessResizeEnabled) are only public from API 33. I tried them on the S runtime and they fail with NoSuchMethodError. So the TIRAMISU tests now assert the real values (aspect ratio, title, seamless resize and auto-enter), plus a new case where the configuration disables auto-enter. The S and O tests keep the non-null assertion, with a comment explaining why.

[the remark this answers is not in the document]
andremion: it's pushed

gpunto: This test is named "with correct params" but no longer asserts anything about the params, and it's now strictly weaker than the MockK version it replaced.

I decompiled `ShadowActivity` from `shadows-framework-4.11.1.jar`: both `enterPictureInPictureMode()` overloads have identical bodies — `putfield isInPictureInPictureMode = true` — and the `PictureInPictureParams` argument is discarded outright. So `assertTrue(activity.isInPictureInPictureMode)` can't distinguish the two branches, and the shadow keeps no record of the params to assert against.

Confirmed by mutation: if I change the `Build.VERSION.SDK_INT >= Build.VERSION_CODES.O` check in `enterPictureInPicture` to `if (false)`, so the params path is never taken at all, all 10 tests still pass. The old `verify(exactly = 1) { activity.enterPictureInPictureMode(any()) }` would have caught that.

Net effect is that the SDK branching in `enterPictureInPicture` is now uncovered, and this test overlaps completely with `should enter pip mode without params when below O` below. Worth either reaching for the params some other way, or renaming the test so it doesn't claim coverage it no longer has.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
```diff
@@ -166,7 +147,21 @@ public class PictureInPictureTest {
     public fun `should set title and seamless resize for TIRAMISU`() {
         val builder = getPictureInPictureParams(Rational(9, 16), pipConfig)
         val params = builder.build()
-        assertNotNull(params)
+
+        assertEquals(Rational(9, 16), params.aspectRatio)
+        assertEquals("Video Player", params.title.toString())
+        assertTrue(params.isSeamlessResizeEnabled)
```
gpunto: This assertion can't fail. In the exact jar Robolectric loads for `sdk = TIRAMISU` (`android-all-instrumented-13-robolectric-9030017-i4`), `isSeamlessResizeEnabled()` is `mSeamlessResizeEnabled == null ? true : mSeamlessResizeEnabled` — it defaults to `true` when never set.

Confirmed by mutation: deleting `setSeamlessResizeEnabled(true)` from `getPictureInPictureParams` leaves all 10 tests green, so the "seamless resize" half of this test's name has zero coverage behind it.

No public getter distinguishes set-true from unset, so the options are asserting via `params.toString()` (it prints `isSeamlessResizeEnabled=`), or just dropping the claim from the test name. Worth noting the other three assertions in this test are all genuinely load-bearing — `aspectRatio` and `title` default to null, and `isAutoEnterEnabled` defaults to `false`, so those three do real work.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
```diff
@@ -166,7 +147,21 @@ public class PictureInPictureTest {
...
+    }
+
+    @Test
+    @Config(sdk = [Build.VERSION_CODES.TIRAMISU])
+    public fun `should disable auto-enter when configuration disables it for TIRAMISU`() {
+        val config = PictureInPictureConfiguration(enable = true, autoEnterEnabled = false)
+
+        val params = getPictureInPictureParams(Rational(9, 16), config).build()
+
+        assertFalse(params.isAutoEnterEnabled)
```
gpunto: Non-blocking, since this test does earn its keep — but it's only half a test. `isAutoEnterEnabled()` is `mAutoEnterEnabled == null ? false : ...`, so `assertFalse` passes both when auto-enter is correctly disabled *and* when `setAutoEnterEnabled` is never called at all. I verified this: removing the `setAutoEnterEnabled(defaultAutoEnterEnabled)` call from production fails only `should set title and seamless resize for TIRAMISU`, while this test stays green.

That said, it does catch the regression it's aimed at — dropping the `SDK_INT >= TIRAMISU` branch that overrides `defaultAutoEnterEnabled` makes this test fail, which I also confirmed by mutation. So it's genuinely useful; adding a positive case in the same test (assert `true` for `autoEnterEnabled = true`) would just close the "was it set at all" gap.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
```diff
@@ -158,6 +137,8 @@ public class PictureInPictureTest {
         val builder = getPictureInPictureParams(Rational(16, 9), pipConfig)
         val params = builder.build()
 
+        // The PictureInPictureParams getters are only public from API 33,
+        // so the params values can only be asserted in the TIRAMISU tests.
         assertNotNull(params)
```
gpunto: Non-blocking, and pre-existing rather than introduced here. `Builder.build()` never returns null, so `assertNotNull(params)` is unfalsifiable and this test asserts nothing at all — same for `should only set aspect ratio for Oreo` at line 172. Both stay green against a `getPictureInPictureParams` whose entire body is `return PictureInPictureParams.Builder()`.

The new comment is accurate and useful — the typed getters really are API 33+ — but it currently documents *why* a dead assertion is dead rather than replacing it. `params.toString()` does expose `aspectRatio=` and `isAutoPipEnabled=` on every API level, if it's worth making these two tests real while the file is already being reworked.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto: Agreed with your earlier reply that `ShadowActivity` can't expose the entered params — I decompiled 4.11.1 and both `enterPictureInPictureMode` overloads have identical bodies that discard the argument. So this assertion is the only option here.

The part that's still worth acting on is the name. Since neither overload is distinguishable, changing the `SDK_INT >= O` check in `enterPictureInPicture` to `if (false)` keeps all 10 tests green — the SDK branch is uncovered, and this test is now identical in power to `should enter pip mode without params when below O`. Renaming it (it doesn't check params) would stop it overstating what it covers.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto: This comment is a bit broader than the platform actually is, and it's costing you a real assertion. Only `getAspectRatio()` and `getTitle()` are 33+. `isAutoEnterEnabled()` and `isSeamlessResizeEnabled()` exist on API 31 with identical signatures — the `NoSuchMethodError` you hit is specific to `getAspectRatio()`, which was `float getAspectRatio()` through API 31 and only became `Rational` in 33.

So `assertTrue(params.isAutoEnterEnabled)` runs fine here. I added it locally against the S runtime and the test passed, no `NoSuchMethodError`. That's exactly what this test's name promises, and right now `assertNotNull(builder.build())` can never fail.

The Oreo test at line 172 genuinely has nothing available (no `toString()` override, no matching getters), so `assertNotNull` there is fair enough as a smoke test.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto: This one was added to close the earlier thread, but it can't fail: `isSeamlessResizeEnabled()` is `mSeamlessResizeEnabled == null ? true : ...`, so it's `true` when never set. Deleting `setSeamlessResizeEnabled(true)` from production leaves all 10 tests green.

No public getter distinguishes set-true from unset, so it's either `params.toString()` (it prints `isSeamlessResizeEnabled=` from API 31 on) or dropping the claim from the test name. The other three assertions here are load-bearing — aspect and title default to null, auto-enter to `false`.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto: Non-blocking. `isAutoEnterEnabled()` defaults to `false`, so this also passes when `setAutoEnterEnabled` is never called — removing that call fails only the test above, not this one.

It does catch what it's aimed at (dropping the `SDK_INT >= TIRAMISU` override makes it fail), so it's useful as-is; a positive case in the same test would close the gap.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto, line 75: Agreed with your earlier reply that `ShadowActivity` can't expose the entered params — I decompiled 4.11.1 and both `enterPictureInPictureMode` overloads have identical bodies that discard the argument. So this assertion is the only option here.

The part that's still worth acting on is the name. Since neither overload is distinguishable, changing the `SDK_INT >= O` check in `enterPictureInPicture` to `if (false)` keeps all 10 tests green — the SDK branch is uncovered, and this test is now identical in power to `should enter pip mode without params when below O`. Renaming it (it doesn't check params) would stop it overstating what it covers.

gpunto (commented):
Flake fix is right and I'd land it — `@Before` mocked the whole `PictureInPictureKt` file class while `@After` unmocked `ContextCompat::class`, so the instrumentation leaked across the JVM fork. Ran the class 5x plus the full module suite, all green.

My comments are about the assertions, not the fix, and build on the resolved thread rather than reopening it. All found by mutation: gutting `getPictureInPictureParams` to `return PictureInPictureParams.Builder()` leaves 9 of the 10 tests green.

Two non-blocking notes, on lines outside the diff so I couldn't inline them:
- Pre-existing: the `getAspect` tests pass `ActivityInfo.SCREEN_ORIENTATION_*`, but production feeds `Configuration.ORIENTATION_*`. Portrait matches by coincidence (both `1`); real landscape (`2`) is never tested. Same behaviour either way, just the wrong contract.
- Description says "All 9 tests pass"; the second commit made it 10.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto: This comment is a bit broader than the platform actually is, and it's costing you a real assertion. Only `getAspectRatio()` and `getTitle()` are 33+. `isAutoEnterEnabled()` and `isSeamlessResizeEnabled()` exist on API 31 with identical signatures — the `NoSuchMethodError` you hit is specific to `getAspectRatio()`, which was `float getAspectRatio()` through API 31 and only became `Rational` in 33.

So `assertTrue(params.isAutoEnterEnabled)` runs fine here. I added it locally against the S runtime and the test passed, no `NoSuchMethodError`. That's exactly what this test's name promises, and right now `assertNotNull(builder.build())` can never fail.

The Oreo test at line 172 genuinely has nothing available (no `toString()` override, no matching getters), so `assertNotNull` there is fair enough as a smoke test.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto, line 156: This one was added to close the earlier thread, but it can't fail: `isSeamlessResizeEnabled()` is `mSeamlessResizeEnabled == null ? true : ...`, so it's `true` when never set. Deleting `setSeamlessResizeEnabled(true)` from production leaves all 10 tests green.

No public getter distinguishes set-true from unset, so it's either `params.toString()` (it prints `isSeamlessResizeEnabled=` from API 31 on) or dropping the claim from the test name. The other three assertions here are load-bearing — aspect and title default to null, auto-enter to `false`.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
gpunto: Non-blocking. `isAutoEnterEnabled()` defaults to `false`, so this also passes when `setAutoEnterEnabled` is never called — removing that call fails only the test above, not this one.

It does catch what it's aimed at (dropping the `SDK_INT >= TIRAMISU` override makes it fail), so it's useful as-is; a positive case in the same test would close the gap.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
andremion: Fair point. Renamed it to `should enter pip mode when O and above`, so it no longer claims to check the params.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
andremion: You are right, thanks. I checked the API 31/32 runtime jar and getAspectRatio() returns float there (it only became Rational in 33), so the NoSuchMethodError was specific to that getter while the boolean getters match. The S test now asserts isAutoEnterEnabled, and the comment explains the float vs Rational detail instead of the broader claim.

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
andremion: Good catch. I also tried the toString() option, but it does not work either: toString() prints the value from the getter, not the raw field, so the set-true and unset params produce identical strings on both S and TIRAMISU (verified with a probe test). Since the difference is not observable at all, I took the other option: dropped the claim from the test name and kept the assertion with a comment saying it only guards a regression to setSeamlessResizeEnabled(false).

stream-video-android-ui-compose/src/test/kotlin/io/getstream/video/android/compose/pip/PictureInPictureTest.kt
andremion: Done. Reworked it into a single test that builds the params with the configuration enabled and disabled and asserts both, so removing the setAutoEnterEnabled call in production now fails it. With that, your original mutation (gutting getPictureInPictureParams to a bare builder) fails 3 tests instead of 1.

andremion:
Thanks for the mutation pass, all four comments are addressed in 308ca0cc8a. Also acted on the two notes from the review body: the getAspect tests now pass Configuration.ORIENTATION_* (production still compares against ActivityInfo.SCREEN_ORIENTATION_PORTRAIT, which has the same value, so behavior is unchanged; happy to fix that contract in a follow-up), and the description now says 10 tests.

gpunto (approved):

PR #1782, opened 2026-08-25
Stabilize the flaky E2E emulator tests

andremion:
### Goal

Fix AND-1445. The nightly `e2e-test-cron` was red on 11 of the last 13 runs, and PR batches fail intermittently with a rotating set of tests. Almost all recent nightly failures came from two tests: `ReconnectionTests#testReconnectionDuringCallRecording` and `RingingTests#testUserRejectsTheOutgoingAudioCall`.

### Implementation

Every reported failure means 3 consecutive `RetryRule` attempts failed. The main root cause is that the attempts were not independent. The instrumentation runs inside the app process (the test APK targets the app package), so the app cannot be killed between attempts. A failed attempt could leave the internet connection disabled (reconnection tests) or a call still active or ringing, and the next attempts inherited that state and failed the same way.

- `RetryRule` now recovers between attempts: it re-enables the internet connection and leaves any leftover active or ringing call through `StreamVideo.instanceOrNull()`. All post-failure steps (database cleanup, Allure attachments, attempt result) are now best-effort, so an error there can no longer replace the real failure or skip the remaining retries.
- The UiAutomator wait helpers are replaced with the polling, stale-safe versions from stream-chat-android (PRs 6383, 6570, 6588, 6594, 6603 there). `waitToAppear` polls and absorbs `StaleObjectException`, and a timeout now throws a clear error naming the selector and timeout instead of the misleading `findObject(...) must not be null` NPE seen in the nightly logs. All `UserRobot` clicks go through the new `waitToAppearAndClick`, which re-finds and re-clicks when the node goes stale.
- Wider wait windows on the paths that timed out in CI: joining a call (`waitForCallToStart`, 15s to 30s) and the outgoing ringing screen (20s to 30s, it failed mostly on API 28). The recording label assertion now polls, because the label briefly reads "Reconnecting.." after a network drop.
- `run_e2e_test` accepts a validated `test_class` option, and the PR workflow gets `api_level` and `test_class` dispatch inputs. Dispatching with a test class makes every batch job run only that test, giving independent samples of one flaky test on a chosen API level.
- Failure artifacts now include `fastlane/allure-results`, because TestOps deduplicates same-historyId results and the failed-attempt attachments disappear when the test passes elsewhere.
- The E2E concurrency groups are prefixed with the workflow name so different workflows on the same ref do not cancel each other.

Not covered here, tracked as follow-ups in AND-1445: a few nightly failures are app process crashes (the crash files are only on the device), and the buddy server has no status endpoints for explicit sync barriers.

### 🎨 UI Changes

Not applicable, test infrastructure only.

### Testing

Both top flaky tests pass locally on an API 35 emulator through the real fastlane flow (orchestrator, Allure runner, video-buddy participant), using the new `test_class` option:

- `ReconnectionTests#testReconnectionDuringCallRecording`: OK in 43s, including the network drop and restore.
- `RingingTests#testUserRejectsTheOutgoingAudioCall`: OK in 16.5s.

Since a flaky test passing twice is not proof by itself, the new workflow dispatch inputs are the follow-up verification tool: run the E2E workflow with `test_class` on the API levels where each test failed (32-34 for the reconnection test, 28 for the ringing one) to get repeated independent samples.

## Summary by CodeRabbit

* **New Features**
  * Added optional workflow inputs for selecting an Android API level and running a specific end-to-end test class.
  * Test runs now upload additional Allure results for improved reporting.

* **Bug Fixes**
  * Improved UI test stability with more reliable waits, clicks, retries, and call-state cleanup.
  * Increased select test timeouts and preserved original failures when recovery steps encounter errors.

* **Testing**
  * Enhanced support for targeted test execution and more resilient end-to-end test assertions.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -85,13 +85,17 @@ jobs:
       - name: Run tests
         uses: reactivecircus/android-emulator-runner@a421e43855164a8197daf9d8d40fe71c6996bb0d
         timeout-minutes: 45
+        env:
+          # Passed through the environment and expanded quoted, so the value cannot
+          # inject shell commands into the script line; the Fastfile validates it too.
+          TEST_CLASS: ${{ inputs.test_class }}
         with:
           api-level: ${{ env.ANDROID_API_LEVEL }}
           disable-animations: true
           profile: pixel
           arch : x86_64
           emulator-options: -no-snapshot-save -no-window -no-audio -no-boot-anim -gpu swiftshader_indirect -camera-back none -camera-front none
-          script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }} ${{ inputs.test_class && format('test_class:{0}', inputs.test_class) || '' }}
+          script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }} test_class:"$TEST_CLASS"
       - name: Allure TestOps Upload
         if: ${{ env.LAUNCH_ID != '' && (success() || failure()) }}
         run: bundle exec fastlane allure_upload
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -4,9 +4,19 @@ on:
   pull_request:
 
   workflow_dispatch:
+    inputs:
+      api_level:
+        description: 'Android API level to run on'
+        type: string
+        required: false
+        default: '34'
+      test_class:
+        description: 'Fully qualified test class to run on its own, optionally with #method. Every batch job then runs it, giving independent samples of a flaky test.'
+        type: string
+        required: false
 
 concurrency:
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 env:
@@ -61,7 +71,7 @@ jobs:
           - batch: 2
       fail-fast: false
     env:
-      ANDROID_API_LEVEL: 34
+      ANDROID_API_LEVEL: ${{ inputs.api_level || '34' }}
       LAUNCH_ID: ${{ needs.allure_testops_launch.outputs.launch_id }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
@@ -81,7 +91,7 @@ jobs:
           profile: pixel
           arch : x86_64
           emulator-options: -no-snapshot-save -no-window -no-audio -no-boot-anim -gpu swiftshader_indirect -camera-back none -camera-front none
-          script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }}
+          script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }} ${{ inputs.test_class && format('test_class:{0}', inputs.test_class) || '' }}
       - name: Allure TestOps Upload
         if: ${{ env.LAUNCH_ID != '' && (success() || failure()) }}
         run: bundle exec fastlane allure_upload
@@ -100,7 +110,11 @@ jobs:
         timeout-minutes: 10
         with:
           name: test_artifacts_${{ matrix.batch }}
+          # The allure results carry the retry rule's failure artifacts (logcat,
+          # screenshot, hierarchy per failed attempt), which TestOps deduplicates
+          # away when the same test passes in another batch.
           path: |
             fastlane/recordings/*
+            fastlane/allure-results/*
             fastlane/video-buddy-server.log
             fastlane/video-buddy-console.log
--- a/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
+++ b/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
@@ -32,11 +32,11 @@ import io.getstream.video.android.uiautomator.device
 import io.getstream.video.android.uiautomator.findObject
 import io.getstream.video.android.uiautomator.findObjects
 import io.getstream.video.android.uiautomator.isDisplayed
-import io.getstream.video.android.uiautomator.retryOnStaleObjectException
 import io.getstream.video.android.uiautomator.seconds
 import io.getstream.video.android.uiautomator.typeText
 import io.getstream.video.android.uiautomator.waitForText
 import io.getstream.video.android.uiautomator.waitToAppear
+import io.getstream.video.android.uiautomator.waitToAppearAndClick
 
 class UserRobot {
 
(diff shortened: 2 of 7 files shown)


andremion:
@coderabbitai review

.github/workflows/e2e-test.yml
```diff
@@ -81,7 +91,7 @@ jobs:
           profile: pixel
           arch : x86_64
           emulator-options: -no-snapshot-save -no-window -no-audio -no-boot-anim -gpu swiftshader_indirect -camera-back none -camera-front none
-          script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }}
+          script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }} ${{ inputs.test_class && format('test_class:{0}', inputs.test_class) || '' }}
```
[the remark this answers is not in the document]
andremion: Fixed in b555ea598c. The input now goes through the step environment as TEST_CLASS and is expanded quoted in the script line, so the value reaches fastlane as plain text. The Fastfile regex stays as a second layer.

fastlane/Fastfile
```diff
@@ -75,16 +75,29 @@ lane :run_e2e_test do |options|
...
+      batch_tests(
+        batch: options[:batch],
+        batch_count: options[:batch_count],
+        test_apk_path: stream_test_path
+      )
+    else
+      # Targeted run: a single class and no batching, so every batch job runs it and gives an
+      # independent sample. Used to reproduce a failure that only shows on one API level.
+      "-e class #{targeted_class}"
+    end
```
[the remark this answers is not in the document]
andremion: Fixed in b555ea598c by removing '$' from the allowed characters instead of escaping. All E2E test classes here are top-level, so nested class support was theoretical, and escaping would still leave the device shell expansion after adb.

.github/workflows/e2e-test.yml
```diff
@@ -4,9 +4,19 @@ on:
   pull_request:
 
   workflow_dispatch:
+    inputs:
+      api_level:
+        description: 'Android API level to run on'
+        type: string
+        required: false
+        default: '34'
```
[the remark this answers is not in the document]
andremion: Keeping type: string on purpose, to stay aligned with the same workflow in stream-chat-android. The value never reaches a shell, it only feeds the emulator action's api-level input, so the worst case for a wrong value is a failed SDK provisioning step in that dispatched run. A choice list would also duplicate the supported levels and drift from the cron matrix.

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/rules/RetryRule.kt
```diff
@@ -71,25 +73,45 @@ public class RetryRule(private val count: Int) : TestRule {
...
+                            )
+                        }
+                    }.onFailure {
+                        System.err.println(
+                            "$testName: capturing failure artifacts failed: $it",
                         )
                     }
                     if (attempt < count) {
-                        writeFailedAttemptResult(attempt, t, startMillis)
+                        recoverForNextAttempt(testName)
```
gpunto: This sits in the pre-existing `if (attempt < count)` block, which is the right guard for `writeFailedAttemptResult` (on the final attempt the running result *is* the real one) but the wrong guard for the connection restore — so the **last** failed attempt never recovers.

`enableInternetConnection()` runs `svc data enable` / `svc wifi enable`, i.e. device-global state that outlives the test process; `clearPackageData` only wipes app data. All three tests in `ReconnectionTests` call `disableInternetConnection()` in one `step` and `enableInternetConnection()` in the *next*, and there's no `@After` anywhere in the source set. With `RetryRule(count = 3)`, a test that fails all 3 attempts inside that window — or where `enableInternetConnection()` itself throws `IllegalStateException("There is no internet connection.")` after `waitForInternetConnection`'s 10s timeout — leaves wifi and data off, and the remaining network-dependent tests in the batch then fail for an unrelated reason. One flaky test takes out the batch, which is the opposite of the goal.

Moving the connection restore out of the `if` so it always runs should do it; leaving the call teardown gated is fine.

(Chat's `RetryRule` has no equivalent of this method, so it's genuinely new code rather than something the port brought over.)

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
```diff
@@ -353,30 +355,31 @@ class UserRobot {
...
 
     fun waitForParticipantsOnCall(count: Int = 1, timeOutMillis: Long = 100.seconds): UserRobot {
         val user = 1
         val participants = user + count
-        device.retryOnStaleObjectException {
-            CallPage.participantsCountBadge
-                .waitToAppear()
-                .waitForText(expectedText = participants.toString(), timeOutMillis = timeOutMillis)
-        }
+        CallPage.participantsCountBadge.waitForText(
```
gpunto: The return value is discarded, and the new `waitForText` never throws — it returns the last observed text, or `""` if the node was never found. Its own KDoc says "callers should wrap the result in an assertion to surface a mismatch or timeout".

I checked every call site in both repos: 2 in `UserRobotCallAsserts.kt` here and 15 in Chat, and all 17 wrap the result in an `assertEquals`. This is the only one that doesn't, so the helper silently cannot fail here.

`assertParticipantsCountOnCall` in `UserRobotCallAsserts.kt` is the same badge with the same expected value and already does it the Chat way — worth matching.

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
```diff
@@ -281,22 +280,22 @@ class UserRobot {
...
-        CallPage.hangUpButton.waitToAppear().click()
+        CallPage.hangUpButton.waitToAppearAndClick()
         return this
     }
 
     fun raiseHand(): UserRobot {
         settings(UserControls.ENABLE)
         if (SettingsMenu.raiseHandButton.isDisplayed()) {
-            SettingsMenu.raiseHandButton.findObject().click()
+            SettingsMenu.raiseHandButton.waitToAppearAndClick()
```
gpunto, line 297: Non-blocking follow-up — the guard line is untouched context, so this PR didn't introduce it. Raising it because the new `setView` comment two functions down documents exactly this race: "the opened menu can take several seconds to land in the accessibility tree".

`BySelector.isDisplayed()` is `findObjects().isNotEmpty()` (`Element.kt:26`) — an instant check with no wait. `raiseHandButton` is `By.text("✋ Raise hand")` in the same menu whose items `setView` just gave a 15s window, so when it hasn't landed yet the guard is false and `raiseHand()` silently no-ops instead of failing.

Same shape at three other live sites: `closedCaption` (line 266, mitigated by the bare `sleep()` above it), `transcription` (line 282, no mitigation), and `setBackground` (line 329, `By.res` rather than `By.text` but the same instant guard). `switchNoiseCancellationToggle` doesn't gate at all, so it does get the 5s wait. `lowerHand` (line 307) is also this shape but its selector is `By.text("TODO")` pending AND-562, so it's inert for an unrelated reason.

Chat's answer is `BySelector.waitDisplayed(timeOutMillis): Boolean` — a polling boolean check. It's the only public helper in Chat's `Wait.kt` that didn't come across in this port, and it's why Chat is down to 2 bare `isDisplayed()` guards against 6 here. Fine as a separate PR.

gpunto (commented):
Nice change. Most of this is a port of what Chat already runs in `stream-chat-android-e2e-test` — `Wait.kt` is near-identical down to the KDoc and the private `currentObjectOrNull`/`currentTextOrNull` helpers, and the `api_level`/`test_class` inputs and `allure-results/*` artifact path match Chat's `e2e-test.yml`. Good reason to land it, and the `runCatching` hardening in `RetryRule` is ahead of Chat's version.

So I've only flagged the parts that aren't a straight port: one blocker in the new recovery code, and one place where the port diverges from how Chat uses these helpers. Verified `:demo-app:compileE2etestingDebugAndroidTestKotlin` passes.

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/rules/RetryRule.kt
andremion: Good catch. The guard was there for the Allure write and the restore slipped in with it. Fixed in 4c0301745b71478f8ec555e84501f95f7ddd7ade: the restore now runs on every failed attempt, and only the call teardown and the attempt result stay gated.

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
andremion: Fixed in 4c0301745b71478f8ec555e84501f95f7ddd7ade. It now delegates to assertParticipantsCountOnCall, same badge and expected value, so the wait asserts and the duplication is gone.

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
andremion: Took this one here as well instead of a follow-up: 4c0301745b71478f8ec555e84501f95f7ddd7ade ports waitDisplayed and moves the five guards to it. The sleep() in closedCaption is gone too, the polling guard replaces it. testUserChangesBackground passes locally; the other guard users are still ignored (AND-562, AND-566, AND-571), so only that path has live coverage for now.

gpunto (approved):

PR #1783, opened 2026-08-25
[Awaiting Alignment] Track concurrent Call.join attempts in join analytics

rahul-lohra:
### Goal
Closes #1471

Track whether each `Call.join()` invocation starts a new join or joins an existing in-flight attempt in join analytics

### Implementation

- Hoisted the **_join single-flight processor_** to `Call` so `CallJoinCoordinator` and `JoinAnalytics` inspect the same in-flight join state.
- Added `isFlightActive()` to distinguish an active shared join from a completed or cancelled flight awaiting cleanup.
- Introduced `JoinInvocation` to represent each public `Call.join()` invocation as either:
  - `Standalone`, when no join is currently in flight.
  - `Concurrent`, when the invocation is coalesced into an active join.
- Generate a fresh stage-attempt ID for every `Call.join()` invocation.
- Preserve the active attempt ID for concurrent invocations so downstream Coordinator and SFU events remain correlated with the shared execution.
- Report concurrent invocations with `joinReason = concurrent-with:<active-attempt-id>`.
- Added tests covering standalone invocations, concurrent invocations, ID preservation, missing active IDs, and event payload serialization.

### 🎨 UI Changes

None

### Testing

Smoke test analytics

## Summary by CodeRabbit

* **New Features**
  * Improved handling of simultaneous call join attempts by coordinating them as a shared operation.
  * Enhanced join analytics to distinguish standalone joins from concurrent attempts.
  * Concurrent join events now include a reference to the active join attempt for clearer reporting.

* **Bug Fixes**
  * Prevented duplicate concurrent join operations while preserving accurate analytics for each invocation.

* **Tests**
  * Added coverage for concurrent joins, standalone joins, and cases without a prior active attempt.


rahul-lohra:
Before proceeding further, we’ll first align internally with the other platform teams on the analytics changes, as each platform currently has slightly different implementations.

PR #1785, opened 2026-08-26
Route the remaining internal render paths through VideoComponentFactory

andremion:
### Goal

Resolves [AND-1441](https://linear.app/stream/issue/AND-1441/route-the-remaining-internal-render-paths-through).

Follow-up to AND-762 (#1778). That PR covered the components behind the existing lambda slots. Some internal render paths still called components directly, so a factory override applied to the main grid but not everywhere in the same call. This PR routes those remaining defaults through `VideoTheme.componentFactory`, so an override of `ParticipantVideo`, the label or the connection indicator applies consistently on the grid, the floating self tile, picture-in-picture and screen share.

Was stacked on #1778, which is now merged, so this PR targets `develop` directly.

### Implementation

Routed existing paths, with no visual change:

- `FloatingParticipantVideo`: the default `videoRenderer` now calls `VideoComponentFactory.ParticipantVideo` instead of the composable directly.
- `DefaultPictureInPictureContent` (`CallContent.kt`): the participant tile now goes through `VideoComponentFactory.ParticipantVideo`.
- `ScreenShareVideoRenderer`: the hardcoded `ParticipantLabel` and `NetworkQualityIndicator` now go through `ParticipantVideoLabelContent` and `ParticipantVideoConnectionIndicatorContent`. This is pixel-identical because `GenericIndicator` already sizes itself to `componentHeightM`, so the extra height modifier in the factory default changes nothing.

New factory methods (5 methods plus 5 params classes, following the AND-762 conventions):

- `ParticipantsLayoutScreenSharingFallbackContent`: the screen sharing fallback (the avatar shown while the shared screen is loading). The duplicated default lambdas in `ParticipantsLayout`, `ParticipantsScreenSharing`, the portrait and landscape renderers and `ScreenShareVideoRenderer` all route through it now.
- `ParticipantLabelSoundIndicatorContent`: the sound indicator inside the participant label. The two `ParticipantLabel` overloads have different default paddings, so the params carry a `modifier` and each call site passes its own padding in it. This is the approach AND-762 deferred, with no visual change.
- `AudioOnlyCallHeaderContent`, `AudioOnlyCallDetailsContent`, `AudioOnlyCallControlsContent`: cover the `headerContent` / `detailsContent` / `controlsContent` slots of `AudioOnlyCallContent`.

One decision to note: the deprecated `AudioCallContent` is left unchanged. It delegates to `OutgoingCallContent`, and its details slot uses `List<MemberState>` while `AudioOnlyCallContent` uses `List<ParticipantState>`, so one set of factory methods cannot serve both without a visual change. I did not want to grow the new factory API around a deprecated component.

### 🎨 UI Changes

No visual changes. The snapshot suite is unchanged (no golden regenerated). One new golden is added for the new `factory screen sharing fallback` test, because the fallback is never reached in preview mode, so no existing golden pinned it.

### Testing

- `:stream-video-android-ui-compose:verifyPaparazziDebug` passes against the existing goldens.
- `apiDump` updated: 80 added lines, 0 removed (additions only, as required by the ticket). `apiCheck` passes.
- `spotlessApply` and debug compilation of the module pass.

## Summary by CodeRabbit

* **New Features**
  * Added customizable theme components for audio-only call headers, details, and controls.
  * Added configurable participant sound indicators and screen-sharing fallback content.
  * Expanded video component customization for participant video, labels, connection quality, and screen sharing.
* **Bug Fixes**
  * Improved fallback presentation when screen sharing is loading or unavailable.
* **Tests**
  * Added light- and dark-mode coverage for screen-sharing fallback content.


andremion:
@coderabbitai review

stream-video-android-ui-compose/api/stream-video-android-ui-compose.api
```diff
@@ -78,6 +78,50 @@ public final class io/getstream/video/android/compose/state/ui/participants/Invi
...
+	public final fun component2-D9Ej5fM ()F
+	public final fun component3 ()Ljava/lang/String;
+	public final fun copy-lG28NQ4 (Ljava/util/List;FLjava/lang/String;)Lio/getstream/video/android/compose/theme/AudioOnlyCallDetailsContentParams;
+	public static synthetic fun copy-lG28NQ4$default (Lio/getstream/video/android/compose/theme/AudioOnlyCallDetailsContentParams;Ljava/util/List;FLjava/lang/String;ILjava/lang/Object;)Lio/getstream/video/android/compose/theme/AudioOnlyCallDetailsContentParams;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getDuration ()Ljava/lang/String;
+	public final fun getRemoteParticipants ()Ljava/util/List;
+	public final fun getTopPadding-D9Ej5fM ()F
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
```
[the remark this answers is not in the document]
andremion: The default intentionally does not apply topPadding. Before this PR, the default branch of AudioOnlyCallContent rendered AudioOnlyCallDetails without top padding, and topPadding was only passed to a custom detailsContent lambda. Applying it in the factory default would change the rendering, and this PR must keep the snapshot suite unchanged.

This also follows the convention set in AND-762 for IncomingCallDetailsContentParams and OutgoingCallDetailsContentParams: the default implementation does not apply topPadding, it is provided for custom implementations. The AudioOnlyCallDetailsContentParams kdoc documents the same.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/FloatingParticipantVideo.kt
```diff
@@ -78,13 +81,15 @@ public fun BoxScope.FloatingParticipantVideo(
...
         RegularVideoRendererStyle(isShowingConnectionQualityIndicator = false),
     videoRenderer: @Composable (ParticipantState) -> Unit = {
-        ParticipantVideo(
-            modifier = Modifier
-                .fillMaxSize()
-                .clip(VideoTheme.shapes.dialog),
-            call = call,
-            participant = participant,
-            style = style,
+        VideoTheme.componentFactory.ParticipantVideo(
```
gpunto, line 84: This now honours the factory, but the `videoRenderer` defaults of `ParticipantsLayout` (:69), `ParticipantsRegularGrid` (:53), `ParticipantsSpotlight` (:56) and `ParticipantsScreenSharing` (:57) still call the concrete `ParticipantVideo`, which doesn't route back through the factory.

Through `CallContent` it stays consistent, since `DefaultVideoRendererSlot` goes through the factory, so this only shows for someone hosting `ParticipantsLayout` directly — they'd get an overridden PiP/floating tile next to default grid tiles on the same screen.

I wasn't sure whether AND-762 meant to leave these behind: it deferred the paths that never had a lambda slot, and `videoRenderer` is one, so they may just have slipped through. Would it make sense to route them here too, or would you rather keep this PR to the paths in the ticket?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactory.kt
```diff
@@ -351,6 +372,21 @@ public interface VideoComponentFactory {
...
     }
 
+    /**
+     * Content shown in place of the shared screen while the screen sharing session is loading or
+     * not available. The default implementation renders the sharing participant's avatar.
+     *
+     * @param params Parameters for this component.
+     */
+    @Composable
+    public fun ParticipantsLayoutScreenSharingFallbackContent(
```
gpunto: This is the shared default for `ParticipantsLayout`, `ParticipantsScreenSharing`, `Portrait`/`LandscapeScreenSharingVideoRenderer` and `ScreenShareVideoRenderer`, so the `ParticipantsLayout` prefix reads a bit narrower than what it actually covers. It lands in the `.api` dump, which is why I'd rather raise it now than after a release.

Would `ScreenSharingFallbackContent` work? It would line up with `ParticipantVideoFallbackContent`, which is named for what it falls back from rather than for a host. (Chat does the same where a component isn't owned by one host, e.g. `ScrollToBottomButton`.)

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/ScreenShareVideoRenderer.kt
```diff
@@ -52,9 +53,7 @@ public fun ScreenShareVideoRenderer(
     isShowConnectionQualityIndicator: Boolean = true,
     isZoomable: Boolean = true,
     fallbackContent: @Composable (ScreenSharingSession) -> Unit = {
-        val userName by it.participant.userNameOrId.collectAsStateWithLifecycle()
-        val userImage by it.participant.image.collectAsStateWithLifecycle()
-        UserAvatarBackground(userImage = userImage, userName = userName)
+        DefaultScreenSharingFallbackContent(session = it)
```
gpunto, line 56: Non-blocking, and pre-existing rather than something this PR introduces: `videoRendererConfig` on line 69 is a keyless `remember`, so it holds the closure over the first `session`. When the sharing participant changes, the fallback shows the previous participant's avatar until the new track loads. `remember(session, fallbackContent)` would cover it — happy for you to leave it if you'd rather keep the diff focused.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/ScreenShareVideoRenderer.kt
```diff
@@ -82,14 +81,25 @@ public fun ScreenShareVideoRenderer(
             videoRendererConfig = videoRendererConfig,
         )
 
-        ParticipantLabel(call, screenShareParticipant, labelPosition)
+        with(VideoTheme.componentFactory) {
+            ParticipantVideoLabelContent(
```
gpunto, line 85: Non-blocking: the shared-screen label now resolves to the same hook as a grid tile's label, and `ParticipantVideoLabelContentParams` doesn't carry anything that tells the two apart, so an override sized for a small grid cell would also land on top of the shared screen. The connection indicator below (:96) is in the same position — `ParticipantVideoConnectionIndicatorContentParams` carries only `networkQuality`.

Not necessarily something to solve here. `ParticipantLabelSoundIndicatorContent` in this PR handles the same shape by putting the differing part in the params as a `modifier`, so there's a pattern to reach for if these ever need to diverge. Mentioning it mainly because the params classes become public API with this PR.

stream-video-android-ui-compose/src/testDebug/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactoryTest.kt
```diff
@@ -115,6 +115,13 @@ internal class VideoComponentFactoryTest : PaparazziComposeTest {
         }
     }
 
+    @Test
+    fun `factory screen sharing fallback`() {
```
gpunto: Non-blocking: this baseline is byte-identical to `factory participant video fallback.png` (both `734b5684…`), so it pins that the hook renders an avatar but can't distinguish it from the video fallback.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/FloatingParticipantVideo.kt
andremion: Good catch, done in e4db962b13. I added an internal DefaultParticipantVideoRenderer helper in ParticipantsLayout.kt that delegates to the factory, and routed the videoRenderer defaults through it. Besides the four you listed, the internal renderers had the same default (LazyColumn/LazyRow, Portrait/Landscape, Orientation, Spotlight and the two screen sharing ones), so I routed those too. Every videoRenderer default now resolves through the factory and the snapshot suite is unchanged.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/VideoComponentFactory.kt
andremion: Agreed, and better to fix it before it ships. Renamed to ScreenSharingFallbackContent and ScreenSharingFallbackContentParams in e4db962b13, so it lines up with ParticipantVideoFallbackContent. The api dump now only contains the new names.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/ScreenShareVideoRenderer.kt
andremion: Took it in e4db962b13: remember(session, fallbackContent). It is a real staleness issue and the file is already in this PR's scope, so it felt wrong to leave it behind. No snapshot impact since previews never reach the fallback.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/internal/ScreenShareVideoRenderer.kt
andremion: Agreed on leaving it as is. Adding a context field now would be speculative API with no consumer yet. If the shared screen ever needs a different label or indicator than the grid tiles, the ParticipantLabelSoundIndicatorContent approach (the differing part travels in the params as a modifier) is the pattern to follow.

gpunto (approved):

PR #1788, opened 2026-08-27
Fix the outgoing ringing state stuck on Connecting in join-and-ring

andremion:
### Goal

Fix AND-1454. After #1782 the nightly E2E run is green on API 28, 31 and 32, but [run 33062751723](https://github.com/GetStream/stream-video-android/actions/runs/33062751723) still fails on API 33, 34 and 35 with two issues in the outgoing ringing tests.

### Implementation

**Issue 1: the outgoing call UI stays on "Connecting..." while the call is already connected.** The run artifacts show the exact sequence. In the join-and-ring flow the SFU `JoinCallResponseEvent` sets the ringing state to `Outgoing` directly, but the ring request registers the call in `client.state.ringingCall` only later. In the failing attempts a `call.session_started` coordinator event landed in that window, re-ran `updateRingingState()` with `hasRingingCall = false`, and downgraded Outgoing back to Idle (logcat: `Updating ringing state Outgoing -> Idle`). Nothing recomputed the state afterwards and no `CallRingEvent` is delivered to the caller, so the state stayed Idle and the UI rendered the full screen "Connecting..." (`LoadingContent`) until the 30s wait for `Stream_DeclineCallButton` timed out. Passing runs recovered only because some later event happened to re-run `updateRingingState()` after the ring completed.

Two changes in the core:

- `CallState.updateRingingState()` keeps the current Outgoing state instead of falling back to Idle while join-and-ring is in progress and the ringing call is not registered yet.
- `CallJoinCoordinator.joinAndRing()` recomputes the ringing state right after the ring request succeeds, so the state recovers deterministically without depending on a later coordinator event.

**Issue 2: instant asserts in `assertOutgoingCall` race the async control state.** On API 34 the outgoing screen was rendered but the microphone toggle still showed the muted state at the instant the check ran. The label, avatar, microphone and camera checks now poll with `waitDisplayed`, the same pattern AND-1445 applied to the settings menu guards.

The `testUserAcceptsTheIncomingVideoCall` failure seen on #1776 is a different bucket (leave-when-last-in-call firing during an SFU reconnect) and is tracked in AND-1455.

**Further nightly buckets fixed on this branch** (found while verifying the fix with nightly runs on the branch):

- `assertUserMicrophone` polled the control toggle but checked the participant view icon with an instant `isDisplayed()`. The hierarchy dumps taken right after the failures already contained the expected icon, so the state arrives and only the instant check misses it. Both checks poll now. This bucket failed `testCameraAndMicrophoneConfigurationInLobby` (all 3 attempts, API 32) and `testUserMicrophone` (all 3 attempts on PR #1764, plus single attempts on several API levels).
- `assertRecordingView` could leak a raw `StaleObjectException` from `waitToAppear(...).isDisplayed()` and hide the real failure. It uses the stale-safe `waitDisplayed` now.
- The caller had no outgoing call notification in the join-and-ring flow: the notification is posted by the foreground service started with `TRIGGER_OUTGOING_CALL`, which only `registerOutgoingRing()` starts, and only the create-with-ring path called it. On develop this was sometimes masked when the ringing state flapped to Idle and the ongoing service started instead; with the deterministic Outgoing state it never rendered. `joinAndRing` now calls `registerOutgoingRing()` on ring success, and the outgoing ringing E2E test asserts the notification is shown while ringing and dismissed after the decline.
- A second cause behind the reconnection test flake (the connection state staying "Reconnecting.." after a recovered reconnect) is fixed separately in #1802, split out on review request because it changes `RtcSession`. Until #1802 merges, that bucket can still occasionally exhaust a Test compose batch here; a re-run covers it.
- `testReconnectionDuringCallRecording` raced its own recording window: the buddy stops the recording 30s after its start request, the composite recorder needs 20-30s to start, and the drop plus reconnect plus asserts consumed the rest on slow emulators. The recording actually survives the reconnect fine (it is server side and the recording participant stays online). The window is now 90s. For comparison, neither the JS nor the Swift SDK has a reconnect-plus-recording scenario or any reconnect-time recording handling, so no SDK change is needed for this one.

For the core change, the Swift SDK confirms the approach: its `joinAndRingCall` sets the outgoing state explicitly and suppresses competing call state updates while the ring is in flight (`skipCallStateUpdates` in `CallViewModel`), which is the same principle as keeping Outgoing authoritative during join-and-ring here. The JS SDK avoids the bug class entirely by using explicit state transitions instead of recomputing from flags.

### 🎨 UI Changes

Not applicable. No visual changes, the fix removes a state where the outgoing ringing UI never appeared.

### Testing

- New `RingingStateJoinAndRingTest` reproduces the race on a real `CallState` with a mocked client: one test asserts that a recompute in the pre-ring window keeps Outgoing, one asserts that the recompute after `markRinging()` yields `Outgoing(acceptedByCallee = false)`. Both fail without the fix and pass with it.
- Full `:stream-video-android-core` unit test suite and `apiCheck` pass locally, and the E2E androidTest source set compiles.
- `CallStateTelecomHoldTest` and the `CallJoinCoordinatorTest` additions cover the remaining touched branches.
- Verified locally on an API 35 emulator through the real fastlane flow: the outgoing ringing test fails at the notification assert with the old code and passes with the fix.
- CI verification: dispatch the E2E workflow on this branch with `test_class: io.getstream.video.android.tests.RingingTests` on API 33, 34 and 35, where the nightly failed. The full branch nightly (run 33155106721) was green on 5 of 6 API levels before the notification and ICE fixes; the remaining bucket is AND-1455.

### How to reproduce the issues and verify the fixes

**The stuck "Connecting..." bug (state race).** It rarely fires naturally on a fast local emulator, so either use CI sampling or widen the window locally:

- CI: dispatch the "E2E Tests" workflow with `test_class: io.getstream.video.android.tests.RingingTests#testUserRejectsTheOutgoingAudioCall` and `api_level: 35`. On develop it failed 3 of 3 batch jobs (runs 33062751723 and 33142442425, where API 28 failed 3 of 3 too). On this branch the same dispatch passes, and the full branch nightly (run 33155106721) is green on all six API levels for the ringing tests.
- Manual: on develop, apply the patch below to make the race deterministic, then start a Direct Call (the "Join first" checkbox is on by default). The caller sticks on the full screen "Connecting..." forever. Same patch on this branch shows the outgoing screen.

<details>
<summary>Patch to make the race deterministic</summary>

```kotlin
// CallJoinCoordinator.joinAndRing: delay the ring request so call.session_started
// always lands in the window between the SFU join response and the ring completion.
).flatMap { rtcSession ->
    logger.d { "[joinAndRing] Joined #ringing; #track; ring: $members" }
    delay(3_000) // repro only
    apiClient.ring(RingCallRequest(isVideoEnabled(), members)).map {
```

</details>

**The missing outgoing call notification.** Deterministic, no patch needed: on develop (or this branch before commit e6c6ad9cee), start a Direct Call with "Join first" checked. No caller notification appears, and logcat prints `Outgoing call service should already be running`. With the fix the "Calling..." notification appears while ringing and disappears after the decline. The E2E test now asserts exactly that, verified locally on an API 35 emulator through the real fastlane flow: the test fails at the notification assert with the old code and passes with the fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved outgoing call behavior so the ringing interface remains visible while a join-and-ring call is being established.
  - Ensured ringing state updates correctly after the call request is registered.
  - Improved call controls’ display behavior, including correct camera visibility for audio-only calls.
  - Reduced timing-related issues when displaying outgoing call labels, avatars, and microphone controls.

- **Tests**
  - Added regression coverage for join-and-ring state handling and outgoing call controls.

### DIFF
--- a/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
+++ b/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
@@ -16,12 +16,15 @@
 
 package io.getstream.video.android.robots
 
+import android.app.Notification
+import android.app.NotificationManager
 import androidx.test.uiautomator.BySelector
 import io.getstream.video.android.pages.CallPage
 import io.getstream.video.android.pages.CallPage.SettingsMenu
 import io.getstream.video.android.pages.RingPage
 import io.getstream.video.android.robots.UserControls.DISABLE
 import io.getstream.video.android.robots.UserControls.ENABLE
+import io.getstream.video.android.uiautomator.appContext
 import io.getstream.video.android.uiautomator.defaultTimeout
 import io.getstream.video.android.uiautomator.device
 import io.getstream.video.android.uiautomator.findObject
@@ -302,6 +305,30 @@ fun UserRobot.assertOutgoingCall(audioOnly: Boolean = true, isDisplayed: Boolean
             RingPage.declineCallButton.waitToDisappear().isDisplayed(),
         )
     }
+    return this
+}
+
+/**
+ * Asserts the presence of the outgoing call notification, which the outgoing call foreground
+ * service posts with the "Calling..." title (on the ongoing calls channel, see
+ * getSimpleOngoingCallNotification). The instrumentation runs inside the app process, so the
+ * check reads NotificationManager.activeNotifications directly instead of matching text in
+ * the notification shade, where the outgoing screen shows the same "Calling..." text.
+ * The service start and stop are asynchronous, so both directions poll.
+ */
+fun UserRobot.assertOutgoingCallNotification(isDisplayed: Boolean): UserRobot {
+    val title = appContext.getString(
+        io.getstream.video.android.core.R.string.stream_video_outgoing_call_notification_title,
+    )
+    val notificationManager = appContext.getSystemService(NotificationManager::class.java)
+    fun displayed() = notificationManager.activeNotifications.any {
+        it.notification.extras.getCharSequence(Notification.EXTRA_TITLE)?.toString() == title
+    }
+    val endTime = System.currentTimeMillis() + defaultTimeout
+    while (displayed() != isDisplayed && System.currentTimeMillis() < endTime) {
+        Thread.sleep(250)
+    }
+    assertEquals("Outgoing call notification displayed", isDisplayed, displayed())
     return this
 }
 
--- a/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
+++ b/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
@@ -71,15 +71,23 @@ fun UserRobot.assertThatCallIsEnded(): UserRobot {
 }
 
 fun UserRobot.assertUserMicrophone(isEnabled: Boolean, videoCall: Boolean = true): UserRobot {
+    // The participant view icon updates slightly after the control toggle, so both
+    // checks poll instead of asserting the icon at the instant the toggle appears.
     if (isEnabled) {
-        assertTrue(CallPage.microphoneEnabledToggle.waitToAppear().isDisplayed())
+        assertTrue("Microphone enabled toggle", CallPage.microphoneEnabledToggle.waitDisplayed())
         if (videoCall) {
-            assertTrue(CallPage.ParticipantView.microphoneEnabledIcon.isDisplayed())
+            assertTrue(
+                "Participant microphone enabled icon",
+                CallPage.ParticipantView.microphoneEnabledIcon.waitDisplayed(),
+            )
         }
     } else {
-        assertTrue(CallPage.microphoneDisabledToggle.waitToAppear().isDisplayed())
+        assertTrue("Microphone disabled toggle", CallPage.microphoneDisabledToggle.waitDisplayed())
         if (videoCall) {
-            assertTrue(CallPage.ParticipantView.microphoneDisabledIcon.isDisplayed())
+            assertTrue(
+                "Participant microphone disabled icon",
+                CallPage.ParticipantView.microphoneDisabledIcon.waitDisplayed(),
+            )
         }
     }
     return this
--- a/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
+++ b/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
@@ -29,6 +29,7 @@ import io.getstream.video.android.uiautomator.findObjects
 import io.getstream.video.android.uiautomator.isDisplayed
 import io.getstream.video.android.uiautomator.retryOnStaleObjectException
 import io.getstream.video.android.uiautomator.seconds
+import io.getstream.video.android.uiautomator.waitDisplayed
 import io.getstream.video.android.uiautomator.waitForCount
 import io.getstream.video.android.uiautomator.waitForText
 import io.getstream.video.android.uiautomator.waitToAppear
@@ -271,14 +272,17 @@ fun UserRobot.assertOutgoingCall(audioOnly: Boolean = true, isDisplayed: Boolean
             "Decline call button",
             RingPage.declineCallButton.waitToAppear(timeOutMillis = 30.seconds).isDisplayed(),
         )
-        assertTrue("Call label", RingPage.outgoingCallLabel.isDisplayed())
-        assertTrue("Avatar", RingPage.callParticipantAvatar.isDisplayed())
-        assertTrue("Microphone", RingPage.microphoneEnabledToggle.isDisplayed())
-        assertEquals(
-            "Camera should be displayed: ${!audioOnly}",
-            !audioOnly,
-            RingPage.cameraEnabledToggle.isDisplayed(),
-        )
+        // The control toggles reflect async call state (the microphone can still show the
+        // muted state right after the screen renders), so poll instead of instant asserts.
+        assertTrue("Call label", RingPage.outgoingCallLabel.waitDisplayed())
+        assertTrue("Avatar", RingPage.callParticipantAvatar.waitDisplayed())
+        assertTrue("Microphone", RingPage.microphoneEnabledToggle.waitDisplayed())
+        if (audioOnly) {
+            assertFalse("Camera enabled toggle", RingPage.cameraEnabledToggle.isDisplayed())
+            assertFalse("Camera disabled toggle", RingPage.cameraDisabledToggle.isDisplayed())
+        } else {
+            assertTrue("Camera", RingPage.cameraEnabledToggle.waitDisplayed())
+        }
     } else {
         assertFalse(
             "Decline call button",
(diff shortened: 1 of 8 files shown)


andremion:
@coderabbitai review

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/RingingStateJoinAndRingTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+    private val call = mockk<Call>(relaxed = true) {
+        every { type } returns "default"
+        every { id } returns "join-and-ring-test"
+        every { cid } returns "default:join-and-ring-test"
+    }
+
+    @After
+    fun tearDown() {
+        scope.cancel()
```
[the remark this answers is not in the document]
andremion: Fixed in 4f8ef25627. The test now extends TestBase, runs with Robolectric, and uses the shared test dispatcher from DispatcherRule. This also fixed the CI unit test failure: CallState collects call.events, and on the relaxed mock that threw KotlinNothingValueException into the global uncaught handler, which failed the next runTest (SpeakerManagerTest). The events flow is now a real MutableSharedFlow and the scope has an exception handler that fails this test instead.

demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobotCallAsserts.kt
```diff
@@ -271,14 +272,16 @@ fun UserRobot.assertOutgoingCall(audioOnly: Boolean = true, isDisplayed: Boolean
...
-            !audioOnly,
-            RingPage.cameraEnabledToggle.isDisplayed(),
-        )
+        // The control toggles reflect async call state (the microphone can still show the
+        // muted state right after the screen renders), so poll instead of instant asserts.
+        assertTrue("Call label", RingPage.outgoingCallLabel.waitDisplayed())
+        assertTrue("Avatar", RingPage.callParticipantAvatar.waitDisplayed())
+        assertTrue("Microphone", RingPage.microphoneEnabledToggle.waitDisplayed())
+        if (audioOnly) {
+            assertFalse("Camera", RingPage.cameraEnabledToggle.isDisplayed())
```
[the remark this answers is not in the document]
andremion: Fixed in 4f8ef25627. The audio-only branch now asserts that both the enabled and the disabled camera toggles are absent.

andremion:
@coderabbitai review

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -191,6 +191,10 @@ internal class CallJoinCoordinator(
             apiClient.ring(RingCallRequest(isVideoEnabled(), members)).map {
                 logger.d { "[joinAndRing] Ringed #ringing; #track; ring: $members" }
                 callRegistry.markRinging()
+                // An event that arrived before the ring completed (e.g. call.session_started)
+                // computed the ringing state without the ringing call registered. Recompute so
+                // the state cannot stay Idle when no further coordinator event arrives.
+                state.updateRingingState()
```
gpunto: Non-blocking: this yields `Outgoing(acceptedByCallee = false)`, but the `JoinCallResponseEvent` handler (`CallState:1131`) unconditionally sets `acceptedByCallee = true` while join-and-ring is in progress. A reconnect inside the outgoing window re-emits that event and flips it back, stopping the ringback with nobody having accepted. Pre-existing, but this is the first code to disagree with it — worth routing :1131 through `updateRingingState()`.

rahul-lohra, line 338: Could you elaborate on why you invoked `state.updateRingingState()` here? Specifically, which properties are you trying to ensure are updated correctly?

andremion: It is not about one property. markRinging() only sets client.state._ringingCall and nothing recomputes the derived ringing state after that. The SFU join response sets Outgoing directly, but until the ring request completes hasRingingCall is false, so any updateRingingState() call in that window used to compute Idle and hide the outgoing UI (in the failing nightly runs a call.session_started event did exactly that). The coordinator does not send CallRingEvent to the caller, so if no other event arrives after the ring succeeds, nothing ever recomputes and the state stays Idle forever. This call re-evaluates the state at the first moment hasRingingCall is guaranteed true, so it lands on Outgoing(acceptedByCallee = false) deterministically instead of depending on a later unrelated event. The second test in RingingStateJoinAndRingTest fails without this line.

rahul-lohra:
Why don't we update `client.state.ringingCall` right away after updating `call.state.ringing`

Like this - 
```kotlin
            is JoinCallResponseEvent -> {
                // time to update call state based on the join response
                updateFromJoinResponse(event)
                if (!isJoinAndRingInProgress.get()) {
                    updateRingingState()
                } else {
                    _ringingState.value = RingingState.Outgoing(acceptedByCallee = true)
                    client.state._ringingCall = this //NEW CODE
                }
                ...
```

andremion:
> Why don't we update client.state.ringingCall right away after updating call.state.ringing

That closes the same window, but it makes `client.state.ringingCall` claim a ring that does not exist yet: at `JoinCallResponseEvent` time the ring request has not even been sent. Today `_ringingCall` is only written on confirmation, and the busy handler, `hasActiveOrRingingCall()` and the service/sound handling all read it. It would also need cleanup if the ring request fails. So I kept registration on confirmation and made the derived state handle the gap instead.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
rahul-lohra: okay

rahul-lohra:
While testing, I was unable to see any ringing notification on the caller side. This needs a revisit. 
cc @andremion

andremion:
> While testing, I was unable to see any ringing notification on the caller side. This needs a revisit.

The notification comes from the foreground service started with TRIGGER_OUTGOING_CALL, and joinAndRing never started it (only the create-with-ring path did, via registerOutgoingRing). It was already broken on develop but sometimes masked by the state flap this PR removes, so you saw it consistently. Fixed in 8075ccec13: joinAndRing now calls registerOutgoingRing() on ring success. The outgoing ringing E2E test now asserts the notification appears while ringing and is gone after the decline, verified locally on an API 35 emulator: it fails with the old code and passes with the fix. See the new "How to reproduce" section in the PR description if you want to verify manually.

andremion:
@rahul-lohra One more nightly bucket is fixed here: after a network blip the call could stay on "Reconnecting.." forever. Recovery to Connected happens only in the ICE health monitor, and it required both peer connections to reach an established ICE state, but a peer connection with nothing to negotiate stays NEW forever (the subscriber right after a reconnect), so recovery never fired. This is why testReconnectionDuringCallRecording kept exhausting retries on CI. The decision is now extracted into iceHealthTransition: recover when the SFU socket is connected and no ICE side is DISCONNECTED or FAILED. Unit tests pin the exact state combination from the CI logs and the recovery test fails against the old predicate; the E2E reconnection test passes locally with a real network drop.

rahul-lohra:
Adding more context here for better clarity on the fixes
<html>
<head>

</head>
<body>

  | Before | After
-- | -- | --
Ringing state transition | Idle → Outgoing → Idle → Outgoing | Idle → Outgoing
Service trigger | TRIGGER_ONGOING_CALL ❌ | TRIGGER_OUTGOING_CALL ✅
Outgoing notification text | Call in progress ❌ | Calling… ✅
Behavior | Service could be launched with the wrong trigger due to the unnecessary state transition | Service is launched directly with the correct trigger, and the notification correctly reflects the outgoing call state


</body>
</html>

rahul-lohra:
@PratimMallick please review the the code in `RtcSession.kt` and see its test

rahul-lohra (approved):
LGTM

PR #1575, opened 2025-11-25
Configure coverage through common convention plugins

gpunto:
### Goal

[AND-892](https://linear.app/stream/issue/AND-892/extract-coverage-configuration-to-the-convention-plugins)

As the title says, we're moving to the common coverage convention instead of configuring it in here.

### Implementation

Remove the custom coverage configuration logic and rely on the plugins

### 🎨 UI Changes

None

### Testing

Coverage should be reported correctly to Sonar

#### General
- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [ ] Assigned a person / code owner group (required)
- [ ] Thread with the PR link started in a respective Slack channel (required internally)
- [ ] PR targets the `develop` branch
- [ ] PR is linked to the GitHub issue it resolves

#### Code & documentation
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (KDocs, docusaurus, tutorial)
- [ ] Tutorial starter kit updated
- [ ] Examples/guides starter kits updated (`stream-video-examples`)

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Compose sample runs & works
- [ ] Tutorial starter kit
- [ ] Example starter kits work
- [ ] UI Changes correct (before & after images)
- [ ] Bugs validated (bugfixes)
- [ ] New feature tested and works
- [ ] Release notes and docs clearly describe changes
- [ ] All code we touched has new or updated KDocs
- [ ] Check the SDK Size Comparison table in the CI logs

### 🎉 GIF

_Please provide a suitable gif that describes your work on this pull request_

PR #1789, opened 2026-08-28
Suppress the last participant auto leave while the call is reconnecting

andremion:
### Goal

Fixes AND-1455.

This PR targets `develop-v2`. The fix changes when the last-participant auto leave fires (only while the connection is `Connected`). The video team considers that a behavior change that some customers might see as breaking, so it ships with the next major version instead of a patch release on `develop`.

`RingingTests#testUserAcceptsTheIncomingVideoCallWithCameraAndMicrophoneEnabled` fails on CI when the SFU websocket drops right after the callee accepts (for example run 33070269061 on PR #1776, and the API 34 job of run 33155106721). The chain is:

1. The SFU socket drops during the join and `CallReconnector` starts a REJOIN.
2. While the reconnect is running, the participant roster is unreliable: the rejoin removes our previous participant record and the remote participant of the failing SFU is gone. The roster falls to 1 or less.
3. `StreamCallActivity` observes the roster with `leaveWhenLastInCall = true`, sees "last participant", and fires `LeaveCall`.
4. The leave cancels `call.scope`. The reconnect loop runs in that scope, so it dies in the middle of its retry.
5. The screen stays on "Connecting..." until the test times out.

### Implementation

- Extracted the last participant detection from `StreamCallActivity.processParticipantLeftEvent` into an internal `lastParticipantSignal` flow helper (`LastParticipantSignal.kt`).
- The helper combines the participant roster with `call.state.connection`, keeps the existing debounce, and only emits while the connection is `Connected`. It is an allowlist, not a list of bad states: the roster is also unreliable during the initial join (`Joined` is set before `connectInternal` runs), and new `RealtimeConnection` subtypes stay covered. This mirrors the Swift SDK's `LastParticipantAutoLeavePolicy`, which only acts when the reconnection status is connected.
- The signal is a rising edge of the roster, not of the combined condition. Becoming the last participant arms it, and it fires on the first connected evaluation after that. A reconnect flap with an unchanged roster does not call `onLastParticipant` again. A remote participant joining and leaving again re-arms it and signals a second time.
- Because the connection state is part of the combined stream, the roster is re-evaluated when the reconnect settles:
  - The reconnect succeeds and the roster is restored: no leave. This is the CI failure case.
  - The reconnect succeeds but the user is really alone: the leave still fires.
  - The reconnect fails terminally: the helper stays silent. `CallReconnector` already leaves the call itself when retries are exhausted, so there is no stuck screen and no duplicate leave.
- The diagnostic logging of every roster evaluation is kept and now also logs the connection state.
- `stream-video-android-ui-core` is added to `coverage.includedModules` in the root build file so its unit tests run in CI. This also un-skips the module in Sonar for the first time, so the Sonar check reports the module's legacy findings. Proper Sonar onboarding of the module is tracked in AND-1470.

### 🎨 UI Changes

No UI changes.

### Testing

- New `LastParticipantSignalTest` (12 tests, Turbine with virtual time) in a new unit test source set for `stream-video-android-ui-core`. The module had no unit tests before, so the test dependencies were added to its build file.
- Mutation checks, so the tests catch the exact regressions:
  - Removing the connection gate (`connected = true`) fails 7 of the 12 tests.
  - Replacing the rising edge with `armed = isLast` fails the reconnect flap test.
- `spotlessCheck`, `testDebugUnitTest`, `apiCheck` and the debug and release compilations are green on `develop-v2`. The helper is internal, so the public API dump is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when the last participant leaves a call.
  * Prevented premature end-of-call behavior while reconnecting or migrating.
  * Ensured the call state is re-evaluated after reconnection settles.
  * Reduced incorrect triggers caused by brief participant roster changes.

* **Tests**
  * Added coverage for connection transitions, participant changes, and debounced updates.

### DIFF
--- a/stream-video-android-ui-core/build.gradle.kts
+++ b/stream-video-android-ui-core/build.gradle.kts
@@ -52,4 +52,10 @@ dependencies {
     implementation(libs.androidx.lifecycle.runtime)
 
     implementation(libs.stream.log)
+
+    testImplementation(libs.junit)
+    testImplementation(libs.mockk)
+    testImplementation(libs.turbine)
+    testImplementation(libs.kotlinx.coroutines.test)
+    testImplementation(libs.kotlin.test.junit)
 }
--- a/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
+++ b/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
@@ -78,9 +78,9 @@ import io.getstream.video.android.ui.common.StreamCallActivity.Companion.callInt
 import io.getstream.video.android.ui.common.models.StreamCallActivityException
 import io.getstream.video.android.ui.common.permission.PermissionManager
 import io.getstream.video.android.ui.common.util.StreamCallActivityDelicateApi
+import io.getstream.video.android.ui.common.util.lastParticipantSignal
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
@@ -88,7 +88,6 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.collectLatest
-import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
@@ -1478,15 +1477,17 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
     /**
      * Processes participant leave events for the given [call].
      *
-     * Observes [cachedCall.state.participants] and triggers [onLastParticipant] if only
-     * one or fewer participants remain. Debouncing is applied to handle quick network
-     * disconnect/reconnect scenarios.
+     * Observes [cachedCall.state.participants] together with the connection state and triggers
+     * [onLastParticipant] if only one or fewer participants remain while the connection is
+     * settled. Debouncing is applied to handle quick network disconnect/reconnect scenarios,
+     * and the check is suppressed while the connection is reconnecting or migrating, because
+     * the roster is unreliable then and leaving would cancel the reconnect. See
+     * [lastParticipantSignal].
      *
      * @param call the active [Call] associated with the event.
      * @param event the [VideoEvent] that triggered this processing, typically a [ParticipantLeftEvent]
      *              or [CallSessionParticipantLeftEvent].
      */
-    @OptIn(FlowPreview::class)
     private fun processParticipantLeftEvent(call: Call, event: VideoEvent) {
         /**
          * - participantCountJob will be null when activity is newly created
@@ -1494,25 +1495,23 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
          */
         if (participantCountJob == null) {
             participantCountJob = lifecycleScope.launch(supervisorJob) {
-                cachedCall.state.participants
-                    /**
-                     * A debounce is applied here to handle quick disconnect/reconnect scenarios
-                     * caused by unstable network conditions. Without the debounce, other devices
-                     * may receive a [ParticipantLeftEvent] prematurely, which could trigger
-                     * unintended reactions in the call flow.
-                     */
-                    .debounce(getParticipantUpdateDebounce(call))
-                    .collect {
-                        logger.d { "Participant left, remaining: ${it.size}" }
+                lastParticipantSignal(
+                    participants = cachedCall.state.participants,
+                    connection = cachedCall.state.connection,
+                    debounceMs = getParticipantUpdateDebounce(call),
+                    onEvaluated = { roster, connection ->
+                        logger.d {
+                            "Participant left, remaining: ${roster.size}, connection: $connection"
+                        }
                         lifecycleScope.launch(Dispatchers.Default) {
-                            it.forEachIndexed { i, v ->
+                            roster.forEachIndexed { i, v ->
                                 logger.d { "Participant [$i]=${v.name.value}" }
                             }
                         }
-                        if (it.size <= 1) {
-                            onLastParticipant(call)
-                        }
-                    }
+                    },
+                ).collect {
+                    onLastParticipant(call)
+                }
             }
         }
     }
--- a/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
+++ b/stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
@@ -1479,10 +1479,10 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
      *
      * Observes [cachedCall.state.participants] together with the connection state and triggers
      * [onLastParticipant] if only one or fewer participants remain while the connection is
-     * settled. Debouncing is applied to handle quick network disconnect/reconnect scenarios,
-     * and the check is suppressed while the connection is reconnecting or migrating, because
-     * the roster is unreliable then and leaving would cancel the reconnect. See
-     * [lastParticipantSignal].
+     * [RealtimeConnection.Connected]. Debouncing is applied to handle quick network
+     * disconnect/reconnect scenarios, and the check requires a connected state because the
+     * roster is unreliable while a join or reconnect is running and leaving would cancel it.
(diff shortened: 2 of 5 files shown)


andremion:
@coderabbitai review

andremion:
@coderabbitai TestBase lives in stream-video-android-core's own test source set and is not published as test fixtures, so tests in stream-video-android-ui-core cannot reference it. This module had no unit tests before this PR. Keeping plain runTest here, the test only needs virtual time for the debounce.

stream-video-android-ui-core/build.gradle.kts
```diff
@@ -52,4 +52,10 @@ dependencies {
     implementation(libs.androidx.lifecycle.runtime)
 
     implementation(libs.stream.log)
+
+    testImplementation(libs.junit)
```
gpunto, line 56: These tests never run in CI. The unit-test job runs `./gradlew :testCoverage`, which only depends on the modules in `coverage.includedModules` — currently `stream-video-android-core` and `stream-video-android-ui-compose` (root `build.gradle.kts:50`). `:stream-video-android-ui-core:testCoverage` isn't even registered (`task 'testCoverage' not found in project`), so nothing will catch a regression here.

Adding `"stream-video-android-ui-core"` to `includedModules` fixes it.

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/util/LastParticipantSignal.kt
```diff
@@ -0,0 +1,63 @@
...
+): Flow<List<ParticipantState>> =
+    combine(participants, connection) { roster, connectionState -> roster to connectionState }
+        .debounce(debounceMs)
+        .onEach { (roster, connectionState) -> onEvaluated(roster, connectionState) }
+        .filter { (roster, connectionState) ->
+            roster.size <= 1 && !connectionState.isReconnectInProgress()
+        }
+        .map { (roster, _) -> roster }
+
+private fun RealtimeConnection.isReconnectInProgress(): Boolean =
```
gpunto: The blocklist leaves the initial join unguarded. `CallJoinCoordinator` sets `Joined` before `connectInternal()`, and `RtcSession` sets `InProgress` during the handshake — so for the whole SFU join the roster is 0 or 1 and this returns `false`. The activity subscribes to call events in `initializeCallOrFail` before the join runs, so a coordinator participant-left event arriving mid-join starts the job and fires `LeaveCall` a second later.

An allowlist (`connectionState is RealtimeConnection.Connected`) closes that in the same expression and is robust to new `RealtimeConnection` subtypes. Nothing is lost on `ReconnectingFailed`: `CallReconnector` already calls `lifecycle.leave(RetryExhausted)` itself when retries are exhausted.

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
```diff
@@ -1478,41 +1477,41 @@ public abstract class StreamCallActivity : ComponentActivity(), ActivityCallOper
...
-                     * caused by unstable network conditions. Without the debounce, other devices
-                     * may receive a [ParticipantLeftEvent] prematurely, which could trigger
-                     * unintended reactions in the call flow.
-                     */
-                    .debounce(getParticipantUpdateDebounce(call))
-                    .collect {
-                        logger.d { "Participant left, remaining: ${it.size}" }
+                lastParticipantSignal(
+                    participants = cachedCall.state.participants,
+                    connection = cachedCall.state.connection,
```
gpunto, line 1500: Non-blocking: with `connection` in the combined stream and no `distinctUntilChanged`, `onLastParticipant` now re-fires on connection transitions alone. A `Connected → Reconnecting → Connected` flap with a roster stably at 1 re-invokes the hook once per settled transition, where previously only roster changes could trigger it. `atomicLeave` absorbs the duplicate leave, but `onLastParticipant` is public and open, so integrators overriding it will see the repeats. A `.distinctUntilChanged()` after the filter, or a rising-edge check, restores the old one-shot semantics.

rahul-lohra:
PR LGTM, I only need to reproduce this issue and verify this flow

rahul-lohra:
Gr8 work

stream-video-android-ui-core/build.gradle.kts
andremion: Good catch, thanks. Added the module to includedModules in c87f69c39ec4842aaab06614111eafcb996cf0a7 and verified locally that :testCoverage now runs :stream-video-android-ui-core:testDebugUnitTest.

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/util/LastParticipantSignal.kt
andremion: Agreed. I verified both premises: Joined is set before connectInternal() in CallJoinCoordinator, and CallReconnector calls leave(RetryExhausted) itself, so the allowlist loses nothing on ReconnectingFailed. Changed the filter to connectionState is RealtimeConnection.Connected in c87f69c39ec4842aaab06614111eafcb996cf0a7, with new tests for PreJoin/InProgress and ReconnectingFailed.

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
andremion: True, the combine made the hook re-fire on connection transitions. Added distinctUntilChanged after the filter in c87f69c39ec4842aaab06614111eafcb996cf0a7, plus a test covering a Connected -> Reconnecting -> Connected flap with a stable roster.

gpunto: Thanks — this does stop the transition repeats. One thing I missed when I suggested `distinctUntilChanged`, though: it sits *after* the filter, so it only ever sees last-participant rosters. The states that should reset it are dropped upstream and never reach it.

Sequence:

1. Alone → `[local]` passes the filter, emitted. `distinctUntilChanged` retains `[local]`.
2. A remote joins → `[local, remote]`, size 2, dropped by the filter. The retained value is still `[local]`.
3. The remote leaves again → `[local]` arrives, equals the retained value, dropped.

So the second genuine last-participant event is lost. `ParticipantState` is a data class and the local instance survives the round trip, so the lists compare equal. Still non-blocking: with `leaveWhenLastInCall = true` the first signal leaves and finishes the activity, so there's no second event to miss — it only bites the `false` default combined with an overridden `onLastParticipant`.

I pushed the alternative as 69aca0d on `gianmarcodavid/and-1455-last-participant-rising-edge`, based on your `c87f69c` so it cherry-picks cleanly. It tracks "already signalled" in an `armed` flag instead of inferring it from roster equality: a roster rising edge arms the signal, the first connected evaluation consumes it. The connection can release a pending signal but never create one, so an unchanged roster across connection transitions still doesn't repeat. `distinctUntilChanged` comes off the end — it would drop the second `[local]` again.

Your 11 tests pass unchanged, including `does not emit the same roster again after a reconnect flap`, and the commit adds one for the rejoin case. Mutation-checked both ways: the new test fails on `distinctUntilChanged`, and collapsing the fold to `armed = isLast` fails your flap test.

Cost is three lines of predicate instead of one, plus a small private state class. Happy for it to be a follow-up if you'd rather land the fix as it is.

gpunto (approved):

stream-video-android-ui-core/build.gradle.kts
andremion: We're keeping the module in includedModules and doing the proper Sonar onboarding in a follow-up ticket: AND-1470 (https://linear.app/stream/issue/AND-1470). Context: this is the module's first Sonar analysis ever, so Sonar counts the whole module as new code and the quality gate on this PR fails on legacy findings (39 issues, one S6384 BLOCKER, new coverage 59.1% vs 80%). Some history: the old scripts/coverage.gradle always listed ui-core in coverageModules but only activated modules that had a src/test folder, so the module was silently skipped until this PR added the first tests, and #1575 carried that effective set over into the static includedModules.

gpunto: Sounds good to me!

stream-video-android-ui-core/src/main/kotlin/io/getstream/video/android/ui/common/StreamCallActivity.kt
andremion: You're right, and nice catch on the reset states never reaching distinctUntilChanged. I verified the rising-edge fold against all the test scenarios and ran your mutation check locally (armed = isLast fails exactly the flap test). Cherry-picked as 67212b3dc4b57a64261b802c246f8bf95966049c, with you as the author. Thanks for pushing the branch, that made it easy.

andremion:
Converting to draft until we face this issue again

PR #1791, opened 2026-08-31
Harden the racy signal assertions in NoiseCancellationSignalTest

andremion:
### Goal

Fix the test that fails deterministically on the develop-v2 merge PR #1790: `NoiseCancellationSignalTest > a replacement session is told the current state` fails there with `expected:<[true]> but was:<[]>` on every attempt of the [Android CI run](https://github.com/GetStream/stream-video-android/actions/runs/33366592143), while it passes on develop.

Resolves AND-1465.

### Implementation

The failure is a race in the test, not in the production code. The `record` helper collects the signalled states into a plain `mutableListOf` from the mock's `coAnswers` block, which runs on the signal coroutine. The tests assert on that list right after a `coVerify(timeout = ...)`. MockK records the invocation before the answer block finishes, and the list is not thread safe, so the assert can run before the append is visible to the test thread. develop's coroutines 1.9.0 happens to give the timing the test needs; develop-v2's coroutines 1.10.2 consistently does not.

Changes, all in `NoiseCancellationSignalTest.kt`:

- `record()` now collects into a `CopyOnWriteArrayList`, so writes from the signal coroutine are visible to the test thread.
- The four instant asserts on the recorded list are replaced with polling helpers that wait for the expected content with the existing 5s deadline: `awaitSignalled` for the exact-content checks and `awaitFinalSignal` for the "rapid changes" test where only the last state matters.

No production code change.

### 🎨 UI Changes

Not applicable, test-only change.

### Testing

- Ran the test class 4 times locally (once normally, 3 times with `--rerun-tasks`): 8 tests, 0 failures each time.
- The definitive proof will be the next develop merge into develop-v2, where the previous version failed on both CI attempts.

## Summary by CodeRabbit

* **Tests**
  * Improved signal-related test reliability by accommodating asynchronous event recording.
  * Replaced immediate snapshot checks with polling assertions and timeout handling.
  * Added thread-safe recording for mocked signal events.

### DIFF
--- a/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
+++ b/stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/NoiseCancellationSignalTest.kt
@@ -32,6 +32,7 @@ import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 import stream.video.sfu.signal.StartNoiseCancellationResponse
 import stream.video.sfu.signal.StopNoiseCancellationResponse
+import java.util.concurrent.CopyOnWriteArrayList
 import kotlin.test.assertEquals
 
 @RunWith(RobolectricTestRunner::class)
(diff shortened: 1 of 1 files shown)


PratimMallick (approved):

andremion:
@coderabbit review

PR #1794, opened 2026-08-31
Null-check the stop intent before stopService in cleanup

andremion:
### Goal

Port the stop-intent null check from #1792 to develop. `ServiceIntentBuilder.buildStopIntent(...)` returns null when the call service is not running, and `cleanup()` passed that null into `context.stopService(...)`, throwing a `NullPointerException` on every logout without a running call service. The exception is swallowed by `safeCall`, so it only shows up as noise in the logs, but the `stopService` call itself never worked in that state.

The develop-v2 side of this defect is fixed in #1792, together with the logout main-thread freeze (that half is develop-v2-only code). Since nothing merges develop-v2 back into develop, this half needs its own PR here, as raised in the #1792 review.

Part of AND-1466.

### Implementation

`serviceIntent.let { context.stopService(serviceIntent) }` becomes `buildStopIntent(...)?.let { serviceIntent -> context.stopService(serviceIntent) }`. The block is byte-identical to the one in #1792, so the next develop into develop-v2 merge resolves it without a conflict.

### 🎨 UI Changes

Not applicable.

### Testing

- `StreamVideoClientTest` (which includes the existing `cleanup` test) and spotless pass on this branch.
- The regression tests exercising the cleanup paths live in #1792 (`StreamVideoClientCleanupTest`) and reach develop-v2 with that PR; they are not portable here because they cover the develop-v2-only disconnect bridge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved call service cleanup by avoiding unnecessary stop requests when the service is not running.
  * Prevented potential cleanup issues when no service stop intent is available.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
@@ -260,11 +260,11 @@ internal class StreamVideoClient internal constructor(
         val runCallServiceInForeground = callConfig.runCallServiceInForeground
         if (runCallServiceInForeground) {
             safeCall {
-                val serviceIntent = ServiceIntentBuilder().buildStopIntent(
+                // buildStopIntent returns null when the service is not running.
+                ServiceIntentBuilder().buildStopIntent(
                     context = context,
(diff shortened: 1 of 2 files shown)


andremion:
@coderabbitai review

rahul-lohra (approved):
lgtm

PR #1796, opened 2026-09-01
Isolate per-track mute RPCs and atomic mute map

PratimMallick:
### Goal

Two **pre-existing** races in `setMuteState`, both easy to hit when fast reconnect restarts `listenToMediaChanges` and audio/video collectors fire together:

1. One shared `muteStateSyncJob` cancelled every in-flight `UpdateMuteStates` when another track published. Logs: `HTTP FAILED: java.io.IOException: Canceled` while outbound audio RTP kept climbing.
2. The local mute map was a non-atomic read–copy–write, so concurrent collectors could drop another track's bit and `retryWhen` could stop for the wrong reason.

Fixes [AND-1474](https://linear.app/stream/issue/AND-1474/mute-sync-cancels-in-flight-updatemutestates-for-other-tracks). Independent of AND-1455.

### Implementation

- Per-`TrackType` jobs (`TrackKeyedJobs`): video/screen-share sync no longer cancels an in-flight audio unmute. Same-track updates still cancel the previous job so retries coalesce.
- `muteState.update { it + (trackType to isEnabled) }` so concurrent map writes cannot lose another track's entry.
- Mute-state retries compare only that track's intended value.
- Session teardown still cancels all mute-sync jobs.

### Testing

- [x] `./gradlew :stream-video-android-core:testDebugUnitTest --tests 'io.getstream.video.android.core.call.TrackKeyedJobsTest'` — passed
- [x] `./gradlew :stream-video-android-core:spotlessApply`
- [ ] Manual: two-person video call, debug menu **SFU fast reconnect**. Confirm `UpdateMuteStates` for audio and video both complete (no `IOException: Canceled` from a sibling track). Mic/camera stay in the state they were in before reconnect.

#### General

#### Code & documentation

### ☑️Reviewer Checklist
- [ ] XML sample runs & works
- [ ] Tutorial starter kit

### 🎉 GIF

N/A — signaling-only fix, no UI change.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
@@ -326,7 +326,7 @@ public class RtcSession internal constructor(
         )
     },
 ) {
-    private var muteStateSyncJob: Job? = null
+    private val muteStateSyncJobs = TrackKeyedJobs()
     private val oneBasedSessionCounter = sessionCounter + 1
 
     /**
@@ -1341,6 +1341,10 @@ public class RtcSession internal constructor(
      * matches the web SDK, which only signals the track(s) that actually changed. On SFU
      * (re)connect/migration each enabled track is re-signalled individually via
      * [listenToMediaChanges], so the full state is still restored.
+     *
+     * Sync jobs are keyed by [trackType]. Cancelling a shared job used to drop an in-flight
+     * audio unmute when video (or screen-share) published a moment later — reconnect restarts
+     * [listenToMediaChanges] and fires those collectors together.
      */
     private fun setMuteState(isEnabled: Boolean, trackType: TrackType) {
         logger.d { "[setPublishState] #sfu; $trackType isEnabled: $isEnabled" }
@@ -1352,12 +1356,8 @@ public class RtcSession internal constructor(
         muteState.value = new
 
         val currentSfu = sfuUrl
-        // prevent running multiple of these at the same time
-        // if there's already a job active. cancel it
-        muteStateSyncJob?.cancel()
-        // start a new job
-        // this code is a bit more complicated due to the retry behaviour
-        muteStateSyncJob = coroutineScope.launch {
+        // Coalesce retries for this track only. Other tracks keep their in-flight RPCs.
+        muteStateSyncJobs.launch(coroutineScope, trackType) {
             flow {
                 val request = UpdateMuteStatesRequest(
                     session_id = sessionId,
@@ -1372,7 +1372,7 @@ public class RtcSession internal constructor(
                 emit(response)
             }.flowOn(DispatcherProvider.IO).retryWhen { cause, attempt ->
                 if (cause is SessionFatalException) return@retryWhen false
-                val sameValue = new == muteState.value
+                val sameValue = muteState.value[trackType] == isEnabled
                 val sameSfu = currentSfu == sfuUrl
                 val isPermanent = isPermanentError(cause)
                 val willRetry = !isPermanent && sameValue && sameSfu && attempt < 30
@@ -2121,8 +2121,7 @@ public class RtcSession internal constructor(
         if (cancelEventJob) eventJob?.cancel()
         iceMonitoringJob?.cancel()
         iceMonitoringJob = null
-        muteStateSyncJob?.cancel()
-        muteStateSyncJob = null
+        muteStateSyncJobs.cancelAll()
         participantsMonitoringJob?.cancel()
         participantsMonitoringJob = null
         serialProcessor.stop()
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
@@ -121,6 +121,7 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.retryWhen
+import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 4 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1352,12 +1356,8 @@ public class RtcSession internal constructor(
...
 
         val currentSfu = sfuUrl
-        // prevent running multiple of these at the same time
-        // if there's already a job active. cancel it
-        muteStateSyncJob?.cancel()
-        // start a new job
-        // this code is a bit more complicated due to the retry behaviour
-        muteStateSyncJob = coroutineScope.launch {
+        // Coalesce retries for this track only. Other tracks keep their in-flight RPCs.
+        muteStateSyncJobs.launch(coroutineScope, trackType) {
```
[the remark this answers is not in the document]
PratimMallick: Addressed — the local map write is now atomic:

```kotlin
muteState.update { it + (trackType to isEnabled) }
```

Concurrent audio/video collectors can no longer lose another track's bit via a stale read–copy–write. `retryWhen` still reads `muteState.value[trackType]`, so it now sees the latest intended state for that track.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -2121,8 +2121,7 @@ public class RtcSession internal constructor(
         if (cancelEventJob) eventJob?.cancel()
         iceMonitoringJob?.cancel()
         iceMonitoringJob = null
-        muteStateSyncJob?.cancel()
-        muteStateSyncJob = null
+        muteStateSyncJobs.cancelAll()
```
rahul-lohra, line 2160: ### Minor Lifecycle hygiene (non-blocker)

Do you we should clear the muteStateSyncJobs in `RtcSession.cleanup()`

Like this - 

```kotin
mediaScope.cancel()
muteStateSyncJobs.cancelAll()
supervisorJob.cancel()
```

This will clear the list of jobs held inside `TrackKeyedJobs`

rahul-lohra:
There is a race during reconnection or migration: a media-state collector can trigger a new mute-sync job after `cancelActiveWork()` cancels the existing jobs. That job may send `UpdateMuteStates` to the old SFU connection.  
>   
> We should continue recording the user’s latest desired mute state during the transition, but defer synchronization until reconnection or migration completes and the active SFU is ready.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Good call — `cleanup()` now pauses mute sync and calls `muteStateSyncJobs.cancelAll()` so the map is cleared when the session dies, not only in `cancelActiveWork()`.

Shipped in 616ec1c7ab.

PratimMallick:
@rahul-lohra Agreed — cancelling the jobs was not enough, because `listenToMediaChanges` keeps running on `mediaScope` and could start a new `UpdateMuteStates` against the old SFU.

`setMuteState` still always writes the local `muteState` map. HTTP sync is gated by `muteSyncEnabled`: paused in `cancelActiveWork()` / `cleanup()`, then resumed from `connectRtc()` (`JoinCallResponse` + `whenConnected`) which flushes the latest bits to the active SFU. On migration that resume runs on the new session; the old one stays paused until cleanup.

Covered in 616ec1c7ab.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
```diff
@@ -1195,6 +1204,7 @@ public class RtcSession internal constructor(
 
     private suspend fun connectRtc() {
         logger.d { "[connectRtc] #sfu; #track; no args" }
+        resumeMuteSync()
```
aleksandar-apostolov, line 1207: `resumeMuteSync()` launches one `UpdateMuteStates` per recorded track, then `listenToMediaChanges()` two lines down re-collects the camera/mic/screenshare `StateFlow`s — `collectLatest` replays on re-subscription, so each replay reaches `syncMuteStateToSfu` for the same `TrackType` and cancels the flush job launched microseconds earlier.

So every reconnect gets a redundant in-flight RPC per track, cancelled mid-flight, logging the `IOException: Canceled` quoted in the Goal section. Not a correctness bug — the winning job carries the same value — but it puts the line back in customer logs.

Reordering won't help; the collectors launch on `mediaScope` and the interleaving is nondeterministic either way. Recording which tracks were actually deferred while paused and flushing only those avoids it, and also stops the flush posting for tracks that were never published when `publisher.value == null`.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/rtc/RtcSessionTest2.kt
```diff
@@ -1075,6 +1076,53 @@ class RtcSessionTest2 {
...
+        testScheduler.advanceUntilIdle()
+        coVerify(exactly = 0) { signalService.updateMuteStates(any()) }
+        assertEquals(false, muteSyncEnabled(rtcSession).get())
+
+        RtcSession::class.java.getDeclaredMethod("resumeMuteSync").apply {
+            isAccessible = true
+            invoke(rtcSession)
+        }
+
+        assertEquals(true, muteSyncEnabled(rtcSession).get())
```
aleksandar-apostolov, line 1123: This asserts the flag flipped, but not that anything was flushed. Gut the body of `resumeMuteSync()` down to `muteSyncEnabled.set(true)` and the test still passes — so it doesn't cover the half of the change the name refers to.

`coVerify(exactly = 1) { signalService.updateMuteStates(any()) }` after an `advanceUntilIdle()` would close it. The migration test above already uses `exactly = 0`, so the pattern's right there.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/RtcSession.kt
PratimMallick: Addressed in 5a5f0417ac.

`pendingMuteSyncTracks` records which tracks actually tried to POST while paused. `resumeMuteSync` snapshots that set, clears it, and flushes only those from `muteState` — not the pre-seeded audio/video/screen-share map.

That avoids posting unpublished tracks when `publisher.value == null`, and avoids racing `listenToMediaChanges` `collectLatest` replay (no redundant same-track jobs for the flush to cancel). A re-check after the add covers a resume that races the pause path: we fall through and post, and same-track `launch` coalesces with any flush already in flight.

Named it pending rather than deferred so it does not collide with `kotlinx.coroutines.Deferred`.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/rtc/RtcSessionTest2.kt
PratimMallick: Addressed in 5a5f0417ac.

The resume test now publishes audio while paused, asserts `pendingMuteSyncTracks` contains AUDIO, then `coVerify(timeout = 3000, exactly = 1)` on `updateMuteStates` after `resumeMuteSync`, and that the pending set is empty.

A sibling test covers the other half: resume with an empty pending set must not POST for the pre-seeded map.

Could not keep spy + `StandardTestDispatcher` + `advanceUntilIdle()` here — `flowOn(IO)` never hit the mock. The flush test uses a real `RtcSession` with stubbed socket/subscriber flows and Unconfined.

aleksandar-apostolov (approved):
LGTM

PR #1797, opened 2026-09-02
Avoid fetching FCM token during logout in demo-app

rahul-lohra:
### Goal
Closes #AND-1477
Delete stored FCM token during logout instead of fetching from FCM

### Implementation

Delete stored FCM token during logout instead of fetching from FCM

### 🎨 UI Changes

None

### Testing

Smoke test ringing notifications

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved logout cleanup by removing the first available registered device, helping ensure device registrations are cleared more reliably.
  - Updated logout behavior to avoid relying on Firebase push-token retrieval, reducing potential failures when push notification tokens are unavailable.


demo-app/src/main/kotlin/io/getstream/video/android/ui/join/CallJoinViewModel.kt
```diff
@@ -144,16 +141,9 @@ class CallJoinViewModel @Inject constructor(
...
             StreamVideo.instanceOrNull()?.let { streamVideo ->
-                fcmToken?.let { fcmToken ->
-                    streamVideo.deleteDevice(
-                        Device(
-                            id = fcmToken,
-                            pushProvider = PushProvider.FIREBASE.key,
-                            pushProviderName = "firebase",
-                        ),
-                    )
+                streamVideo.getDevice().first()?.let { device ->
```
aleksandar-apostolov: `getDevice()` is DataStore-backed, so `.first()` can throw `IOException` and nothing here catches it — if it does, `removeClient()` and `isLoggedOut = true` never run. Worth `runCatching { ... firstOrNull() }`? The deleted `fcmToken` had a try/catch for the same reason.

aleksandar-apostolov:
`CallLobbyViewModel.logOut()` is the other logout path and it does not call `deleteDevice` at all — still on the deprecated `StreamVideo.logOut()`, which only clears local storage. Intentional to leave it out of scope? That path leaves the server-side device row intact, so ringing pushes would still arrive after logging out from the lobby.

demo-app/src/main/kotlin/io/getstream/video/android/ui/join/CallJoinViewModel.kt
rahul-lohra: Corrected the code by catching `IOException` and replaced `first` with `firstOrNull` . Thank you

PratimMallick (approved):

rahul-lohra:
> CallLobbyViewModel.logOut() is the other logout path and it does not call deleteDevice at all — still on the deprecated StreamVideo.logOut(), which only clears local storage. Intentional to leave it out of scope? That path leaves the server-side device row intact, so ringing pushes would still arrive after logging out from the lobby.

Reverted and corrected the code

PR #1798, opened 2026-09-02
Surface coordinator error body from apiCall

PratimMallick:
### Goal
Fixes [AND-1478](https://linear.app/stream/issue/AND-1478/surface-coordinator-error-body-from-streamvideoclientapicall) — Coordinator rejections from `StreamVideoClient.apiCall` should expose the server message, not a bare HTTP status.

`parseError` consumes the response body. The old path used it only for the auth-retry check, then rethrew the raw `HttpException`, so callers only saw `HTTP 400`.

### Implementation
- Parse the error body once in `apiCall`.
- Use the parsed `NetworkError` for the auth-retry check.
- On non-auth failures, log and throw the coordinator message.
- Add unit tests for the surfaced message and the auth retry.

### Testing
- `./gradlew :stream-video-android-core:testDebugUnitTest --tests "io.getstream.video.android.core.StreamVideoClientTest"`
- `./gradlew spotlessApply`
- Trigger a coordinator 403/400 (for example a `user` calling `UpdateCall`) and confirm the log/result includes the server reason.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of network errors by surfacing clearer server-provided error messages.
  * Authentication failures now trigger a single token refresh and retry.
  * Prevented duplicate parsing of error responses for more reliable error handling.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
@@ -317,23 +317,27 @@ internal class StreamVideoClient internal constructor(
         try {
             apiCall()
         } catch (e: HttpException) {
-            // Retry once with a new token if the token is expired
-            if (e.isAuthError()) {
+            // parseError consumes the body, so this has to happen once. Rethrowing the raw
+            // HttpException would leave callers with only "HTTP 400" and drop the
+            // coordinator's actual reason.
+            val failure = parseError(e)
+            val networkError = failure.value as? Error.NetworkError
+            if (networkError != null && networkError.isAuthError()) {
                 val newToken = tokenProvider.loadToken()
                 tokenRepository.updateToken(newToken)
                 token = newToken
                 coordinatorConnectionModule.updateToken(newToken)
                 apiCall()
             } else {
-                throw e
+                val serverMessage = networkError?.message ?: e.message
+                logger.e { "[apiCall] HTTP ${e.code()}: $serverMessage" }
+                throw Exception(serverMessage, e)
             }
         }
     }
 
-    private fun HttpException.isAuthError(): Boolean {
-        val failure = parseError(this)
-        val parsedError = failure.value as Error.NetworkError
-        return when (parsedError.serverErrorCode) {
+    private fun Error.NetworkError.isAuthError(): Boolean {
+        return when (serverErrorCode) {
             VideoErrorCode.AUTHENTICATION_ERROR.code,
             VideoErrorCode.TOKEN_EXPIRED.code,
             VideoErrorCode.TOKEN_NOT_VALID.code,
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
@@ -302,7 +302,7 @@ internal class StreamVideoClient internal constructor(
      */
     internal suspend fun <T : Any> apiCall(
         apiCall: suspend () -> T,
-    ): Result<T> = safeSuspendingCallWithResult {
+    ): Result<T> {
         // Guest users have an asynchronous setup (createGuest) that fetches their JWT.
         // Any authenticated request that fires before that completes goes out without
         // an Authorization header and stream-auth-type "anonymous", so the backend
@@ -313,27 +313,56 @@ internal class StreamVideoClient internal constructor(
         // await throw if setupGuestUser failed — safeSuspendingCallWithResult turns
         // it into Result.Failure, which is the right outcome (the caller didn't get
         // a valid guest session, so the request can't proceed).
-        guestUserJob?.takeIf { currentCoroutineContext()[Job] !== it }?.await()
-        try {
-            apiCall()
-        } catch (e: HttpException) {
-            // parseError consumes the body, so this has to happen once. Rethrowing the raw
-            // HttpException would leave callers with only "HTTP 400" and drop the
-            // coordinator's actual reason.
-            val failure = parseError(e)
-            val networkError = failure.value as? Error.NetworkError
-            if (networkError != null && networkError.isAuthError()) {
-                val newToken = tokenProvider.loadToken()
-                tokenRepository.updateToken(newToken)
-                token = newToken
-                coordinatorConnectionModule.updateToken(newToken)
-                apiCall()
-            } else {
-                val serverMessage = networkError?.message ?: e.message
-                logger.e { "[apiCall] HTTP ${e.code()}: $serverMessage" }
-                throw Exception(serverMessage, e)
+        val prepared = safeSuspendingCallWithResult {
+            guestUserJob?.takeIf { currentCoroutineContext()[Job] !== it }?.await()
+            Unit
+        }
+        if (prepared is Failure) return prepared
+
+        return try {
+            try {
+                Success(apiCall())
+            } catch (e: HttpException) {
+                handleCoordinatorHttpFailure(e, apiCall)
+            }
+        } catch (e: Exception) {
+            safeSuspendingCallWithResult<T> { throw e }
+        }
+    }
+
+    /**
+     * parseError consumes the body, so each HttpException is parsed once. The failure
+     * keeps [HttpException] as [Error.ThrowableError.cause] so callers that classify
+     * HTTP errors by type still work. Wrapping it in a generic Exception would break
+     * that contract. The retry is parsed the same way — a second HttpException after
+     * token refresh must not fall through as a raw "HTTP 4xx".
+     */
+    private suspend fun <T : Any> handleCoordinatorHttpFailure(
+        first: HttpException,
+        apiCall: suspend () -> T,
+    ): Result<T> {
+        val firstError = parseError(first).value as Error.NetworkError
+        if (firstError.isAuthError()) {
+            val newToken = tokenProvider.loadToken()
+            tokenRepository.updateToken(newToken)
+            token = newToken
+            coordinatorConnectionModule.updateToken(newToken)
+            return try {
+                Success(apiCall())
+            } catch (retry: HttpException) {
+                coordinatorHttpFailure(retry)
             }
         }
+        return coordinatorHttpFailure(first, firstError)
+    }
+
+    private fun coordinatorHttpFailure(
+        exception: HttpException,
+        parsed: Error.NetworkError? = null,
+    ): Failure {
+        val networkError = parsed ?: parseError(exception).value as Error.NetworkError
+        logger.e { "[apiCall] HTTP ${exception.code()}: ${networkError.message}" }
+        return Failure(Error.ThrowableError(networkError.message, exception))
     }
 
     private fun Error.NetworkError.isAuthError(): Boolean {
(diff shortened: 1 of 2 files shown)


PratimMallick:
@coderabbit review

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -317,23 +317,27 @@ internal class StreamVideoClient internal constructor(
...
                 val newToken = tokenProvider.loadToken()
                 tokenRepository.updateToken(newToken)
                 token = newToken
                 coordinatorConnectionModule.updateToken(newToken)
                 apiCall()
             } else {
-                throw e
+                val serverMessage = networkError?.message ?: e.message
+                logger.e { "[apiCall] HTTP ${e.code()}: $serverMessage" }
+                throw Exception(serverMessage, e)
```
rahul-lohra: Request changes. The goal is correct, but the implementation introduces a silent exception-contract regression.

### [P0] `HttpException` becomes a generic `Exception`

At `StreamVideoClient.kt:334`, non-auth HTTP failures change from:

```text
ThrowableError.cause = HttpException
```

to:

```text
ThrowableError.cause = Exception
ThrowableError.cause.cause = HttpException
```

Any consumer doing this will silently stop recognizing HTTP failures:

```kotlin
(error as? Error.ThrowableError)?.cause is HttpException
```

That can break retry classification, status-code handling, and authentication/error telemetry. It also discards the parsed `NetworkError.serverErrorCode` and `statusCode`, retaining only its message.

This is public observable behavior because `stream-result` is exported as an API dependency. For a public SDK with backward-compatibility requirements, this is a merge blocker.

The test at `StreamVideoClientTest.kt:377` only checks the message. It completely misses the exception-type regression.

### [P1] The post-refresh request bypasses the new handling

At `StreamVideoClient.kt:330`, the retry happens from inside the `catch`. If that second request throws another `HttpException`, the same catch does not handle it. It goes directly to `safeSuspendingCallWithResult`, producing the old bare `"HTTP 4xx/5xx"` message.

So the PR does not consistently achieve its stated goal after token refresh.

### About `ImmediateEventDispatcher`

The `ImmediateEventDispatcher.shouldRetry()` demonstrates exactly why wrapping transport exceptions carelessly is dangerous: if it ever receives the new wrapper, HTTP 5xx errors become non-retryable.

Recommended fix: parse each attempt into a structured `Error.NetworkError` and preserve the original `HttpException` as its cause. If changing the existing `ThrowableError` subtype is considered too risky, retain `cause = e` and only replace the error’s displayed message.

The targeted `StreamVideoClientTest` and `ImmediateEventDispatcherTest` pass locally, but the missing type/status assertions mean those tests do not cover this regression.

PratimMallick: Good catch — both points were valid.

Kept `Error.ThrowableError` (existing `apiCall` contract / `CallCrudTest`) instead of switching the public subtype to `NetworkError`. The failure is now `Failure(Error.ThrowableError(serverMessage, httpException))`, so the coordinator message is surfaced and `cause` stays the original `HttpException`.

The post-refresh attempt is parsed the same way, so a second HTTP error no longer falls through as a raw `HTTP 4xx`. Tests now assert the cause type and cover retry-then-fail.

Fixed in 5faa7f6b73.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -313,27 +313,60 @@ internal class StreamVideoClient internal constructor(
...
+        return coordinatorHttpFailure(first, firstError)
+    }
+
+    private fun coordinatorHttpFailure(
+        exception: HttpException,
+        parsed: Error.NetworkError? = null,
+    ): Failure {
+        val networkError = parsed ?: parseError(exception).value as Error.NetworkError
+        logger.e { "[apiCall] HTTP ${exception.code()}: ${networkError.message}" }
+        return Failure(Error.ThrowableError(networkError.message, exception))
```
aleksandar-apostolov, line 365: **Non-JSON error bodies now produce a worse message than before this PR.**

`coordinatorHttpFailure` uses `networkError.message` unconditionally, but when the body isn't Stream JSON `parseError` never reaches the branch that reads `error.message` — it returns one of two fallbacks, both of which say `"failed to parse error response from server"`. (Empty bodies land there too: `errorBody()?.bytes()` returns a zero-length array, not `null`, so `decodeFromString("")` throws into the parse-failure branch.)

So for a 502/504/captive-portal response the caller used to get `"HTTP 502 Bad Gateway"` and now gets `"failed to parse error response from server: Expected start of the object '{' ..."`. The status code is gone from the message entirely — strictly less useful than what was there before, on the failure mode users hit most.

Suggest falling back to `"HTTP ${exception.code()} ${exception.message()}"` when the parse didn't produce a real coordinator error. Watch out that the no-body branch sets `serverErrorCode = e.code()` rather than `PARSER_ERROR`, so both fallbacks need covering, not just the `PARSER_ERROR` one.

Worth a test that throws an `HttpException` with a 502 and an HTML body, and asserts the status still appears in the resulting `ThrowableError.message`. That case is also most of the gap to the 80% Sonar gate this PR is currently under.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -313,27 +313,60 @@ internal class StreamVideoClient internal constructor(
...
+     * that contract. The retry is parsed the same way — a second HttpException after
+     * token refresh must not fall through as a raw "HTTP 4xx".
+     */
+    private suspend fun <T : Any> handleCoordinatorHttpFailure(
+        first: HttpException,
+        apiCall: suspend () -> T,
+    ): Result<T> {
+        val firstError = parseError(first).value as Error.NetworkError
+        if (firstError.isAuthError()) {
+            val newToken = tokenProvider.loadToken()
```
aleksandar-apostolov, line 346: **A failing `loadToken()` swallows the coordinator message.**

If the customer's `TokenProvider` throws here, the exception escapes to the outer `catch (e: Exception)` in `apiCall` and the caller ends up with `"Safe call failed with <token provider message>"`. The original 401 reason — the thing this PR exists to surface — is discarded, on the one path where the SDK definitely has it.

Wrapping the refresh and falling back to `coordinatorHttpFailure(first, firstError)` keeps the coordinator message and still reports the refresh failure via the log.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -373,23 +406,30 @@ internal class StreamVideoClient internal constructor(
...
+                cause = e,
+            ),
         )
         return Failure(
             Error.NetworkError(
                 message = error.message,
                 serverErrorCode = error.code,
                 statusCode = error.statusCode,
-                cause = Throwable(error.moreInfo),
+                cause = e,
```
aleksandar-apostolov, line 432: **Two pieces of the parsed error end up unreachable by callers.**

Switching `cause` to `e` is right — that's what fixes @rahul-lohra's classification point — but `moreInfo` isn't kept anywhere else, and the conversion to `ThrowableError` keeps only the message, so `serverErrorCode` goes too. That half of rahul's comment is still open, and `parseError` consumes the body so callers can't re-derive it.

If keeping `ThrowableError` is the decision — fair, `CallCrudTest` depends on it — folding both into the message (`"[<code>] <message> (<moreInfo>)"`) at least keeps them recoverable to a human reading logs.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
PratimMallick: Good catch. Unparsed / empty / missing bodies now fall back to `HTTP <code> <reason>`, so a 502 still shows the status instead of a JSON parse error.

Both fallbacks are covered: parse failures stay `PARSER_ERROR`, and a missing body uses `NO_ERROR_BODY` (not the HTTP status as `serverErrorCode`). Tests cover HTML 502, an empty body, and a null response.

Fixed in d0d6e043f0.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
PratimMallick: Wrapped the refresh. If `loadToken()` throws we log it and return `coordinatorHttpFailure(first, firstError)`, so the original 401 stays on the result instead of `Safe call failed with …`. `CancellationException` is rethrown. Added a test for a throwing `TokenProvider`.

Fixed in d0d6e043f0.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
PratimMallick: Kept `ThrowableError` and folded both into the message as `[<code>] <message> (<moreInfo>)`, since the parsed body is already consumed. Tests assert the code prefix and `moreInfo`.

Fixed in d0d6e043f0.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/StreamVideoClient.kt
```diff
@@ -373,27 +442,51 @@ internal class StreamVideoClient internal constructor(
...
+     *
+     * [Error.ThrowableError] only stores a message, so the server error code
+     * and moreInfo are included here for callers and logs.
+     */
+    private fun formatCoordinatorErrorMessage(error: ErrorResponse): String {
+        val moreInfo = error.moreInfo.takeIf { it.isNotBlank() }
+        return buildString {
+            append("[${error.code}] ${error.message}")
+            if (moreInfo != null) {
+                append(" ($moreInfo)")
```
rahul-lohra, line 485: This PR introduces a new product-and-platform-specific error-message template:

```text
[<error code>] <server error message> (<more info>)
```

For example:

```text
[4] User is not allowed to perform this action (https://getstream.io/errors/4)
```

Before considering consistency across platforms, I think we should first keep error handling consistent across our Android products.

Currently, the Android Chat SDK exposes the server message without prefixing it with the error code:

```text
<server error message>
More information available at <more info>
```

For example:

```text
User is not allowed to perform this action
More information available at https://getstream.io/errors/4
```

I also checked how the Video SDK exposes the equivalent error on other platforms:

**JavaScript**

```text
Stream error code <error code>: <server error message>
```

Example:
```text
Stream error code 4: User is not allowed to perform this action
```

JavaScript additionally exposes the code, HTTP status and response as structured properties.

**Swift**
```text
<server error message>
```
Example:

```text
User is not allowed to perform this action
```
Swift exposes the error code, HTTP status, `moreInfo` and other metadata as separate properties on `APIError`.

This is from android feeds - (this is also not standardised)
```text
Failed request: ${chain.request().url}
```


My recommendation for this PR 
1. Drop the `[<error code>]` prefix and return the server message without embedding the code into it. This would keep Video Android closer to the existing Android Chat behavior. Or we can use chat template ~ like add a new line before appending `<more-info>` text

2. The error code and HTTP status are structured metadata and should eventually be exposed as properties of a shared Android error type rather than encoded into `message`. Coordinator error parsing and representation could then be moved into the shared Android result/core module so that Chat and Video cannot introduce separate conventions. cc @aleksandar-apostolov 

This matters because customers frequently integrate both the Chat and Video Android SDKs. Equivalent backend failures should not produce two different Android-specific error formats depending on which Stream product generated them.

PratimMallick:
On further analysis this looks like a problem to be solved in android-core lib rather than on video sdk level. 
Closing this PR now and we will fix this in v2 where core is already integrated and if core doesnt fix it yet

PR #1801, opened 2026-09-02
End-to-end encryption for call media

PratimMallick:
### Goal
Closes [AND-1480](https://linear.app/stream/issue/AND-1480/end-to-end-encryption-for-android-video-call-media) — Apps can encrypt call media with framed AES-GCM so the SFU forwards frames it cannot read.

The app owns `StreamEncryptionManager` and its keys. The SDK attaches encryptors/decryptors to media and sends the E2EE intent on join. Encryption mode remains a server-side call/type setting.

### App integration
Create and configure the manager before joining, handle the Kotlin `Result` values, retain the manager for key rotation and eventual disposal, and register an event listener for runtime encryption health:

```kotlin
val manager = StreamEncryptionManager.create(user.id).getOrElse { error ->
    // E2EE is unavailable or native manager creation failed.
    return
}

manager.setEventListener { event ->
    when (event.type) {
        E2EEEventType.DECRYPTION_FAILED,
        E2EEEventType.DECRYPTION_STALLED,
        E2EEEventType.ENCRYPTION_FAILED,
        E2EEEventType.MISSING_KEY,
        E2EEEventType.UNENCRYPTED_FRAME,
        E2EEEventType.UNSUPPORTED_VERSION,
        -> showEncryptionWarning(event)

        E2EEEventType.DECRYPTION_RESUMED -> clearEncryptionWarning(event)
        else -> logE2EEEvent(event)
    }
}

manager.setSharedKey(keyIndex = 0, key = keyBytes)

call.setE2EEManager(manager).getOrElse { error ->
    manager.dispose()
    return
}

call.join()
```

The listener runs on a WebRTC internal thread, so apps must switch to their UI dispatcher before updating UI state. Runtime events report frame-level conditions after encryptors/decryptors are attached: failed or stalled decryption, resumed decryption, failed encryption, missing keys, cleartext frames, unsupported frame versions, requested key state, and optional performance reports. `CallState.e2eeEnabled` remains the manager attachment state and is not changed by individual runtime events.

For per-user keys, use `setKey(userId, keyIndex, key)`. Rotate keys during a call by writing the next key index; remove old keys with `removeSharedKey` or `removeKey` when they are no longer needed.

The app owns the manager lifecycle. `Call.leave()` detaches the Call reference but does not dispose the manager. Keep it in an app-level owner and call `dispose()` only when no call/session can still use it, such as final app-session or logout cleanup.

Apps normally do not call `encrypt()` or `decrypt()`. Those are SDK callbacks for custom `E2EEManager` implementations and return `kotlin.Result<Unit>` so failed attachment cannot be treated as success.

### New APIs
- `E2EEManager` — custom encryption plug-in with `encrypt(...)` and `decrypt(...)` attachment callbacks.
- `StreamEncryptionManager.create(...)` — creates the default WebRTC AES-GCM manager and returns `kotlin.Result<StreamEncryptionManager>`.
- `StreamEncryptionManager.isSupported()` — optional runtime capability check.
- `StreamEncryptionManager.setSharedKey(...)`, `setKey(...)`, `removeSharedKey(...)`, `removeKey(...)`, and `removeAllKeys(...)` — key management and rotation.
- `StreamEncryptionManager.setEventListener(...)`, `requestKeyState()`, and `enablePerformanceReporting(...)` — runtime failure, state, and performance diagnostics.
- `StreamEncryptionManager.dispose()` — releases native encryption resources; app-owned.
- `Call.setE2EEManager(...)` — attaches or detaches a manager before join and returns `kotlin.Result<Unit>`.
- `CallState.e2eeEnabled` — observable manager attachment state.
- `E2EEAlgorithm`, `E2EETrackType`, `E2EEEvent`, `E2EEEventType`, `E2EEKeyState`, and `E2EEPerformance` — public E2EE configuration and event models.

### Implementation
- Uses stable `stream-video-webrtc-android:145.17.0` with AES-128/256-GCM framed encryption.
- First join passes `e2ee` from call state. Rejoin/migrate derive it from the attached manager.
- Publisher refuses to cache/negotiate a sender when encryptor attachment fails.
- Subscriber retries tracks whose decryptor attachment fails.
- Demo lobby provides a passphrase lock using the same PBKDF2 parameters as Pronto and logs all native E2EE events with warning severity for runtime failures.
- Binary-compatibility API dump covers all new public types.

### Testing
- `./gradlew :stream-video-android-core:testDebugUnitTest --tests "io.getstream.video.android.core.e2ee.*"`
- `./gradlew :stream-video-android-core:testDebugUnitTest --tests "io.getstream.video.android.core.call.components.CallApiClientTest" --tests "io.getstream.video.android.core.call.components.CallJoinCoordinatorTest"`
- Pre-push `spotlessCheck` and `apiCheck`
- Manual: enable the lobby lock with the same Pronto `encryption_key` passphrase, join an `auto-on` call, confirm `e2ee=true`, verify encrypted media in both directions, and inspect runtime E2EE events in Logcat.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end encryption support for calls, including shared-key management and encryption status tracking.
  * Added an encryption toggle to the Android demo lobby with passphrase protection and support checks.
  * Added encryption and decryption for outgoing and incoming audio, video, and screen-sharing media.
  * Added encryption events, key-state reporting, and performance monitoring APIs.
* **Bug Fixes**
  * Prevented media from being published without encryption when encryption setup fails.
* **Tests**
  * Added coverage for encryption setup, media handling, join flows, and native type mappings.


PratimMallick:
@coderabbit review

demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyE2EE.kt
```diff
@@ -0,0 +1,194 @@
...
+ */
+@Composable
+internal fun E2EELobbyButton(call: Call, modifier: Modifier = Modifier) {
+    val encrypted by call.state.e2eeEnabled.collectAsStateWithLifecycle()
+    var showDialog by remember { mutableStateOf(false) }
+    var error by remember { mutableStateOf<String?>(null) }
+    // Held here so the demo can dispose it when the user turns encryption off. Do not dispose
+    // when this composable leaves the tree: Join starts CallActivity with CLEAR_TASK, which
+    // destroys the lobby and would free the native manager the Call still holds.
+    var manager by remember { mutableStateOf<StreamEncryptionManager?>(null) }
```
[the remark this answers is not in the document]
PratimMallick: Fixed in bbfb29a2e4. The ViewModel now owns the StreamEncryptionManager across composition recreation. Disabling E2EE clears the Call reference first, then disposes the app-owned manager and clears the ViewModel reference in a finally block.

demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyE2EE.kt
```diff
@@ -0,0 +1,194 @@
...
+                    val failure = runCatching {
+                        check(StreamEncryptionManager.isSupported()) {
+                            "This WebRTC build has no end-to-end encryption support."
+                        }
+                        // The key has to be in place before the manager reaches the call, so that
+                        // the publisher never gets a chance to send a frame without one.
+                        // Encryption mode is a call/type setting, not something a regular user
+                        // can flip with UpdateCall — Pronto's `user` role is denied that.
+                        val created = StreamEncryptionManager.create(call.user.id)
+                        created.setSharedKey(KEY_INDEX, deriveE2EEKey(passphrase))
```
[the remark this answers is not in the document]
PratimMallick: Fixed in bbfb29a2e4. PBKDF2 key derivation now runs in the ViewModel with withContext(Dispatchers.Default), off the composition dispatcher.

demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyE2EE.kt
```diff
@@ -0,0 +1,194 @@
...
+                        modifier = Modifier.padding(horizontal = 16.dp),
+                        text = error,
+                        style = VideoTheme.typography.bodyS,
+                        color = VideoTheme.colors.alertWarning,
+                    )
+                }
+            }
+        },
+        positiveButton = Triple("Enable", ButtonStyles.secondaryButtonStyle()) {
+            onConfirm(passphrase.text)
```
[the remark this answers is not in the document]
PratimMallick: Fixed in bbfb29a2e4. The dialog now rejects blank passphrases, displays a field error, and invokes onConfirm only for non-blank input.

gradle/libs.versions.toml
```diff
@@ -49,7 +49,9 @@ tink = "1.9.0"
 turbine = "0.13.0"
 itu = "1.7.3"
 
-streamWebRTC = "145.6.0"
+# TODO(e2ee): snapshot needed for org.webrtc.EncryptionManager (GetStream/webrtc#110). Pin to a
+# release before merging.
+streamWebRTC = "145.16.0-SNAPSHOT"
```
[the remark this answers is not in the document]
PratimMallick: Fixed in bbfb29a2e4. Updated stream-video-webrtc-android to the stable 145.17.0 release and removed the snapshots-only Maven repository.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/e2ee/StreamEncryptionManager.kt
```diff
@@ -0,0 +1,237 @@
...
+    private inline fun ifActive(operation: String, block: () -> Unit) {
+        if (native.isDisposed) {
+            logger.w { "[$operation] ignored, manager is disposed" }
+            return
+        }
+        try {
+            block()
+        } catch (e: Throwable) {
+            logger.e(e) { "[$operation] failed" }
+        }
```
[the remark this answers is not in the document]
PratimMallick: Fixed in bbfb29a2e4. StreamEncryptionManager.encrypt now propagates native attachment failures. Publisher.attachEncryptor returns false on that failure, stops the transceiver, and does not cache it, preventing negotiation without an encryptor.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/e2ee/CallE2EETest.kt
```diff
@@ -0,0 +1,213 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.webrtc.RtpReceiver
+import org.webrtc.RtpSender
+
+/**
+ * Covers the [Call]-level end-to-end encryption surface: attaching a manager, the guard that keeps
+ * it before join, the key conveniences, and the flag the coordinator validates the join against.
+ */
+class CallE2EETest {
```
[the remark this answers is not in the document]
PratimMallick: I tested this migration, but TestBase is not compatible with this JVM-only test: its constructor calls ApplicationProvider and all six tests fail with “No instrumentation registered”. Keeping the existing DispatcherRule fixture so CallE2EETest remains a fast JVM test.

rahul-lohra:
Can you write how to use the new api?
Any best practices?
What are the new public apis now?

aleksandar-apostolov (commented):
Read through the E2EE path and ran the test command from the description. A few things inline — the failing tests and the subscriber re-attach are the two I'd want sorted before merge. Good call getting the trailer format into the fork rather than reusing the LiveKit transformer; that's what keeps web interop working.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinatorTest.kt
```diff
@@ -141,7 +142,7 @@ class CallJoinCoordinatorTest {
     private fun stubJoinCall(result: io.getstream.result.Result<JoinCallResponse>) {
         coEvery {
             apiClient.joinRequest(
-                any(), any(), any(), any(), any(), any(), any(), any(),
+                any(), any(), any(), any(), any(), any(), any(), any(), any(),
```
aleksandar-apostolov, line 145: `stubJoinCall` picked up the ninth `any()` but the six `coVerify` blocks further down didn't, so they bind a literal `null()` matcher to `e2ee` while the coordinator actually passes `false`. I get 6 failures here on this branch and a clean run on develop — can you re-run the command from the description?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
```diff
@@ -676,4 +692,73 @@ internal class Subscriber(
...
 
     fun isEnabled() = enabled
+
+    // region End-to-end encryption
+
+    /** Tracks whose owning session had no known user yet, keyed by track id. */
+    private val pendingDecryptors = ConcurrentHashMap<String, Pair<String, TrackType>>()
+
+    /** Track ids already handed to the manager, so re-delivered streams don't attach twice. */
+    private val decryptedTrackIds = ConcurrentHashMap.newKeySet<String>()
```
aleksandar-apostolov, line 704: `decryptedTrackIds` is only cleared in `clear()`, but `onRemoveStream` drops the other track maps. So if the SFU re-adds a track it removed earlier, this early-returns and the new receiver never gets a decryptor — that participant stays undecodable for the rest of the call. I reproduced it by adding a remove/re-add to `E2EEMediaAttachmentTest`; the second attach never happens. Would keying the dedupe on the receiver rather than the track id sort it?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -614,9 +688,15 @@ public class Call(
     // endregion
 
     @InternalStreamVideoApi
-    fun leave(reason: CallLeaveReason) = lifecycle.leave(reason)
+    fun leave(reason: CallLeaveReason) {
+        detachE2EEManager()
```
aleksandar-apostolov, line 699: `join()` calls `resetLeaveGuard()`, so leave-then-join on the same Call is a supported path — but the manager is gone by then, and the second join goes out with `e2ee=false` and the lock indicator off. The KDoc on `setE2EEManager` says the manager only needs to be set once. Should leave hold onto it, or should the doc call this out?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/e2ee/StreamEncryptionManager.kt
```diff
@@ -0,0 +1,263 @@
...
+    public fun removeAllKeys(userId: String) {
+        ifActive("removeAllKeys") { native.removeAllKeys(userId) }
+    }
+
+    /**
+     * Observes encryption state changes, most importantly decryption failures. Pass `null` to stop
+     * observing. The listener is invoked on a WebRTC internal thread, so hop to your own
+     * dispatcher before touching UI state.
+     */
+    public fun setEventListener(listener: E2EEEventListener?) {
```
aleksandar-apostolov, line 191: The native side forwards frames with no trailer as cleartext and just fires `unencrypted_frame`. Nothing in the SDK listens for it, so `e2eeEnabled` stays true while cleartext renders — and neither this example nor the integration snippet in the description ever wires a listener. Should the SDK observe this itself and reflect it in call state?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Publisher.kt
```diff
@@ -412,6 +416,16 @@ internal class Publisher(
                 ),
             )
             applyDegradationPreference(transceiver, publishOption)
+            if (!attachEncryptor(transceiver, publishOption)) {
+                // Caching this transceiver would publish plaintext on a call the app believes is
+                // encrypted, so drop it instead. Stop only — the PeerConnection owns the native
+                // transceiver and disposing here is a use-after-free on network_thread.
+                logger.e {
```
aleksandar-apostolov, line 423: Dropping the transceiver instead of publishing in the clear is the right call. But it's log-only — the user joins, nobody hears them, and there's nothing for the UI to show. Worth surfacing an event here?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
PratimMallick: `resetLeaveGuard()` only re-arms the one-shot guard that prevents duplicate teardown calls; it does not make a destroyed `Call` reusable. `leave()` is terminal for that instance: it marks the lifecycle as destroyed, removes the call from the client cache, cancels its scopes, and cleans up its media/session resources. A subsequent `StreamVideo.call(type, id)` creates a new `Call` instance.

The KDoc’s “survives rejoins and migrations” refers to the internal reconnect paths, which do not call `leave()`. We intentionally detach the app-owned manager on terminal `leave()` without disposing it, so the app can attach that manager to a newly created call if desired.

aleksandar-apostolov: Right.

stream-video-android-core/src/test/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinatorTest.kt
PratimMallick: Fixed. The remaining `coVerify` / `coEvery` `joinRequest` matchers now include the ninth `e2ee` argument, so they no longer bind `null()` while the coordinator passes `false`. `CallJoinCoordinatorTest` is green locally.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Subscriber.kt
PratimMallick: Good catch. `onRemoveStream` now clears `decryptedTrackIds` and `pendingDecryptors` for the removed track, so a later re-add can attach a decryptor on the new receiver. Added a remove/re-add regression in `E2EEMediaAttachmentTest`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/e2ee/StreamEncryptionManager.kt
PratimMallick: Agreed that cleartext frames should be visible to the app, but `e2eeEnabled` should stay as manager-attachment state, not per-frame health. Native already reports `UNENCRYPTED_FRAME` (and the other runtime events) through `setEventListener`; JS and iOS do the same rather than flipping the enabled flag.

The demo now registers that listener and logs the events, the PR description shows the integration snippet, and `E2EEEventType` has KDoc for what each event means. Apps can surface a warning from `UNENCRYPTED_FRAME` without implying the call is no longer encrypted.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/connection/Publisher.kt
PratimMallick: Keeping this log-only for now, deliberately.

On the media side we checked the other SDKs and this matches them. JS calls `e2ee.encrypt(...)` before `negotiate()` with no try/catch, so a throw skips negotiation and no `SetPublisher` goes out. iOS is moving to remove the track and clean up the transceiver on attach failure. So "refuse to publish rather than publish in the clear" is the agreed behaviour across all three.

The gap you spotted is real though: `setEventListener` cannot cover it, because native `ENCRYPTION_FAILED` only fires once an encryptor is actually attached and a frame fails. Attach-time failure never reaches native, so there is nothing for the app to observe.

The fix would be an SDK-synthesized event (something like `ENCRYPTOR_ATTACH_FAILED` emitted from `StreamEncryptionManager.encrypt()`), but neither JS nor iOS has an equivalent, so adding it here would put an Android-only value on an enum we have kept deliberately aligned. Would rather agree the shape cross-platform first than diverge in this PR. Note that neither JS nor iOS informs the app of attach failure today either.

aleksandar-apostolov (approved):

PR #1810, opened 2026-09-04
[Telecom-4] Stabilize notification-owned incoming ringing

rahul-lohra:
### Goal
Closes: #[AND-1493]

### Implementation
- Suppress semantically duplicate incoming-call notification updates with an internal comparator
- Preserve the original ringing channel across unaccepted-call updates
- Defer only updates that are not duplicates
- Avoid cancelling the Android 17 ringing notification from the dismiss trampoline as it   stops the ringing from notification

### Public API changes
- Adds `NotificationConfig.incomingRingingNotificationUpdateDelayMillis: Long`, defaulting to `1_300L` which will used for Android 17 and higher otherwise the ringing sound can get muted

### Stack

PR 4 of 5. Depends on #1809.

### Testing

- `NotificationUpdateDeduplicatorTest`
- `CallServiceNotificationUpdateObserverTest`
- `StreamDefaultNotificationHandlerTest`

## Summary by CodeRabbit

* **New Features**
  * Improved incoming-call notification handling on Android 17 and later.
  * Duplicate notification updates are now suppressed to reduce unnecessary refreshes.
  * Incoming ringing notifications retain their alert behavior and existing channel during updates.
  * Added a configurable delay before replacing active incoming-call notifications, defaulting to 1.3 seconds.


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/NotificationConfig.kt
```diff
@@ -52,4 +52,10 @@ public data class NotificationConfig(
      */
     @ExperimentalStreamVideoApi
     val notificationUpdateTriggers: (Call) -> Flow<*>? = { _ -> null },
+    /**
+     * Delay before replacing an unaccepted incoming-call notification whose channel owns ringing.
+     * Android may stop an insistent alert when that notification is immediately replaced.
+     * Set this to `0` to disable the delay.
+     */
+    val incomingRingingNotificationUpdateDelayMillis: Long = 1_300L,
```
aleksandar-apostolov: Adding this as a constructor parameter changes the generated `<init>` and `copy` signatures on a public data class. The api dump replaces the old ones rather than adding to them, so `apiCheck` stays green while an app compiled against the current release hits `NoSuchMethodError` on `NotificationConfig(...)` or `.copy(...)`.

Could this move into the class body instead? That leaves the primary constructor and `copy` intact — same shape we used for `Error.GenericError.code`.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/DismissNotificationActivity.kt
```diff
@@ -21,12 +21,17 @@ import android.content.Context
...
+import io.getstream.video.android.core.utils.isAndroid17OrHigher
 
 internal class DismissNotificationActivity : Activity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        NotificationManagerCompat.from(this)
-            .cancel(intent.getIntExtra(KEY_NOTIFICATION_ID, 0))
+        // On Android 17 the full-screen trampoline must not cancel the notification that owns
+        // the incoming ringtone. Accept/reject handling remains responsible for dismissing it.
+        if (!isAndroid17OrHigher()) {
```
aleksandar-apostolov: The comment scopes this to the notification that owns the incoming ringtone, but the guard is on the OS version, and this activity is the trampoline leg for the incoming, outgoing, notification-call, missed-call, live-call and ongoing-call intents alike. On Android 17 none of those get cancelled any more, so missed-call notifications will accumulate.

Should this key off the notification id instead of the version?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceNotificationUpdateObserver.kt
```diff
@@ -48,6 +51,16 @@ internal class CallServiceNotificationUpdateObserver(
 ) {
 
     private val logger by taggedLogger("NotificationUpdateObserver")
+    private val incomingRingingNotificationStabilization by lazy {
+        scope.async {
+            val delayMillis = streamVideo.streamNotificationManager
+                .notificationConfig
+                .incomingRingingNotificationUpdateDelayMillis
+            if (delayMillis > 0 && call.state.notificationIdFlow.value != null) {
```
aleksandar-apostolov: Two things on the stabilization delay. The `notificationIdFlow` check runs once inside the lazy block, at whatever moment the first update happens to arrive — if it is still null then, the delay is disabled for the rest of the call. Is that intended?

And is 1300 ms derived from something measurable, or tuned until it stopped reproducing? It will need to hold on slower devices too.

rahul-lohra: Corrected the logic - 
The core idea was to delay the incoming call notification updates only

The current code is like this

```kotlin

private suspend fun delayIncomingRingingNotificationUpdate() { //Replaced the lazy variable with function
        val delayMillis = streamVideo.streamNotificationManager
            .notificationConfig
            .incomingRingingNotificationUpdateDelayMillis
        if (delayMillis > 0) {
            delay(delayMillis)
        }
    }
    
private suspend fun updateNotification(context: Context) {
       ....
        val notification = streamVideo.onCallNotificationUpdate(call)        
        if (notification != null) {
           ...
            if (
                shouldStabilizeIncomingRingingNotification(ringingState) &&
                call.state.notificationIdFlow.value != null //updated logic
            ) {
                delayIncomingRingingNotificationUpdate()
            }
            showNotificationForState(context, ringingState, notification)
        } 
        ...
    }
    
private fun shouldStabilizeIncomingRingingNotification(ringingState: RingingState): Boolean =
        ringingState is RingingState.Incoming &&
            !ringingState.acceptedByMe &&
            isAndroid17OrHigher()

 ```

Comment (rahul-lohra) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/service/observers/CallServiceNotificationUpdateObserver.kt]: > And is 1300 ms derived from something measurable, or tuned until it stopped reproducing? It will need to hold on slower devices too.


Android says it may drop frequent notification updates when several updates happen within one second: [Android documentation](https://developer.android.com/develop/ui/compose/notifications/create-notification#Update).

Based on this documentation, Codex recommended avoiding updates during the first second. This is not an official Android recommendation.

During testing, immediate updates stopped the ringing sound. I chose 1.3 seconds to add a small safety margin. This value is based on testing and is not guaranteed by Android.

File: stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/NotificationConfig.kt (hunk 1 above)
Comment (rahul-lohra) [reply stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/NotificationConfig.kt]: Corrected it, now the new code is like this

```kotlin
public data class NotificationConfig(
....
    @ExperimentalStreamVideoApi
    val notificationUpdateTriggers: (Call) -> Flow<*>? = { _ -> null },
) {
    /**
     * Delay before replacing an unaccepted incoming-call notification whose channel owns ringing.
     * Android may stop an insistent alert when that notification is immediately replaced.
     * Set this to `0` to disable the delay.
     */
    public var incomingRingingNotificationUpdateDelayMillis: Long = 1_300L // New property with private setter
        private set

    public constructor(
       ...
        incomingRingingNotificationUpdateDelayMillis: Long, // New property in secondary constructor
    ) : this(
        ...
        notificationUpdateTriggers = notificationUpdateTriggers,
    ) {
        this.incomingRingingNotificationUpdateDelayMillis = incomingRingingNotificationUpdateDelayMillis
    }
}
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/DismissNotificationActivity.kt
rahul-lohra: We removed the Android 17-specific logic from `DismissNotificationActivity` because the activity cannot distinguish between a system-triggered full-screen launch and an explicit user action.

Previously, the incoming-call full-screen `PendingIntent` included `DismissNotificationActivity`. Because Android can launch a full-screen intent automatically, this canceled the incoming-call notification without user interaction. On Android 17, the notification owns the ringtone, so canceling it also stopped the ringing prematurely.

The **_incoming-call full-screen `PendingIntent`_** now directly targets the incoming-call activity and bypasses `DismissNotificationActivity`. The existing _**content intent**_ continues to use the dismissal flow for user-initiated notification taps. 
This keeps automatic full-screen launches from canceling the ringing notification while preserving intentional dismissal behavior.

The new code is added here

```kotlin
class StreamDefaultNotificationHandler {

    private fun searchIncomingCallFullScreenPendingIntent(
        callId: StreamCallId,
        payload: Map<String, Any?>,
        contentPendingIntent: PendingIntent?,
    ): PendingIntent? =
        if (isAndroid17OrHigher() && intentResolver is DefaultStreamIntentResolver) {
            intentResolver.searchIncomingCallFullScreenPendingIntent(
                callId = callId,
                payload = payload,
            )
        } else {
            contentPendingIntent
        }
        
}

Class DefaultStreamIntentResolver {
    /**
     * Creates the incoming-call full-screen intent without the notification-dismiss trampoline.
     * Android can launch this intent without user interaction, so launching it must not dismiss a
     * notification that owns the incoming ringtone.
     */
    internal fun searchIncomingCallFullScreenPendingIntent(
        callId: StreamCallId,
        notificationId: Int = callId.getNotificationId(NotificationType.Incoming),
        payload: Map<String, Any?>,
    ): PendingIntent? {
        ...
    }
}
```        

Since this issue is caused specifically by the dismissal trampoline used by `DefaultStreamIntentResolver`, the fix is intentionally scoped to that implementation. 
Custom resolvers control their own PendingIntent behavior and are not automatically affected by this change. We can introduce a separate public capability interface later if custom full-screen intent configuration is requested.

PR #1813, opened 2026-09-07
Distribute v2 demo app builds to Firebase App Distribution

andremion:
### Goal

Closes [AND-1497](https://linear.app/stream/issue/AND-1497/distribute-v2-demo-app-builds-via-firebase-app-distribution)

Nothing was distributed from `develop-v2`, so the team had no way to test v2 work. `app-distribute.yml` (Firebase) triggers on `main` only, and `internal-app-distribute.yml` (Google Play) triggers on `develop` and `main`. This adds a Firebase App Distribution pipeline for `develop-v2`.

### Implementation

- New `.github/workflows/app-distribute-v2.yml`. It triggers on pushes to `develop-v2`, runs `:demo-app:assembleDevelopmentRelease`, and uploads the APK to the `stream-testers` group in the shared `stream-android-32afb` Firebase project, the same project the chat SDK uses.
- The App ID and the service account credentials come from two new repo secrets, `FIREBASE_V2_DEMO_APP_ID` and `STREAM_ANDROID_FIREBASE_CREDENTIALS_APP_CONTENT`. Both are already set.
- `demo-app/google-services.json` is not changed. It stays on `stream-video-9b586` so push and Crashlytics keep working. App Distribution only needs the APK package name to match the app registered in the project.
- The workflow is hardened beyond a copy of the existing one: it requests only `contents: read`, reads the gpg passphrase from stdin rather than the command line, references secrets as shell variables instead of expanding them into the run script, deletes the decrypted signing material before the third-party Firebase action runs, and pins `actions/upload-artifact` to a commit SHA.
- One commit fixes the `Upload APK` step in `app-distribute.yml`. It pointed at `build/outputs/apk/demo-app/release`, but AGP writes to `build/outputs/apk/<flavor>/<buildType>`, so that directory never existed. `upload-artifact` only warns when it finds no files, so the step passed while uploading nothing.

Follow-up for when v2 becomes develop: delete `app-distribute-v2.yml`, move the new App ID and credentials into `app-distribute.yml`, and add `develop` to its trigger. The chat SDK skipped this cleanup and still has a dead `app-distribute-v7.yml` pointing at a branch that no longer exists.

### Testing

The pipeline was run end to end before merging. The branch was temporarily added to the `push` trigger, the workflow ran, and the temporary trigger was then reverted. This was necessary because `workflow_dispatch` is not available until the workflow file reaches the default branch, and the `push` trigger is otherwise limited to `develop-v2`.

Run [34132353521](https://github.com/GetStream/stream-video-android/actions/runs/34132353521) passed every step:

- The keystore and keystore properties decrypted with the new stdin passphrase handling.
- `:demo-app:assembleDevelopmentRelease` built and signed the APK.
- `Remove signing material` ran before the Firebase step.
- `Upload APK` produced a 72,520,380 byte artifact using only `contents: read`, which also confirms the corrected artifact path. The previous path yielded 0 artifacts on every run.
- Firebase reported `uploaded new release 1.32.0 (1)` on `io.getstream.video.android.dogfooding` in `stream-android-32afb`, followed by `distributed to testers/groups successfully`.

Note that `stream-testers` received a real build from that verification run, and will receive another when this merges.

Also verified locally: `./gradlew spotlessCheck` passes, and both workflow files parse as YAML. CI runs `actionlint` over them as well.

`detekt`, `apiCheck`, and `testDebugUnitTest` were not run locally. The diff is two workflow YAML files with no Kotlin or API surface changes, and CI runs them anyway.

### DIFF
--- a/.github/workflows/app-distribute-v2.yml
+++ b/.github/workflows/app-distribute-v2.yml
@@ -13,7 +13,6 @@ on:
 jobs:
   build_v2_demo_app:
     permissions:
-      actions: write
       contents: read
     name: Build and Distribute V2 Demo App
     runs-on: ubuntu-24.04
@@ -27,15 +26,23 @@ jobs:
           RELEASE_KEYSTORE: ${{ secrets.RELEASE_KEYSTORE }}
           RELEASE_KEYSTORE_PROPERTIES: ${{ secrets.RELEASE_KEYSTORE_PROPERTIES }}
         run: |
-          echo "${{ env.RELEASE_KEYSTORE }}" > .sign/release.keystore.asc
-          gpg -d --passphrase "${{ env.PASSPHRASE }}" --batch .sign/release.keystore.asc > .sign/release.keystore
-          echo "${{ env.RELEASE_KEYSTORE_PROPERTIES }}" > .sign/keystore.properties.asc
-          gpg -d --passphrase "${{ env.PASSPHRASE }}" --batch .sign/keystore.properties.asc > .sign/keystore.properties
-          echo "${{ env.ENV_PROPERTIES }}" > .env.properties
+          echo "$RELEASE_KEYSTORE" > .sign/release.keystore.asc
+          gpg -d --batch --pinentry-mode loopback --passphrase-fd 0 \
+            .sign/release.keystore.asc > .sign/release.keystore <<< "$PASSPHRASE"
+          echo "$RELEASE_KEYSTORE_PROPERTIES" > .sign/keystore.properties.asc
+          gpg -d --batch --pinentry-mode loopback --passphrase-fd 0 \
+            .sign/keystore.properties.asc > .sign/keystore.properties <<< "$PASSPHRASE"
+          echo "$ENV_PROPERTIES" > .env.properties
       - name: Assemble
         run: bash ./gradlew :demo-app:assembleDevelopmentRelease --stacktrace
+      - name: Remove signing material
+        if: always()
+        run: |
+          rm -f .sign/release.keystore .sign/release.keystore.asc \
+                .sign/keystore.properties .sign/keystore.properties.asc \
+                .env.properties
       - name: Upload APK
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         with:
           name: demo-app-development-release
           path: demo-app/build/outputs/apk/development/release/
--- a/.github/workflows/app-distribute-v2.yml
+++ b/.github/workflows/app-distribute-v2.yml
@@ -8,6 +8,10 @@ on:
   push:
     branches:
       - develop-v2
+      # TEMPORARY, remove before merging this PR. A push trigger reads the workflow file from
+      # the branch being pushed, so this is the only way to verify the pipeline before it
+      # reaches develop-v2. workflow_dispatch is not available until the file is on develop.
+      - andrerego/and-1497-distribute-v2-demo-app-builds-via-firebase-app-distribution
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/app-distribute-v2.yml
+++ b/.github/workflows/app-distribute-v2.yml
@@ -1,0 +1,48 @@
+name: App Distribute CI v2
+
+# Temporary workflow to publish the v2 demo app to Firebase App Distribution in the shared
+# stream-android-32afb project. When v2 is released and `develop-v2` becomes `develop`, remove
+# this workflow and update the main `app-distribute.yml` instead.
+
+on:
+  push:
+    branches:
+      - develop-v2
+  workflow_dispatch:
+
+jobs:
+  build_v2_demo_app:
+    permissions:
+      actions: write
+      contents: read
+    name: Build and Distribute V2 Demo App
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+      - uses: GetStream/android-ci-actions/actions/setup-java@cb46e4f44d105a68738bfd2a1cfd156bb768665b
+      - name: Prepare environment
+        env:
+          ENV_PROPERTIES: ${{ secrets.ENV_PROPERTIES }}
+          PASSPHRASE: ${{ secrets.PASSPHRASE }}
+          RELEASE_KEYSTORE: ${{ secrets.RELEASE_KEYSTORE }}
+          RELEASE_KEYSTORE_PROPERTIES: ${{ secrets.RELEASE_KEYSTORE_PROPERTIES }}
+        run: |
+          echo "${{ env.RELEASE_KEYSTORE }}" > .sign/release.keystore.asc
+          gpg -d --passphrase "${{ env.PASSPHRASE }}" --batch .sign/release.keystore.asc > .sign/release.keystore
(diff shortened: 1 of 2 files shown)


andremion:
@coderabbit review

.github/workflows/app-distribute-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  push:
+    branches:
+      - develop-v2
+  workflow_dispatch:
+
+jobs:
+  build_v2_demo_app:
+    permissions:
+      actions: write
```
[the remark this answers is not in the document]
andremion: Fixed in a12e2fd8bc. The permissions block is now `contents: read` only.

Verified rather than assumed: run 34132353521 built and uploaded a 72 MB artifact with only `contents: read`, so `upload-artifact` does not need `actions: write` here.

Worth noting the other four workflows in this repo that upload artifacts all still carry `actions: write`. That looks like the same over-grant, but removing it there is out of scope for this PR.

.github/workflows/app-distribute-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+        env:
+          ENV_PROPERTIES: ${{ secrets.ENV_PROPERTIES }}
+          PASSPHRASE: ${{ secrets.PASSPHRASE }}
+          RELEASE_KEYSTORE: ${{ secrets.RELEASE_KEYSTORE }}
+          RELEASE_KEYSTORE_PROPERTIES: ${{ secrets.RELEASE_KEYSTORE_PROPERTIES }}
+        run: |
+          echo "${{ env.RELEASE_KEYSTORE }}" > .sign/release.keystore.asc
+          gpg -d --passphrase "${{ env.PASSPHRASE }}" --batch .sign/release.keystore.asc > .sign/release.keystore
+          echo "${{ env.RELEASE_KEYSTORE_PROPERTIES }}" > .sign/keystore.properties.asc
+          gpg -d --passphrase "${{ env.PASSPHRASE }}" --batch .sign/keystore.properties.asc > .sign/keystore.properties
```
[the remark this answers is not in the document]
andremion: Fixed in a12e2fd8bc, using the stdin approach you suggested.

I originally planned to defer this one, because the decrypt commands were copied from the two existing distribution workflows and I had no way to test a signing change before merge. So I temporarily added this branch to the `push` trigger and ran the workflow. `Prepare environment` passed in run 34132353521, which confirms both files decrypt correctly with `--pinentry-mode loopback --passphrase-fd 0`. The temporary trigger has been reverted.

.github/workflows/app-distribute-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+          echo "${{ env.ENV_PROPERTIES }}" > .env.properties
```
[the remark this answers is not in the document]
andremion: Fixed in a12e2fd8bc. Added a `Remove signing material` step with `if: always()` directly after `Assemble`, deleting the decrypted keystore, the keystore properties, both .asc files and `.env.properties`.

Confirmed in run 34132353521 that the step runs before the Firebase Docker action, and the build still completed and distributed successfully.

.github/workflows/app-distribute-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Assemble
+        run: bash ./gradlew :demo-app:assembleDevelopmentRelease --stacktrace
+      - name: Upload APK
+        uses: actions/upload-artifact@v7
```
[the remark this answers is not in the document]
andremion: Fixed in a12e2fd8bc. Pinned to `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a`, which is the commit the `v7` tag currently points at. That matches how `checkout` and the Firebase action are already pinned in this repo.

.github/workflows/app-distribute-v2.yml
```diff
@@ -0,0 +1,55 @@
+name: App Distribute CI v2
+
+# Temporary workflow to publish the v2 demo app to Firebase App Distribution in the shared
+# stream-android-32afb project. When v2 is released and `develop-v2` becomes `develop`, remove
+# this workflow and update the main `app-distribute.yml` instead.
+
+on:
```
gpunto, line 7: Non-blocking: consider adding a `concurrency` group so overlapping pushes to `develop-v2` don't build and distribute in parallel — otherwise a slower, older run can push a stale build to Firebase after a newer one has already gone out. `internal-app-distribute.yml` already does this:

```yaml
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: true
```

.github/workflows/app-distribute-v2.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Assemble
+        run: bash ./gradlew :demo-app:assembleDevelopmentRelease --stacktrace
+      - name: Remove signing material
+        if: always()
+        run: |
+          rm -f .sign/release.keystore .sign/release.keystore.asc \
+                .sign/keystore.properties .sign/keystore.properties.asc \
+                .env.properties
+      - name: Upload APK
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
```
gpunto: nit (non-blocking): this pins `actions/upload-artifact@043fb46…` (a v4-era SHA), whereas the other four workflows in this repo use `@v7`. `checkout` and `setup-java` are SHA-pinned consistently everywhere. Worth aligning on one version to avoid drift.

.github/workflows/app-distribute.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           name: demo-app-release
-          path: demo-app/build/outputs/apk/demo-app/release/
+          path: demo-app/build/outputs/apk/development/release/
```
gpunto: Non-blocking, and pre-existing rather than introduced by this change — but since you're touching this file, the new v2 workflow improves on two things this one still does:

1. It passes secrets via `env:` + `$VAR` and reads the GPG passphrase from stdin (`--passphrase-fd 0 <<<`) instead of interpolating `${{ }}` into the run script (lines 26–30 here).
2. It builds `assembleDevelopmentRelease` rather than `assembleRelease`, which here compiles all three release flavors (development, e2etesting, production) even though only the development APK is distributed.

Optional to fold those back into this file, or leave for a separate cleanup.

gpunto (approved):

.github/workflows/app-distribute-v2.yml
andremion: Good catch, added.

One correction on the reference though. `internal-app-distribute.yml` does have a concurrency group, but its form is `cancel-in-progress: ${{ !contains(github.ref, 'main') }}`, so it cancels for every ref except `main`. That keeps a Play release from being interrupted mid-publish. This workflow only runs on `develop-v2` and manual dispatch, so that conditional has nothing to do here and plain `cancel-in-progress: true` is the right form, which is what you suggested.

Also worth flagging: `app-distribute.yml` has no concurrency group at all. The same race exists there, but adding one changes behavior on `main`, so I left it out of this PR.

One more thing while you are here: your approval landed on f3ebec41ed, and afb01a9d90 went up about five minutes later. It narrows `assembleRelease` to `assembleDevelopmentRelease` in `app-distribute.yml`, since the job only ever uploads the development APK and was discarding the other two flavors. One line, but it was not in what you reviewed.

gpunto: Sounds good, thanks!

PR #1818, opened 2026-09-09
Send the SDK name and version as X-Stream-Client on SFU requests

aleksandar-apostolov:
### Goal

Closes [AND-1509](https://linear.app/stream/issue/AND-1509/send-x-stream-client-with-the-sdk-version-on-every-sfu-twirp-request)

The SFU's Twirp calls carried no SDK identity — only the coordinator API sent
`X-Stream-Client`. Same change as the JS SDK's GetStream/stream-video-js#2425.

### Implementation

- `SfuHeadersInterceptor` adds `X-Stream-Client: stream-video-android-v<sdk-version>`, built by
  `HeadersUtil.buildSfuSdkTrackingHeader()`.
- `SendStatsRequest.sdk` is unchanged and keeps reporting `stream-android`.
- Registered on `http.newBuilder()` for the signaling Retrofit client, not on the shared SFU
  `OkHttpClient`: the SFU socket is built on that same client and `SocketFactory` already sets
  its own `X-Stream-Client`, so registering there would put two values on the upgrade request.
- The `-v` matches the other SDKs' coordinator token (`stream-video-swift-v…`,
  `stream-video-react-v…`) rather than our own coordinator prefix, which has no `v`. So the same
  app sends `stream-video-android-1.30.0` to the coordinator and `stream-video-android-v1.30.0`
  to the SFU. Deliberate — cross-SDK consistency was chosen over `@` in the format thread.

### 🎨 UI Changes

None.

### Testing

- `SfuHeadersInterceptorTest` pins the header value and that exactly one is sent. Red-proved by
  dropping the `v`, which is the near-miss value given our coordinator prefix doesn't carry one.
- `./gradlew :stream-video-android-core:testDebugUnitTest :stream-video-android-core:spotlessCheck :stream-video-android-core:apiCheck`
- Not yet observed on the wire against a live SFU.

### DIFF
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
@@ -86,6 +86,15 @@ class HeadersUtil {
         }.sanitize()
     }
 
+    /**
+     * Builds the client information header (X-Stream-Client) sent to the SFU, e.g.
+     * `stream-android@1.30.0`. The name matches the one reported as `SendStatsRequest.sdk`.
+     *
+     * @return Header value as a string.
+     */
+    internal fun buildSfuSdkTrackingHeader(): String =
+        "stream-android@${BuildConfig.STREAM_VIDEO_VERSION}"
+
     private fun buildAppVersionForHeader() = (StreamVideo.instanceOrNull() as? StreamVideoClient)?.let { streamVideoImpl ->
         "|app_version=" + (streamVideoImpl.appVersion ?: getAppVersionName())
     } ?: ""
--- a/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
+++ b/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
@@ -86,9 +86,9 @@ class HeadersUtil {
         }.sanitize()
     }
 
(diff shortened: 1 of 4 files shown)


stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/header/HeadersUtil.kt
```diff
@@ -86,6 +86,10 @@ class HeadersUtil {
         }.sanitize()
     }
 
+    /** Builds the X-Stream-Client header sent to the SFU, e.g. `stream-video-android-v1.30.0`. */
+    internal fun buildSfuSdkTrackingHeader(): String =
+        "stream-video-android-v${BuildConfig.STREAM_VIDEO_VERSION}"
```
PratimMallick, line 91: The user agent that we send to cordinator is in a different format than this? Is that fine?

PratimMallick: having the same header across both cordinator and sfu would make it easier for backend to write queries on it or identify other issues IMO

aleksandar-apostolov: Different how?

PratimMallick: discussed .. we will align them in a different PR

PratimMallick (approved):

PR #1821, opened 2026-09-10
Refresh the lobby and Start Call screens with the design system tokens

andremion:
### Goal
Closes [AND-1513](https://linear.app/stream/issue/AND-1513/refresh-the-lobby-and-start-call-screens-with-the-design-system-tokens)

Apply the design tokens and the base components to the lobby and the demo Start Call screen while keeping the current layout and functionality. This is the scoped version of #1816, following the agreement that the design refresh is not a redesign: tokens, base components and like-for-like icon swaps, no new layout decisions.

### Implementation
- `StreamTokens` becomes public with explicit types on its members. The demo reads it, and the core-ui extraction (AND-1482) makes it public across a module boundary anyway.
- `CallLobby` uses the 80dp size token for the camera-off avatar and the large radius token for the preview clip. The controls keep their start alignment and the responsive tile heights stay, no token matches them.
- Demo lobby and Start Call screens move their dimensions, colors and typography to the tokens, replace the material close button with `StreamIconButton`, and use the design system icons where an equivalent exists. The lobby header takes the same padding, avatar size and spacing as the Start Call header. Strings, navigation and test tags are unchanged.
- The `qr-code-fill` icon joins the SDK icon set. It was added to design-system-tokens after the initial import and replaces the material QR scanner icon.
- Demo app draws edge to edge and encodes the call id in the lobby route.
- `VideoPermissionsState` gains `isCameraPermissionDenied` and `isMicrophonePermissionDenied` (default false). Tapping a lobby toggle whose permission was denied requests it again instead of leaving the system settings as the only way back. The error badge from #1816 is not included.
- Smaller ports from #1816: the Start Call description wraps naturally, KDoc on `buildDefaultLobbyControlActions`, and the non-Stream built-in demo user is replaced.

Left out on purpose: the participant label styling, indicator sizes and large-avatar initials style are shared components and belong to the in-call tickets.

### 🎨 UI Changes
Lobby, camera off, before and after:

| Before | After |
|---|---|
| 648484710-ff10431c-2535-48e3-91a6-e982ab0b39e1 | Screenshot_20260910_095550 |
| 648484688-8d2f5fa7-18d2-4f26-ab49-46a166848976 | Screenshot_20260910_095555 |

### Testing
Manual: open the demo Start Call screen and the lobby in light and dark mode. The two headers line up: same top margin, avatar size and gap to the name, close and settings buttons flush with the 16dp margin. Deny the camera permission, then tap the camera toggle in the lobby: the system permission dialog shows again.

Ran locally: `spotlessCheck`, `compileDebugKotlin`, `:demo-app:compileProductionDebugKotlin`, `apiDump` followed by `apiCheck`, `recordPaparazziDebug` and then `verifyPaparazziDebug`. The only golden change is the camera-off `CallLobbyTest` snapshot, from the smaller avatar. The four `LobbyControlsCallActionHandlerTest` cases pass.


andremion:
@coderabbit review

demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyScreen.kt
```diff
@@ -316,7 +302,7 @@ private fun CallLobbyBodyPortrait(
             call = call,
             modifier = Modifier
                 .fillMaxWidth()
-                .padding(16.dp),
+                .padding(StreamTokens.spacingMd),
```
gpunto: The re-request never runs in the demo app, so I do not think the manual test in the description can be coming from this path.

This `CallLobby` call overrides `lobbyControlsContent` just below (line 309) and builds the actions with the screen's own `onCallAction`. That skips `DefaultLobbyControlsSlot`, and with it `lobbyControlsCallActionHandler`, so the action goes straight to the ViewModel and on to `call.camera.setEnabled(...)` with no permission request. `CallLobby`'s internally remembered `permissions` is not reachable from this lambda either, and the only other launcher in the lobby is `DefaultPermissionHandler`, which fires once on entry.

The override predates this PR, so nothing here regressed — it just means the feature ships untested end to end. Hoisting `rememberCallPermissionsState(call)` into the screen and passing it to both `CallLobby(permissions = ...)` and a wrapper around `onCallAction` would work, as would dropping the override and letting the default slot handle it.

demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyScreen.kt
```diff
@@ -381,15 +367,15 @@ private fun CallLobbyBodyLandscape(
...
-                            start = 16.dp,
-                            end = 16.dp,
+                            start = StreamTokens.spacingMd,
+                            end = StreamTokens.spacingMd,
                         ),
                     isCameraEnabled = isCameraEnabled,
                     isMicrophoneEnabled = isMicrophoneEnabled,
                     onCallAction = onCallAction,
-                    lobbyControlsContent = { _, _ ->
+                    lobbyControlsContent = { modifier, _ ->
```
gpunto, line 378: This override (and the portrait one at 309) builds the actions with the screen's own `onCallAction`, so the toggle goes straight to the ViewModel and `call.camera.setEnabled(...)`, skipping `DefaultLobbyControlsSlot` and with it `lobbyControlsCallActionHandler` — the demo lobby never runs the new re-request. Both overrides predate this PR, but it does mean the feature ships without an end-to-end path. Hoisting `rememberCallPermissionsState(call)` into the screen and passing it to both `CallLobby` and a wrapper would cover it.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/lobby/CallLobby.kt
```diff
@@ -287,11 +305,27 @@ private fun DefaultLobbyControlsSlot(
...
 
+/**
+ * Wraps [onCallAction] so that toggling a device whose permission was denied also asks for the
+ * permission again. The toggle itself is still forwarded, so the device turns on once granted.
+ */
+internal fun lobbyControlsCallActionHandler(
+    permissions: VideoPermissionsState,
+    onCallAction: (CallAction) -> Unit,
+): (CallAction) -> Unit = { action ->
+    val needsPermission = (action is ToggleCamera && permissions.isCameraPermissionDenied) ||
```
gpunto: Keys on the action type but not `action.isEnabled`, so switching a device off prompts too. `CameraManager.enable()` sets `_status = Enabled` even without the permission (MediaManager.kt:1087), so the first tap leaves the toggle reading as on while capture never starts; the second emits `ToggleCamera(false)` (ToggleCameraAction.kt:63) and prompts the user while they're turning it off. Gating on `action.isEnabled` fixes it — case 3 of the new suite only covers the off direction with the permission granted, so it still passes.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/permission/CallPermissions.kt
```diff
@@ -50,7 +54,15 @@ public fun rememberCallPermissionsState(
...
     val isMicrophoneEnabled by call.microphone.isEnabled.collectAsStateWithLifecycle()
 
+    // A permission counts as denied only after the user answered a request, so a fresh screen does not
+    // flag the controls before the system dialog had a chance to show. The live status is read on top,
+    // so a permission granted later in the system settings clears the flag.
+    var cameraRequestAnswered by remember(call) { mutableStateOf(false) }
+    var microphoneRequestAnswered by remember(call) { mutableStateOf(false) }
+
     val permissionState = rememberMultiplePermissionsState(permissions) {
+        if (android.Manifest.permission.CAMERA in it) cameraRequestAnswered = true
```
gpunto: Non-blocking: `RequestMultiplePermissions.parseResult` returns `emptyMap()` when `resultCode != RESULT_OK` or `intent == null` — e.g. the dialog dismissed rather than answered. Neither key is then present, both flags stay false, and the toggle never re-requests, which is the state this is meant to recover from. Since the state requests all its permissions together, setting both flags whenever the callback fires closes the gap and still can't fire on a fresh screen.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/lobby/CallLobby.kt
```diff
@@ -287,11 +305,27 @@ private fun DefaultLobbyControlsSlot(
             isCameraEnabled = isCameraEnabled,
             isMicrophoneEnabled = isMicrophoneEnabled,
             modifier = modifier,
-            onCallAction = onCallAction,
+            onCallAction = lobbyControlsCallActionHandler(permissions, onCallAction),
```
gpunto: Non-blocking nit: not remembered, so a fresh lambda each composition makes `CallLobbyControlsContentParams` unequal and the controls subtree recomposes on every camera/mic state change. `remember(permissions, onCallAction)` helps, though callers often pass an unmemoized `onCallAction` anyway.

demo-app/src/main/kotlin/io/getstream/video/android/MainActivity.kt
```diff
@@ -60,6 +61,7 @@ class MainActivity : ComponentActivity() {
     private val firebaseAnalytics by lazy { FirebaseAnalytics.getInstance(this) }
 
     override fun onCreate(savedInstanceState: Bundle?) {
+        enableEdgeToEdge()
```
gpunto, line 64: Non-blocking: this sets `decorFitsSystemWindows = false`, and once the decor no longer fits system windows the soft input mode stops doing the work — the manifest still relies on `adjustPan` (AndroidManifest.xml:47) and nothing applies `imePadding()`. targetSdk 35 means only Android 14 and below change here; could you check the call ID field still clears the keyboard on a pre-15 device?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/design/StreamTokens.kt
```diff
@@ -26,132 +28,132 @@ import androidx.compose.ui.unit.sp
  * Semantic classes build on these; components use them directly.
  */
 @Suppress("MagicNumber")
-internal object StreamTokens {
+public object StreamTokens {
```
gpunto, line 31: Should we make this at least @InternalStreamVideoApi?

demo-app/src/main/kotlin/io/getstream/video/android/ui/lobby/CallLobbyScreen.kt
andremion: Good catch, the demo never hit the new path. The screen now creates `rememberCallPermissionsState(call)` once, passes it to `CallLobby`, and routes its own toggles through the same re-request before calling the view model.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/lobby/CallLobby.kt
andremion: Right, the off tap was prompting. The handler now only asks when the action turns the device on, and I added a test for the off direction with the permission denied.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/permission/CallPermissions.kt
andremion: Missed that `parseResult` returns an empty map on dismiss. Replaced the two flags with one that is set on every result, since the state requests both permissions together anyway.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/lobby/CallLobby.kt
andremion: Done, remembered on `permissions` and `onCallAction`.

demo-app/src/main/kotlin/io/getstream/video/android/MainActivity.kt
andremion: Checked on a Pixel 4 API 30 emulator and on API 35 with this branch: the content still pans up and the call ID field stays above the keyboard in both. Leaving it as is, happy to add `imePadding()` if you see it differently on a device.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/design/StreamTokens.kt
andremion: There is no v2 release planned before the design refresh is done, and AND-1482 makes these tokens public in the shared module anyway, so the annotation would only be removed again shortly after.

andremion:
@coderabbit full review

gpunto (approved):

PR #1822, opened 2026-09-10
Pin the video buddy to 1.8.1 and widen the flaky E2E waits

andremion:
### Goal

Refs [AND-1445](https://linear.app/stream/issue/AND-1445/stabilize-the-video-e2e-emulator-tests)

The nightly E2E has been red most nights. Two separate causes.

The first is in stream-video-buddy. Its lobby page poll calls `page.locator().count()` in a loop, and when the page navigates mid-poll Playwright rejects the call. Nothing catches it, so the buddy process dies and no participant ever joins. The Android test waiting for those participants then fails far from the real cause. Every uncaught exception in the nightly buddy logs from Sep 1 to Sep 8 was this one line, 8 of them. GetStream/stream-video-buddy#13 fixes it and 1.8.1 is the first release that contains it.

The second is our own waits. Counting every timeout in the same nightly artifacts, and leaving out the shard where an emulator launcher ANR covered the screen, the join call button on the call details screen timed out 9 times at the 5s default and the spotlight menu item timed out twice at 15s.

### Implementation

- `fastlane/Fastfile`: `stream_video_buddy_version` goes from `1.6.32` to `1.8.1`.
- `UserRobot.enterLobby`: the join call button wait goes from the 5s default to 15s. It is the first screen after launch or login, so it overlaps app startup.
- `UserRobot.waitForLobbyToOpen`: the lobby close button wait goes from the 5s default to 15s. Same reason, and it failed about half the time locally on the re-enter tests.
- `UserRobot.setView`: the view menu items wait goes from 15s to 30s. 15s still timed out on the six-participant test.

### Testing

- Nightly dispatched on this branch with the buddy pinned to the merged commit of #13 and the artifact upload temporarily set to `always()` so green shards also uploaded `video-buddy-server.log`: [run 34467476054](https://github.com/GetStream/stream-video-android/actions/runs/34467476054). All 6 shards green, 40 tests each. `Execution context was destroyed` appears 0 times across the six buddy logs, against 8 occurrences in the Sep 1 to Sep 8 nightlies. The only retries were `testReconnectionDuringCallRecording` on API 31 and 32, both on the recording consent dialog, which is a pre-existing backend recorder latency flake (same message in the Sep 2 and Aug 28 runs) and passed on the second attempt.
- The `1.8.1` tag points at the same buddy code as that commit plus the version bump.
- Locally, full E2E suite on a fresh API 35 emulator with the widened waits: 39 of 40 on the first attempt, the one failure being the lobby close button at the 5s default, which is what prompted the third widening.
- `bundle exec rubocop fastlane/Fastfile` clean, `:demo-app:assembleE2etestingDebugAndroidTest` compiles.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of call lobby and view-selection flows, particularly on slower devices or emulators.

* **Chores**
  * Updated supporting video development tooling to a newer version.

### DIFF
--- a/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
+++ b/demo-app/src/androidTestE2etestingDebug/kotlin/io/getstream/video/android/robots/UserRobot.kt
@@ -108,7 +108,10 @@ class UserRobot {
         if (callId != null) {
             CallDetailsPage.callIdInputField.waitToAppear().typeText(callId)
         }
-        CallDetailsPage.joinCallButton.waitToAppearAndClick()
+        // The call details screen is the first screen after launch or login, so this wait
+        // overlaps app startup on an emulator that is still settling. The 5s default was the
+        // single most common nightly failure, so the window is deliberately wider.
+        CallDetailsPage.joinCallButton.waitToAppearAndClick(timeOutMillis = 15.seconds)
         waitForLobbyToOpen()
         return this
     }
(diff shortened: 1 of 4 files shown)


andremion:
@coderabbit review

aleksandar-apostolov (approved):

PR #1827, opened 2026-09-11
Refresh the video calling screens with the design system tokens

andremion:
### Goal

Closes [AND-1425](https://linear.app/stream/issue/AND-1425/refresh-the-video-calling-screen-with-the-design-system-tokens)

The in-call screens move to the design system tokens and the base components from AND-1420. This is a refresh, not a redesign, so the layout and the functionality stay as they are on `develop-v2`. The scope rules are on the parent issue [AND-1421](https://linear.app/stream/issue/AND-1421/refresh-the-screens-with-the-design-system-tokens).

### Implementation

SDK:

* `CallAppBar` uses `StreamIconButton` for the back button, a stroke token for the recording dot and a typography token for the center text.
* Participant tiles, labels, indicators and reactions use spacing, radius and emoji tokens. `FloatingParticipantVideo` uses the size and radius tokens and keeps its drag behavior.
* Large avatar initials scale with the avatar instead of staying at one small size, capped at the size the 1.x SDK used.
* The call action buttons default to the 48dp `Large` size. The base component rebuild had left them at 40dp, the 1.x SDK drew 44dp and Figma asks for 48dp.
* The hifi audio toggle goes back to the music note and music off icons. The design icon set has no music glyph and the sliders icon reads as a settings control.
* The ringing screens and the audio call screen paint their background with the app color instead of the surface color. The design system maps the secondary button background to that same surface color, so the mic, camera and speaker buttons had no visible shape.

Demo app:

* The call screen and the share panel use tokens and the design icons. The share panel keeps the surface color because it is a card, so its Call ID button uses the outlined secondary style.
* The call activity draws edge to edge, as `MainActivity` does since AND-1513.
* The settings button received the negated menu state, so it looked pressed while the menu was closed.

### 🎨 UI Changes

1.x on the left, this branch on the right.

**Lobby**

lobby

**In call**

call

**In call, dark mode**

call_dark

**Camera off, participant initials**

call_initials

**Camera off, participant initials, dark mode**

call_initials_dark

**Share panel**

share

**Share panel, dark mode**

share_dark

### Testing

Ran on two emulators, a Pixel 4 on API 30 with the 1.x demo app and a Pixel 8 on API 35 with this branch, in a shared call. Checked the lobby, the call grid, the camera-off tiles, the muted state, the share panel and the ringing screens, in light and dark mode.

To reproduce: start a test call on one device, join it from a second one, then open the share panel and toggle the camera and the microphone on both.

Validation on the final tree:

```
./gradlew spotlessCheck apiCheck testDebugUnitTest
./gradlew :stream-video-android-ui-compose:recordPaparazziDebug
./gradlew :stream-video-android-ui-compose:verifyPaparazziDebug
```

All green. The snapshots were re-recorded once, deliberately, as the parent issue allows. There is no `detekt` task in this repo.

## Summary by CodeRabbit

- **New Features**
  - Added music and music-off icons for HiFi audio controls.
  - Large avatars now scale initials text appropriately for improved readability.

- **UI Improvements**
  - Call controls now use larger default button sizing.
  - Updated call screens, invite dialogs, sharing layouts, and participant actions with consistent spacing, colors, typography, icons, and button styles.
  - Improved call backgrounds, recording indicators, video cards, and spotlight corner styling.
  - Enabled edge-to-edge display in the demo call experience.
  - Updated the default back button and call action visuals for consistency.


andremion:
@coderabbit review

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/avatar/InitialsAvatar.kt
```diff
@@ -78,6 +81,18 @@ internal fun Dp.toAvatarTextStyle(): TextStyle {
         this < StreamTokens.size32 -> typography.captionEmphasis
         this < StreamTokens.size48 -> typography.bodyEmphasis
         this < StreamTokens.size80 -> typography.headingLarge
-        else -> typography.numericExtraLarge
+        else -> {
+            val scaled = with(LocalDensity.current) {
+                (this@toAvatarTextStyle * LARGE_AVATAR_TEXT_RATIO).toSp()
+            }
+            val fontSize = if (scaled > LARGE_AVATAR_MAX_TEXT_SIZE) LARGE_AVATAR_MAX_TEXT_SIZE else scaled
```
gpunto: Non-blocking, but the ceiling here isn't absolute.

`Dp.toSp()` is `(value / fontScale).sp`, so `scaled` has already had the user's font scale divided out. Comparing that against a fixed `48.sp` makes the *effective* ceiling move with the font-scale setting instead of pinning it at 48dp:

- font scale 1.0 → `scaled` = 80sp → capped → initials render at 48dp
- font scale 2.0 → `scaled` = 40sp → under the cap → initials render at ~80dp

On the 200dp avatar in `ParticipantAvatars.kt` that's a visibly different result from what `LARGE_AVATAR_MAX_TEXT_SIZE`'s KDoc promises. Clamping in `Dp` before the conversion keeps it absolute:

```kotlin
val capped = minOf(this@toAvatarTextStyle * LARGE_AVATAR_TEXT_RATIO, LARGE_AVATAR_MAX_TEXT_SIZE)
val fontSize = with(LocalDensity.current) { capped.toSp() }
typography.headingLarge.copy(fontSize = fontSize, lineHeight = fontSize)
```

with `LARGE_AVATAR_MAX_TEXT_SIZE` becoming `48.dp`. At font scale 1 the output is identical, so the re-recorded snapshots stay valid — I verified `verifyPaparazziDebug` passes either way.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/avatar/InitialsAvatar.kt
```diff
@@ -78,6 +81,18 @@ internal fun Dp.toAvatarTextStyle(): TextStyle {
         this < StreamTokens.size32 -> typography.captionEmphasis
         this < StreamTokens.size48 -> typography.bodyEmphasis
         this < StreamTokens.size80 -> typography.headingLarge
```
gpunto: Not for this PR — a parity note, since Chat and Video are now on the same design tokens.

Chat's equivalent ladder (`AvatarPlaceholder.kt`, `Dp.toPlaceholderTextStyle()`) picks a different style in two of the bands both SDKs share:

| avatar size | Chat | here |
|---|---|---|
| 32–39dp | `captionEmphasis` | `bodyEmphasis` |
| 48–79dp | `headingMedium` | `headingLarge` |

Chat also keys off named `AvatarSize` constants rather than raw size tokens, which gives consumers a size vocabulary to match against.

Related: the reason the `else` branch has to compute anything is that `numericExtraLarge` is `fontSizeSm` (14sp) despite its name — and it's the same in Chat. If that token were fixed in both SDKs this could go back to being a plain lookup, which is the pattern Chat follows throughout.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/AudioCallContent.kt
```diff
@@ -194,7 +194,7 @@ public fun AudioOnlyCallContent(
     Box(
         modifier = modifier
             .fillMaxSize()
-            .background(color = VideoTheme.colors.backgroundCoreSurfaceDefault),
+            .background(color = VideoTheme.colors.backgroundCoreApp),
     ) {
         Column(
             modifier = modifier
```
gpunto: Pre-existing, not introduced here — flagging it only because it's adjacent to the line you changed.

The inner `Column` re-applies `modifier`, which the outer `Box` two lines up already has. A caller passing `AudioOnlyCallContent(modifier = Modifier.padding(24.dp))` gets the padding twice, and a caller-supplied `.background(...)` would repaint over the `backgroundCoreApp` this PR just introduced. Should be `Modifier`. Same pattern at lines 113/119.

gpunto (approved):

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/avatar/InitialsAvatar.kt
andremion: Good catch, you're right. Fixed in 072ab8e2a5f3afa21fff6cf0e21c5f7c66ba94df, clamping in Dp before the conversion like you suggested. Goldens verify without re-recording, so font scale 1 is unchanged.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/avatar/InitialsAvatar.kt
andremion: Took the band alignment in 40af253e62866d727b769754d1934b77fe72f632, video follows chat now below 80dp. Four goldens move and it's 20sp to 18sp on the initials, under one percent of the pixels.

One thing on numericExtraLarge though: I checked the design source and numeric.xl maps to font-size.sm there, so the 14sp is intentional. That scale is micro, xxs, xs, sm for badge counts, so extra large means the biggest badge rather than a big heading. Both SDKs read it correctly. The else branch still has to compute because our avatars go up to 200dp and chat tops out at 80dp.

I'll flag the ladder with Jurgen so he can correct both SDKs if chat is the one that's off.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/activecall/AudioCallContent.kt
andremion: Fixed in 7aa1e86176d594cb1ea2f8e7d544b04bd9b1e30d, both sites. Pre-existing as you say, but the background change makes the repaint case real, so it seemed worth taking here.

PR #1828, opened 2026-09-11
Merge develop into develop-v2

andremion:
### Goal

Closes [AND-1524](https://linear.app/stream/issue/AND-1524/merge-develop-into-develop-v2-blocked-on-the-repackaged-webrtc)

Bring develop-v2 up to date with develop. It is 11 commits behind: the SFU pin, the
X-Stream-Client header, the Bluetooth permission fix for audio-only calls, the FCM
logout fix, the extra demo users, E2EE for call media, the outgoing ring timeout fix
and the E2E test updates.

Draft on purpose. It does not compile yet, see Testing.

### Implementation

- Recorded the develop history that #1790 already brought over, then merged
  `origin/develop` on top. #1790 was squash merged, so git no longer knows those
  commits are contained in develop-v2, and a plain merge falls back to the old merge
  base and re-resolves the whole v1/v2 divergence (#1790 went through about 116
  conflicts that way). With the recording commit, only 4 files conflicted.
- `gradle/libs.versions.toml`: kept WebRTC 137.1.1 and noise cancellation 2.0.0, as
  #1790 did. develop moved to WebRTC 145.17.0, which is published only under the
  `org.webrtc` package.
- `CoordinatorConnectionModule`: kept the v2 lazy OkHttp client, which the Compose
  previews need, and the `Unit` socket connection type, and added develop's
  `pinnedSfuId` interceptor inside the lazy builder.
- `CallPermissions`: took develop's BLUETOOTH_CONNECT request on Android 12 and above.
- Demo `CallLobbyScreen`: kept the v2 `StreamIconButton` close button and added
  develop's E2EE button next to it.

### Testing

Blocked. `:stream-video-android-core:compileDebugKotlin` fails with 102 errors, and
all of them come from one missing class.

The E2EE code from #1801 compiles against `org.webrtc.EncryptionManager`, which ships
in `io.getstream:stream-video-webrtc-android:145.17.0`. develop-v2 depends on
`io.getstream:stream-video-webrtc-android-repackaged` (the `io.getstream.webrtc`
namespace), and Maven Central publishes only 137.1.1 of that artifact. Both AARs were
unpacked to confirm: 145.17.0 has `EncryptionManager`, the repackaged 137.1.1 has
`FrameCryptor` and no `EncryptionManager`. The two errors in `Publisher.kt` and
`Subscriber.kt` are `toE2EETrackType()` overload ambiguity caused by the same missing
class, not separate problems.

Nothing else in the merge breaks. Once
`stream-video-webrtc-android-repackaged` is published at 145.17.0 or newer, the
remaining work is a compile of core and compose, `apiDump`, `spotlessApply`, and then
`./gradlew spotlessCheck detekt apiCheck testDebugUnitTest`.

Not run yet, because they need a compiling module: `spotlessCheck`, `detekt`,
`apiCheck`, `testDebugUnitTest`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end encryption for calls, including passphrase setup, encrypted invite links, and automatic encryption when joining from a link.
  * Added encryption status controls and messaging in the call lobby.
  * Added support for Bluetooth audio permissions on Android 12 and later.
  * Added optional SFU selection for local development and testing.

* **Bug Fixes**
  * Improved join-and-ring behavior during reconnects and rejoins.
  * Improved call stability by allowing more time for slower startup and connection flows.

* **Documentation**
  * Updated reported SDK size values in the README.


andremion:
@coderabbit review

PR #1830, opened 2026-09-11
Dispose the demo E2EE manager on leave

PratimMallick:
### Goal
Fixes [AND-1525](https://linear.app/stream/issue/AND-1525/dispose-the-demo-e2ee-manager-on-leave-so-perf-report-stops) — stop leftover native `PERF_REPORT` logs after hangup.

Lobby created the manager, then `CLEAR_TASK` killed the ViewModel. `Call.leave()` only detaches it, so the 1 Hz timer kept firing with empty encode/decode arrays.

### Implementation
- `DemoE2eeKeys` also holds the manager so `CallActivity` can adopt it after the lobby dies.
- `CallActivity` disposes that instance on `leave` / `finish`.
- Lobby back (`leaveCall`) disposes when the call was never joined.
- `StreamEncryptionManager.dispose()` nulls the event listener before native teardown.

### Testing
- Manual: enable E2EE in the lobby, join, leave. `CallLobbyViewModel` `PERF_REPORT` lines should stop at hangup, not keep ticking with empty arrays.
- Pre-push hook: `./gradlew spotlessCheck apiCheck` passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encrypted call handling when transitioning from a lobby or leaving a call.
  * Preserved active encryption sessions when a lobby is unexpectedly closed.
  * Ensured encryption resources and keys are released after leaving or finishing a call.
  * Prevented disposed encryption sessions from retaining application screens or view models.
  * Added support for passing the call leave reason through the call-leaving flow.

### DIFF
--- a/demo-app/src/main/kotlin/io/getstream/video/android/CallActivity.kt
+++ b/demo-app/src/main/kotlin/io/getstream/video/android/CallActivity.kt
@@ -35,6 +35,7 @@ import io.getstream.result.onSuccessSuspend
 import io.getstream.video.android.compose.ui.ComposeStreamCallActivity
 import io.getstream.video.android.compose.ui.StreamCallActivityComposeDelegate
 import io.getstream.video.android.core.Call
+import io.getstream.video.android.core.CallLeaveReason
 import io.getstream.video.android.core.MemberState
 import io.getstream.video.android.core.RingingState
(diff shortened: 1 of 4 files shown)


PratimMallick:
@coderabbit review

rahul-lohra (approved):
LGTM

PR #1831, opened 2026-09-14
Refresh the demo app call surfaces with the design system

andremion:
### Goal

Closes [AND-1428](https://linear.app/stream/issue/AND-1428/refresh-the-demo-app-call-surfaces-with-the-design-system)

Apply the design system icons, tokens and base components to the demo app surfaces that open
from the call screen: the call menus, the participants dialog, the call stats dialog, the chat
sheet and the landscape controls. Layout and behaviour stay as they are on `develop-v2`.

### Implementation

- Menu icons now carry a `MenuIcon`, which is either a design system drawable or a material
  icon. Seven entries have no design system equivalent and keep the material icon they have
  today: the three audio device types, the scale type and video quality entries, noise
  cancellation, and the bad network label. Everything else maps one to one.
- `ParticipantsDialog`, `CallStatsDialog` and the chat sheet were hardcoded dark and now follow
  the theme. Both were broken in light mode: the stats body text used `textPrimary` on a black
  background, and the settings menu close icon was hardcoded white on a white surface, so
  neither was readable.
- The settings menu, landscape controls, layout chooser and audio device menu painted
  themselves with `backgroundCoreApp`, the same token the call screen uses, so none had a
  visible edge. They now use `backgroundCoreSurfaceDefault`. Elevation tokens do not work here:
  in the light theme `backgroundCoreElevation0` to `3` are all `chrome.s0`, the same value as
  `backgroundCoreApp`.
- `button.secondary.bg` maps to that same surface token, so once the panels became visible the
  `secondarySolid` controls inside them lost their shape. The reaction buttons, the unselected
  background filter toggles and the landscape Settings button move to `secondaryOutline`.
- The background filter toggles built their test tag from the icon name, which drawables do not
  have, so they take a label instead. `CallPage` is updated to match.
- Remaining literals move to `StreamTokens` and `VideoTheme` typography. Values with no token
  stay literal: the popup offsets, the layout chooser and landscape panel widths, the chat sheet
  height and the stats chart padding.

### 🎨 UI Changes

Each grid is BEFORE | AFTER by column, Light / Dark by row.

**Settings menu**

settings_menu

**Debug submenu**

debug_submenu

**Call stats**

call_stats

**Participants**

participants

**Menu icons, zoomed (light theme)**

icon_detail

Worth noting when reading these: the BEFORE column is identical in the Light and Dark rows for
the stats and participants screens. That is the screen ignoring the theme. Those screens looked
correct in dark and were unreadable in light.

### Testing

The demo app has no Paparazzi goldens, so this was verified on device.

Open a call in the demo app, then check each surface in both light and dark:
- the settings menu and its debug submenu, including the back button
- Call stats, from the settings menu
- the participants dialog, from the people button in the control bar
- the chat sheet, and the landscape controls in landscape

`./gradlew spotlessCheck apiCheck testDebugUnitTest` passes, plus
`:demo-app:compileDevelopmentDebugKotlin` and `:demo-app:compileE2etestingDebugAndroidTestKotlin`.
There is no `detekt` task in this repo. `apiCheck` shows no drift; the demo app has no API dump.

## Summary by CodeRabbit

- **Style**
  - Updated call screens, dialogs, menus, reactions, and device controls to use consistent design-system colors, spacing, typography, shapes, and sizing.
  - Refreshed icons throughout the app with Stream-designed resources for improved visual consistency.
  - Updated reaction and virtual-background controls with clearer outlined styling and refreshed visuals.
  - Improved dark-mode behavior for the chat dialog and standardized menu backgrounds and loading states.

- **Refactor**
  - Consolidated menu icon handling to support consistent resource-based and vector icons across menus.


andremion:
@coderabbit review

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/MenuDefinitions.kt
```diff
@@ -297,9 +277,9 @@ fun recordingTypeMenu(onSelectRecording: (RecordingType) -> Unit, selectedRecord
                 "Start raw recording"
             },
             icon = if (selectedRecordingTypes.contains(RecordingType.Raw)) {
-                Icons.Default.RawOn
+                menuIcon(ComposeR.drawable.stream_design_ic_recording_fill)
             } else {
-                Icons.Default.RawOff
+                menuIcon(ComposeR.drawable.stream_design_ic_recording_stop_fill)
```
gpunto: Non-blocking: these read backwards against the labels. Off gives "Start raw recording" + a stop square, on gives "Stop raw recording" + a record dot.

The other two rows use the icon slot for the recording type (user, fullscreen) rather than for state, and `highlight` already carries state. `recording_fill` for both, matching the parent submenu, would be consistent.

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/MenuDefinitions.kt
```diff
@@ -189,23 +169,23 @@ fun getCCActionMenu(
...
+                icon = menuIcon(ComposeR.drawable.stream_design_ic_caption_fill),
                 action = onToggleClosedCaptions,
             )
         }
 
         is ClosedCaptionUiState.UnAvailable -> {
             ActionMenuItem(
                 title = "Closed Caption are unavailable",
-                icon = Icons.Default.ClosedCaptionDisabled,
+                icon = menuIcon(ComposeR.drawable.stream_design_ic_caption_fill),
```
gpunto: Non-blocking: all three states share `caption_fill` and none sets `highlight`, so only the title tells them apart — and the unavailable row is a no-op `action = { }` that still looks tappable.

`highlight = true` on Running, or `stream_design_ic_no_sign` for UnAvailable, would separate them.

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallStats.kt
```diff
@@ -348,47 +328,35 @@ fun UserAndCallId(call: Call, clipboardManager: ClipboardManager?) {
...
+                        color = VideoTheme.colors.textSecondary,
                     )
                 }
             }
             clipboardManager?.let {
-                Spacer(modifier = Modifier.size(8.dp))
-                IconButton(
-                    modifier = Modifier.size(32.dp),
+                Spacer(modifier = Modifier.size(StreamTokens.spacingXs))
+                StreamIconButton(
```
gpunto: Non-blocking, pre-existing: the outer `Row` (L316) is `SpaceBetween` with no `weight`, so the left `Row` is measured against the full width first and this button gets the remainder — which a UUID-style `cid` can leave at zero. The 44→48dp avatar and the default-`Medium` button shrink that budget further.

`Modifier.weight(1f)` on the left `Row` fixes it. Same shape in `ParticipantsDialog.kt:130`.

gpunto, line 352: Non-blocking, pre-existing: the outer `Row` (L316) is `SpaceBetween` with no `weight`, so the left `Row` is measured against the full width first and this button gets the remainder — which a UUID-style `cid` can leave at zero. The 44→48dp avatar and the default-`Medium` button shrink that budget further.

`Modifier.weight(1f)` on the left `Row` fixes it. Same shape in `ParticipantsDialog.kt:130`.

gpunto (approved):

demo-app/src/main/kotlin/io/getstream/video/android/ui/menu/MenuDefinitions.kt
andremion: You are right, these were three different material icons before and I collapsed them into one. Did both: `highlight = true` on Running so it matches Stop Transcription, and `no_sign` for the unavailable row.

c1bce1f93644c23c2deaf22020a3ba5937faba14

demo-app/src/main/kotlin/io/getstream/video/android/ui/call/CallStats.kt
andremion: Confirmed, the copy button was not rendering at all with a UUID cid. Added `Modifier.weight(1f)` here and in `ParticipantsDialog`, plus ellipsis on the participant name since `maxLines = 1` now clips instead of pushing the icons out.

c1bce1f93644c23c2deaf22020a3ba5937faba14

PR #1833, opened 2026-09-14
Add a ToggleScreenShare call action

andremion:
### Goal

Fixes [AND-1519](https://linear.app/stream/issue/AND-1519/add-a-togglescreenshare-call-action)

`ScreenShareToggleAction` emitted `ClosedCaptionsAction(!active)` because `stream-video-android-core` had no screen share `CallAction`. Any app that handles the toggle through a shared `(CallAction) -> Unit` handler would have toggled captions instead of screen sharing.

### Implementation

- Added `ToggleScreenShare(isEnabled: Boolean)` to `CallAction` in `stream-video-android-core`.
- `ScreenShareToggleAction` now takes `(ToggleScreenShare) -> Unit` and emits `ToggleScreenShare(!active)`. This changes the public callback type in `stream-video-android-ui-compose`.
- `DefaultOnCallActionHandler` leaves the action unhandled on purpose. Starting a share needs the app's MediaProjection consent intent, so the app has to handle it.
- The demo `CallScreen` caller now branches on `action.isEnabled` instead of the captured `isScreenSharing` state.
- Only the core API dump changes. The compose signature erases the lambda to `Function1`, so its dump is unchanged.

### Testing

Manual: on a tablet (the demo only shows the toggle on tablets), join a call and tap the screen share button. The MediaProjection consent dialog opens. Accept it, tap the button again, and sharing stops.

Ran `./gradlew spotlessCheck apiCheck testDebugUnitTest` plus `compileDebugKotlin` for core and ui-compose, `:demo-app:compileDevelopmentDebugKotlin` and the `ControlActionsTest` Paparazzi test in ui-compose. All green, no golden changes.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated screen-sharing action that reports whether screen sharing should be enabled or disabled.
  - Screen-sharing controls now provide the correct action state when users toggle sharing.
  - Applications can handle the screen-sharing request and present the required system permission prompt.

- **Bug Fixes**
  - Fixed screen-sharing toggles using stale state, ensuring the requested action reflects the current control state.


andremion:
@coderabbit review

gpunto (approved):

PR #1834, opened 2026-09-15
Adopt the design system launcher icons in the demo app

andremion:
### Goal

Closes [AND-1503](https://linear.app/stream/issue/AND-1503/adopt-the-new-video-sample-app-launcher-icons-from-the-design-system)

Replace the demo app launcher icon with the artwork from design-system-tokens PR #70, resized in
PR #75 so the adaptive foreground matches the composed icon. The demo still shipped the old
artwork with the Android robot badge.

### Implementation

- The adaptive icon background is the README gradient (`#0A86FF` to `#005FFF`) as a shape
  drawable. The foreground is a per-density lossless webp in the mipmap folders, replacing the
  vector with the robot badge.
- The exported `foreground.png` is the 72dp visible area of the icon, not a 108dp adaptive layer.
  Used as is, launcher masks clip the glyph on every side. The layer is padded to 108/72 of the
  artwork so the visible window matches the composed `icon.png`.
- Legacy icons for API 25 and below: `ic_launcher` is the composed icon, `ic_launcher_round` is
  the same icon with a circle mask, all five densities as lossless webp. The Play Store artwork
  is regenerated at 512px.
- The development and e2etesting flavors keep a green background so dogfooding builds stay
  distinguishable from production. The green is now a gradient (`#0AB530` to `#008E30`); the
  design system has no dev color, so these values are not from the design.
- The livestream media notification used to download a remote image for its large icon and album
  art, with `stream_calls_logo` as fallback. It now uses the launcher icon. Decoding goes through
  the drawable because `BitmapFactory.decodeResource` returns null for the adaptive icon XML on
  API 26 and above.

### 🎨 UI Changes

Production and development launcher icons on the API 30 emulator, before on the top row and
after on the bottom row.

Launcher icon before and after, production and development

### Testing

Install the production and development flavors and check the launcher icon. On API 26 and above
the adaptive icon shows the glyph unclipped under the circle and rounded masks. Start a
livestream call and check the media notification shows the launcher icon as its large icon.

`./gradlew spotlessCheck apiCheck` passes. The demo app has no unit tests and no API dump.
`:demo-app:compileDevelopmentDebugKotlin`, `:demo-app:processProductionDebugResources` and
`:demo-app:processE2etestingDebugResources` pass. Both flavors were installed on the API 30 and
API 35 emulators and the launcher shows the new icon. The notification path is compile-verified
only.

## Summary by CodeRabbit

- **Visual Updates**
  - Updated the app’s launcher backgrounds with refreshed green and blue gradient designs across development, testing, and production builds.
  - Updated adaptive launcher icon configuration to use the current foreground artwork.

- **Bug Fixes**
  - Notification large icons now use bundled launcher artwork instead of retrieving images from a remote URL, improving consistency and reliability when network access is unavailable.


andremion:
@coderabbit review

demo-app/src/main/res/drawable/ic_launcher_foreground.xml
```diff
@@ -1,44 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
gpunto, line 1: Non-blocking: `demo-app/lint-baseline.xml` still carries a `VectorPath` entry anchored at this file (lines 63-72). With the drawable gone, lint reports it as a no-longer-reproducible baseline issue on every run. Nothing breaks (`abortOnError = false` in `build-logic/convention/src/main/kotlin/io/getstream/video/KotlinAndroid.kt:17`), but the block may as well be deleted alongside the file.

demo-app/src/development/res/drawable/ic_launcher_background.xml
```diff
@@ -14,27 +14,9 @@
...
-          android:endX="0"
-          android:endY="107"
-          android:type="linear">
-        <item android:offset="0" android:color="#FF49DDED"/>
-        <item android:offset="1" android:color="#0049DDED"/>
-      </gradient>
-    </aapt:attr>
-  </path>
-</vector>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
```
gpunto, line 17: Non-blocking: the flavor overrides the background and ships a green `ic_launcher_round.webp`, but there is no flavor `ic_launcher.webp`. With `minSdk = 24`, API 24-25 devices (and any launcher that reads `android:icon` rather than `android:roundIcon`) show the production blue square icon for a development build, which is exactly the case the green is meant to cover. Pre-existing rather than something this PR introduces, but it regenerated these folders, so it is a natural place to add the green square set. Same for `e2etesting`.

demo-app/src/main/kotlin/io/getstream/video/android/notification/LiveStreamMediaNotificationInterceptor.kt
```diff
@@ -94,13 +92,7 @@ class LiveStreamMediaNotificationInterceptor(private val context: Context) : Str
...
     private fun getStreamLogoBitmap(): Bitmap? {
         if (bitmap == null) {
-            bitmap = try {
-                URL("https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQ8dzj-6rSfEfYMOXPSCV3s84Luuqr2c9KzMg&s").openStream()
-                    .use { BitmapFactory.decodeStream(it) }
-            } catch (e: Exception) {
-                // Fallback
-                BitmapFactory.decodeResource(context.resources, R.drawable.stream_calls_logo)
-            }
+            bitmap = ContextCompat.getDrawable(context, R.mipmap.ic_launcher)?.toBitmap()
```
gpunto, line 95: Non-blocking, and I do not think it is worth changing: on API 26+ this resolves to `mipmap-anydpi-v26/ic_launcher.xml`, so `toBitmap()` renders an `AdaptiveIconDrawable`, whose `draw()` paints the composited layers through the `config_icon_mask` path. The large icon and `METADATA_KEY_ALBUM_ART` therefore come out mask-shaped with transparent corners, and the shape varies per OEM rather than being the full-bleed logo the old code produced.

Flagging it mainly so it is a known property. `ic_launcher_round` is not an escape hatch (it is an adaptive icon too), and the only unmasked option is `stream_calls_logo`, which is the artwork this PR is moving away from. Masked album art reads fine in a media notification.

Intrinsic size is positive here, so `toBitmap()` cannot throw: `AdaptiveIconDrawable.getIntrinsicWidth()` takes the max of the layer intrinsics, and the foreground is now a real bitmap even though the `<shape>` background reports -1.

gpunto (approved):
Looks good, just minor things you could ignore

demo-app/src/main/res/drawable/ic_launcher_foreground.xml
andremion: Good catch, removed in 31361ca316f65e137a86b5587ffc6c59b8b9dfd4.

demo-app/src/development/res/drawable/ic_launcher_background.xml
andremion: Fair point, the generator already had the green composed icon so it was a one-line addition. Both flavors now ship ic_launcher.webp at all densities in 31361ca316f65e137a86b5587ffc6c59b8b9dfd4.

demo-app/src/main/kotlin/io/getstream/video/android/notification/LiveStreamMediaNotificationInterceptor.kt
andremion: Agreed, leaving it as is. The masked shape reads fine in the media notification and the only unmasked artwork would be the old logo we are retiring. Thanks for checking the intrinsic size path too.

PR #1835, opened 2026-09-16
Caller recovers a dropped ring outcome by polling ring state

aleksandar-apostolov:
Closes [AND-1534](https://linear.app/stream/issue/AND-1534/caller-side-ring-state-polling-reconcile-a-dropped-ring-outcome-by)

### Goal

Caller-side ring-state polling. Backend VID-1444, JS client GetStream/stream-video-js#2393.

### Implementation

- `core/ringing/`: `RingStatePoller`, `RingStatePollingConfig`, `RingJoinSource`. Caller-only, armed alongside the auto-drop timer.
- Reads after 15s without a ring event, then every 5s. Bounded by the ring window (`autoCancelTimeoutMs` → `missedCallTimeoutMs` → 30s), a 60s ceiling, and a terminal 400/404.
- `CallState.updateFromRingState` merges the polled maps into `joinAcceptedOutgoingCall`, shared with the websocket handlers.
- Ringing state is recomputed after `join()` returns rather than on the next coordinator event.
- Watchdogs stay armed on their original deadlines until the call is joined; the deadline no longer rejects an accepted ring.
- `getCallRingState`, `GetCallRingStateResponse` and `ClientEvent.source` transplanted from generator output — a v1 regeneration carries four breaking renames that belong on v2.
- `CoordinatorJoin` reports `source`: `ring-ws` / `ring-poll-api`.

### Testing

```
./gradlew :stream-video-android-core:testDebugUnitTest
./gradlew :stream-video-android-core:apiDump
./gradlew spotlessCheck
```

24 unit tests. E2E `RingingTests` 11/11 on API-34.

Two devices on Pronto: emulator rings phone, coordinator websocket blackholed 200ms after the ring so `call.accepted` cannot arrive, callee accepts. The poll retrieves it at +15s and `Active` follows 0.76s after the join starts; the socket only reconnected 16.7s later. The same run before the state fix took 17.8s.

Failed-join retry is not covered — it needs a join to fail while the socket is down.

## Summary by CodeRabbit

- **New Features**
  - Added fallback polling for outgoing call ring states when real-time events are delayed or unavailable.
  - Calls can now automatically join when a callee accepts through the polling fallback.
  - Added configurable polling timing and duration settings.
  - Join analytics now identify whether a join came from real-time events or polling.
- **Bug Fixes**
  - Improved handling of missed, accepted, rejected, and completed ring states.
  - Prevented unnecessary polling after calls are settled or cleaned up.
- **Tests**
  - Added coverage for polling, ring-state reconciliation, automatic joining, and join-source tracking.


stream-video-android-core/api/stream-video-android-core.api
```diff
@@ -2477,22 +2478,23 @@ public final class io/getstream/android/video/generated/models/ClientEvent {
...
+	public static synthetic fun copy$default (Lio/getstream/android/video/generated/models/ClientEvent;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Integer;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lorg/threeten/bp/OffsetDateTime;Ljava/lang/Integer;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lorg/threeten/bp/OffsetDateTime;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Boolean;ILjava/lang/Object;)Lio/getstream/android/video/generated/models/ClientEvent;
```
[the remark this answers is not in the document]
aleksandar-apostolov: Deliberate — this file is a byte-identical transplant of generator output, and `source` sits where the generator puts it. Appending it would diverge, so the next regeneration would move it back; that regeneration lands on `develop-v2` where the reorder ships as a released break. Every field is nullable with a default and the model is only built internally, with named arguments.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/call/components/CallJoinCoordinator.kt
```diff
@@ -266,7 +269,7 @@ internal class CallJoinCoordinator(
                 ring,
                 notify,
                 hintHighScaleLivestreamPublisher,
-                JoinAnalyticsModel(retryCount, JoinReason.FirstAttempt),
+                JoinAnalyticsModel(retryCount, JoinReason.FirstAttempt, consumeJoinSource()),
```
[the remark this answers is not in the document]
aleksandar-apostolov: recheck

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
```diff
@@ -0,0 +1,159 @@
...
+    fun stop() {
+        job?.cancel()
+        job = null
+    }
+
+    /** @return false when polling should stop. */
+    private suspend fun poll(callSessionId: String): Boolean {
+        when (val result = fetch(callSessionId)) {
+            is Result.Success -> {
+                logger.d { "[poll] ${result.value}" }
```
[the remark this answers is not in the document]
aleksandar-apostolov: recheck

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePollingConfig.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * When the caller falls back to reading the ring state instead of waiting for it.
+ *
+ * @param startAfterMs how long a ring must go without any ring event before polling begins. Kept
+ * well above the coordinator's websocket ping interval so a merely slow answer does not poll.
+ * @param intervalMs the gap between reads once polling has begun.
+ */
+public data class RingStatePollingConfig(
+    val startAfterMs: Long = DEFAULT_START_AFTER_MS,
+    val intervalMs: Long = DEFAULT_INTERVAL_MS,
```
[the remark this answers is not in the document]
aleksandar-apostolov: recheck

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
```diff
@@ -300,6 +305,58 @@ public class Call(
...
+     * calls never ring at all.
+     */
+    internal val ringStatePoller: RingStatePoller
+        get() = _ringStatePoller ?: RingStatePoller(
+            // A real IO dispatcher rather than the call scope or DispatcherProvider: this is a
+            // background watchdog issuing network reads on a wall-clock schedule, and the ring
+            // window bounding it is wall-clock too. Running it on a scheduler a caller can
+            // replace makes those reads someone else's to drain. Unit tests inject their own
+            // scope and clock instead.
+            scope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
```
rahul-lohra: The ownership here is unclear.

`RingStatePoller` is instantiated and retained by `Call`, which suggests its lifecycle belongs to that `Call`. However, its coroutine runs in a new root scope:

```kotlin
CoroutineScope(SupervisorJob() + Dispatchers.IO)
```

That means the object owner and coroutine owner have different lifecycles. Cancelling the call scope does not cancel work started by the object owned by the call, so cleanup depends on remembering to call `stopRingStatePolling()` manually.

I think we should choose one consistent model:
1. If polling belongs to a `Call`, run it as a child of the call scope and override only the dispatcher if required.
2. If polling is intentionally expected to outlive a `Call`, instantiate and own it at the client/user level, using the client/user scope, rather than constructing it inside `Call`.

Given that the poller reads one call session and writes directly into that call’s `CallState`, the first model seems correct. What specific lifecycle event must this poller survive that the existing call scope does not survive?

Also, the explanation about using a “real IO dispatcher” only explains thread selection. It does not explain why an independent root `SupervisorJob` is required.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePollingConfig.kt
```diff
@@ -0,0 +1,68 @@
...
+        /**
+         * The longest a single ring may be polled for, whatever the call's ring settings say.
+         *
+         * The polling window is otherwise the app's own `auto_cancel_timeout_ms`, which the app
+         * chooses and can set to minutes. One read every [intervalMs] for that whole window,
+         * multiplied by every ringing caller, is real load on a shard — and the endpoint is rate
+         * limited per app, so it would be the customer's own allowance being spent. The ceiling
+         * belongs to the SDK rather than to configuration we do not control.
+         */
+        public const val DEFAULT_MAX_DURATION_MS: Long = 60_000
```
rahul-lohra: I think this companion object and its variable can be made private or internal

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePollingConfig.kt
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param startAfterMs how long a ring must go without any ring event before polling begins. Kept
+ * well above the coordinator's websocket ping interval so a merely slow answer does not poll.
+ * @param intervalMs the gap between reads once polling has begun.
+ */
+public data class RingStatePollingConfig(
+    val startAfterMs: Long = DEFAULT_START_AFTER_MS,
+    val intervalMs: Long = DEFAULT_INTERVAL_MS,
+    val maxDurationMs: Long = DEFAULT_MAX_DURATION_MS,
+    val defaultRingWindowMs: Long = DEFAULT_RING_WINDOW_MS,
```
rahul-lohra: We don't have kdocs for these 2 variable `maxDurationMs` and `defaultRingWindowMs`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePollingConfig.kt
```diff
@@ -0,0 +1,68 @@
...
+        public const val DEFAULT_RING_WINDOW_MS: Long = 30_000
+
+        /**
+         * The longest a single ring may be polled for, whatever the call's ring settings say.
+         *
+         * The polling window is otherwise the app's own `auto_cancel_timeout_ms`, which the app
+         * chooses and can set to minutes. One read every [intervalMs] for that whole window,
+         * multiplied by every ringing caller, is real load on a shard — and the endpoint is rate
+         * limited per app, so it would be the customer's own allowance being spent. The ceiling
+         * belongs to the SDK rather than to configuration we do not control.
```
rahul-lohra: > The ceiling belongs to the SDK 

This KDoc seems to be misleading. `DEFAULT_MAX_DURATION_MS` is only the default value for the publicly configurable `maxDurationMs`; _it is not an SDK-enforced ceiling_. Customers can increase it to any positive value.
Either:
- enforce an internal hard maximum and allow configuration only below it, or
- update the documentation to call this a configurable default rather than an SDK-owned ceiling.
The current documentation promises protection that the code does not guarantee.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -857,24 +859,16 @@ public class CallState(
             }
 
             is CallAcceptedEvent -> {
+                call.recordRingEvent()
```
rahul-lohra: ## Circular & inverse dependency
The dependency direction here feels inverted:
```kotlin
CallState → Call.recordRingEvent() → RingStatePoller
RingStatePoller → CallState.updateFromRingState()
```
`CallState` needs to know when to update the poller’s internal timing, while the poller writes results back into `CallState`.

This creates circular 🌀 orchestration and leaks polling implementation details into the state reducer.

How about this proposal - 

The poller should observe an injected event stream rather than reaching into Call directly:

```kotlin
internal class RingStatePoller(
    private val scope: CoroutineScope,
    private val ringOutcomes: Flow<RingOutcome>,
    ...
)


val ringOutcomes = call.events
    .mapNotNull { event ->
        when (event) {
            is CallAcceptedEvent -> ringStatePoller.onRingEvent()
            is CallRejectedEvent -> ringStatePoller.onRingEvent()
            is CallMissedEvent -> ringStatePoller.onRingEvent()
            else -> null
        }
    }
```
## Incorrect name

_Note: The method name: `RingStatePoller.onRingEvent()` is also incorrect, I will comment a solution on RingStatePoller class_

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
```diff
@@ -0,0 +1,167 @@
...
+                    continue
+                }
+
+                if (!poll(sessionId)) return@launch
+            }
+        }
+    }
+
+    /** Records that a ring event arrived, restarting the quiet period. */
+    fun onRingEvent() {
```
rahul-lohra: The name `onRingEvent()` is misleading and will be difficult to interpret when revisiting this code later. It reads as though the method handles every ringing-related event, potentially including the actual `CallRingEvent`.

In practice, it is called only when a participant’s ring status changes to one of three states: _**accepted, rejected, or missed.**_

Could we rename it to something more explicit, such as:

```kotlin
onRingParticipantStatusUpdate()
```

This communicates both the scope of the method and why it updates the polling timer, without implying that it handles all ringing events.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
```diff
@@ -0,0 +1,167 @@
...
+    private var job: Job? = null
+
+    /**
+     * The last time a ring event proved the socket was still delivering. Polling waits for
+     * [RingStatePollingConfig.startAfterMs] of quiet measured from here, so an event pushes the
+     * first read further out rather than cancelling the poller: in a group ring a single rejection
+     * does not settle the outcome, and the ring may still go silent afterwards.
+     */
+    @Volatile
+    private var lastRingEventAt: Long = 0
```
rahul-lohra: `lastRingEventAt` does not seem to represent the value accurately.
It is initialized when polling starts, even though no ring event may have occurred. It is also updated only for accepted, rejected and missed participant-status updates—not for every ringing-related event—and it stores the local observation time rather than the event’s timestamp.
The field is actually the start of the polling grace period. Could we rename it to:
```kotlin
pollingGracePeriodStartedAtMs
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
```diff
@@ -0,0 +1,167 @@
...
+     * Begins watching a ring.
+     *
+     * @param callSessionId reads the call's current session. Resolved on each turn until it first
+     * returns a value and latched from then on: an outgoing ring does not always have a session by
+     * the time it starts, and once the call ends the call no longer has a current session — which
+     * is precisely the case polling exists to answer. Latching covers both.
+     * @param ringTimeoutMs the ring window, resolved once polling starts. Polling stops there, so
+     * it always resolves before the local auto-drop rather than racing it.
+     */
+    fun start(callSessionId: () -> String?, ringTimeoutMs: () -> Long?) {
```
rahul-lohra, line 95: > SonarQube reports that this method has a cognitive complexity of 20, above the configured limit of 15.

I agree with the warning: `start()` currently handles lifecycle checks, deadline calculation, polling delay calculation, session-ID resolution, request execution and termination conditions. That makes the control flow difficult to follow and verify.

Could we split these responsibilities into smaller named functions—for example, deadline resolution, waiting for the next poll, session-ID resolution and executing one poll? I’ll follow up with a concrete refactoring suggestion after tracing the timing edge cases.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
```diff
@@ -0,0 +1,167 @@
...
+     * The last time a ring event proved the socket was still delivering. Polling waits for
+     * [RingStatePollingConfig.startAfterMs] of quiet measured from here, so an event pushes the
+     * first read further out rather than cancelling the poller: in a group ring a single rejection
+     * does not settle the outcome, and the ring may still go silent afterwards.
+     */
+    @Volatile
+    private var lastRingEventAt: Long = 0
+
+    /**
+     * Begins watching a ring.
```
rahul-lohra, line 86: Do you think we should add this into KDoc as well _(it can help understand the polling algorithm)_: 

```text
Polling uses a fixed-delay schedule:

1. When polling starts, it waits until no relevant participant-status
   update has been received for [startAfterMs].
2. It then performs the first ring-state request.
3. After each request and its reconciliation complete, it waits
   [intervalMs] before starting the next request.
4. If a relevant participant-status update arrives while waiting, the next
    request is postponed until another [startAfterMs] period has elapsed.

Request execution time is not included in [intervalMs]. For example, with
a five-second interval and a two-second request, consecutive requests start approximately seven seconds apart.
```

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
rahul-lohra: I understood the logic, so refactoring is purely optional

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -1491,6 +1495,73 @@ public class CallState(
...
+     * Shared by the `call.accepted` handler and the ring-state poller: one arrives at the outcome
+     * from a websocket event and the other from a polled read, and both must act on it identically.
+     * It reads only state, never an event payload, so the decision cannot drift between them.
+     *
+     * Acting twice is prevented by [autoJoiningCall] and the active-call check rather than by
+     * de-duplicating the trigger, so a polled accept arriving alongside a live one is harmless.
+     *
+     * @return true when a join was started by this call.
+     */
+    internal fun joinAcceptedOutgoingCall(source: RingJoinSource): Boolean {
```
rahul-lohra, line 1543: I don't think this PR has test-case for `joinAcceptedOutgoingCall`. Please double check

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
aleksandar-apostolov: Happy to make it a `Flow` off `call.events` and drop `recordRingParticipantStatusUpdate()`. Separate PR though, so this one stays reviewable — sound good?

rahul-lohra: No problem 👍

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/model/JoinAnalyticsModel.kt
```diff
@@ -16,7 +16,13 @@
 
 package io.getstream.video.android.core.analytics.call.observer.model
 
-internal data class JoinAnalyticsModel(val retryAttempt: Int, val joinReason: JoinReason? = null)
+import io.getstream.video.android.core.ringing.RingJoinSource
+
+internal data class JoinAnalyticsModel(
+    val retryAttempt: Int,
+    val joinReason: JoinReason? = null,
+    val joinSource: RingJoinSource? = null,
```
rahul-lohra, line 24: Could we rename `joinSource` to `ringJoinSource` or `ringJoinTrigger`?

`joinSource` sounds like a generic source for every join lifecycle, but the value is specifically a `RingJoinSource` and only distinguishes `websocket` versus `polling-driven` ring acceptance.

Keeping the internal name ring-specific would prevent readers from assuming this field also describes manual joins, reconnects or notification-driven joins. 

What do you think?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
```diff
@@ -857,24 +859,16 @@ public class CallState(
...
                 updateRingingState()
 
-                // auto-join the call if it's an outgoing call and someone has accepted
-                // do not auto-join if it's already accepted by us
                 val callRingState = _ringingState.value
-                if (callRingState is RingingState.Outgoing && callRingState.acceptedByCallee && _acceptedBy.value.findLast {
-                        it == client.userId
-                    } == null && client.state.activeCall.value == null && autoJoiningCall == null
+                if (!joinAcceptedOutgoingCall(RingJoinSource.WebSocket) &&
+                    callRingState is RingingState.Incoming && event.user.id == client.userId
```
rahul-lohra, line 878: Do you think we should add include this as well `&& callRingState.acceptedByMe` ?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/Call.kt
aleksandar-apostolov: Addressed in the latest push.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePollingConfig.kt
aleksandar-apostolov: Addressed in the latest push.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePollingConfig.kt
aleksandar-apostolov: Addressed in the latest push.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
aleksandar-apostolov: Addressed in the latest push.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
aleksandar-apostolov: Addressed in the latest push.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/ringing/RingStatePoller.kt
aleksandar-apostolov: Addressed in the latest push.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
aleksandar-apostolov: `CallStateRingJoinTest` covers it — joins on a polled accept, skips when the accept is our own, skips when nobody answered, and keeps the watchdogs armed when the join fails.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/model/JoinAnalyticsModel.kt
aleksandar-apostolov: The type already says ring, so `ringJoinSource: RingJoinSource` reads a bit redundant — and it feeds `ClientEvent.source`, which the backend kept generic on purpose. Only ring paths set it today, but a push- or deeplink-driven join would use the same field. Keep `joinSource`?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
rahul-lohra, line 1543: The function is a conditional command:

- It checks whether joining is allowed.
- It starts a join as a side effect.
- It returns whether that side effect was started.

The method name sounds like an unconditional action, but the method first checks several conditions and returns whether an asynchronous join was started.

Do you think we should rename it to something like:
```kotlin
tryStartJoinForAcceptedOutgoingCall()
```
The `try` prefix makes the Boolean return meaningful, while `start` clarifies that the method launches a coroutine rather than completing the join.

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
aleksandar-apostolov: It's implied — `updateRingingState()` above has already folded this accept into `_acceptedBy`, so if the state is still `Incoming` and the accept is ours, `acceptedByMe` is true by construction. Add it anyway if you'd rather it read explicitly?

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/analytics/call/observer/model/JoinAnalyticsModel.kt
rahul-lohra: Okay we can keep it named `joinSource`

stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/CallState.kt
rahul-lohra: Its fine, lets skip this

rahul-lohra (approved):
Tested the flow by skipping all the Coordinator Events. PR is good to merge

PR #1836, opened 2026-09-16
Refresh the remaining SDK call surfaces with the design system tokens

andremion:
### Goal

Closes [AND-1520](https://linear.app/stream/issue/AND-1520/refresh-the-remaining-sdk-call-surfaces-with-the-design-system-tokens)

The SDK call surfaces left after AND-1425 move to the design system tokens: audio room, livestream, screen sharing, participant avatars, the tile pin menu and the ringing details. This is a refresh, not a redesign, so layout and functionality stay as they are on `develop-v2` and every size moves to the nearest token, the same rule the lobby and video calling refreshes used. The scope rules are on the parent issue AND-1421.

### Implementation

Token pass:

* Audio room: app bar title, divider, participant name and role, grid padding and the speaking border use tokens. The grid keeps its 100dp minimum cell because no token fits: 80dp leaves the microphone badge covering the avatar.
* Livestream: the backstage, ended, error and host-video-not-available texts use the typography tokens; the overlay badge, count, duration and speaker button use spacing, size and component tokens.
* Screen sharing: radius tokens only, no visible change.
* `ParticipantAvatars` and `ParticipantActions` use size and spacing tokens; the ringing details spacers use spacing tokens.
* The 45dp size overrides on the audio and livestream toggle buttons are gone, `StreamButtonSize.Large` already enforces 48dp.

Found on the device comparison against 1.x and fixed here:

* `InitialsAvatar` scales the initials from 56dp instead of 80dp, so avatars between the two no longer drop to 18sp on one phone and scale on another.
* The audio room Leave Quietly button is `primarySolid` again. Its 1.x style was the brand blue, the base component rebuild had mapped it to grey.
* New `StreamButtonSize.ExtraLarge` at 64dp for the ringing controls in the SDK and the demo app. 1.x drew them at 66dp, the base component rebuild had left them at 48dp.
* The livestream overlay count and duration sit on the dark overlay pill the tile label uses, and the count icon is tinted. Before, the texts were white on the white host-video-not-available screen in light mode and the icon was black on black in dark mode.

Previews and goldens:

* New previews and goldens for the livestream ended, error, host-video-not-available and deprecated backstage states. The backstage preview renders the real component instead of a hand-drawn copy. The recordings fetch is skipped in inspection mode so the preview does not touch the network.
* The toggle-in-progress preview stacks one row per button size so the fourth size fits the snapshot.

Review round:

* The scope survey had missed a few surfaces, tokenized now: the moderation warning and its theme defaults, the video renderer fallbacks, the diagnostics spacers, the call activity delegate, the audio call spacer, the factory tile menu padding, the floating tile zero paddings and the initials cap.
* Kept literal on purpose: the 100dp tile height in the spotlight and lazy renderers and the 100dp audio grid cell, because the nearest token (80dp) changes the layout; the moderation container offset; the lobby heights per device class; the public `UserAvatarBackground` default; elevations; arithmetic zeros; the diagnostics debug colors. `CallParticipantsList`, `InviteUserList` and `CallParticipantsInfoActions` belong to AND-1538, which waits on the AND-1527 decision.
* The participant label no longer runs under the connection indicator on narrow tiles. When the label sits at its default bottom-start position, both share one row and the label ellipsizes in the space the indicator leaves. This was already the case on 1.x.

### 🎨 UI Changes
1.x on the left, this branch on the right. Dark mode in the table, light mode in the collapsed block.

<details>
<summary>Light mode</summary>

| Screen | Light |
| --- | --- |
| Audio room, Pixel 8 (API 35) | audio_room_api35 |
| Audio room, Pixel 4 (API 30) | audio_room_api30 |
| Livestream, backstage (guest) | livestream_backstage_guest |
| Livestream, live (guest) | livestream_live_guest |
| Livestream, host video not available (guest) | livestream_host_video_not_available_guest |
| Livestream, ended (guest) | livestream_ended_guest |
| Livestream, backstage (host, tutorial UI) | livestream_host_backstage |
| Livestream, live (host, tutorial UI) | livestream_host_live |
| Outgoing call | ringing_outgoing |
| Outgoing call, two callees | ringing_outgoing_two_callees |
| Incoming call | ringing_incoming |

</details>

| Screen | Dark |
| --- | --- |
| Audio room, Pixel 8 (API 35) | audio_room_api35_dark |
| Audio room, Pixel 4 (API 30) | audio_room_api30_dark |
| Livestream, backstage (guest) | livestream_backstage_guest_dark |
| Livestream, live (guest) | livestream_live_guest_dark |
| Livestream, host video not available (guest) | livestream_host_video_not_available_guest_dark |
| Livestream, ended (guest) | livestream_ended_guest_dark |
| Livestream, backstage (host, tutorial UI) | livestream_host_backstage_dark |
| Livestream, live (host, tutorial UI) | livestream_host_live_dark |
| Outgoing call | ringing_outgoing_dark |
| Outgoing call, two callees | ringing_outgoing_two_callees_dark |
| Incoming call | ringing_incoming_dark |

### Testing

Ran the audio, livestream and ringing tutorial apps on two emulators, a Pixel 8 on API 35 and a Pixel 4 on API 30, once built from `develop` and once from this branch, in light and dark mode. Checked the audio room with two participants, the livestream guest through backstage, live, host camera off and ended, and the incoming and outgoing ringing screens with one and two callees.

To reproduce: run `tutorial-livestream` on two devices with the same call id, host on one and guest on the other, then start the broadcast, toggle the host camera and stop the broadcast. Run `tutorial-ringing` on two devices logged in as User 1 and User 2 and dial from one to the other.

Validation on the final tree:

```
./gradlew spotlessCheck apiCheck testDebugUnitTest
./gradlew :stream-video-android-ui-compose:recordPaparazziDebug
./gradlew :stream-video-android-ui-compose:verifyPaparazziDebug
```

The goldens were re-recorded once, as the parent issue allows. The API dump gains one line for `StreamButtonSize.ExtraLarge`. There is no `detekt` task in this repo.

## Summary by CodeRabbit

* **New Features**
  * Added extra-large controls for incoming and outgoing calls, improving visibility and touch targets.
  * Updated livestream overlays with clearer participant and live-duration indicators.
  * Improved avatar text sizing across different avatar dimensions.

* **UI Improvements**
  * Standardized spacing, typography, button styling, colors, and corner radii across audio, calling, livestream, and screen-sharing interfaces.
  * Improved livestream ended, error, and host-video-unavailable presentations, including centered messaging and consistent layouts.

* **Tests**
  * Expanded visual coverage for livestream and call-related UI states.


andremion:
@coderabbit review

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/avatar/InitialsAvatar.kt
```diff
@@ -80,7 +80,7 @@ internal fun Dp.toAvatarTextStyle(): TextStyle {
         this < StreamTokens.size24 -> typography.metadataEmphasis
         this < StreamTokens.size40 -> typography.captionEmphasis
         this < StreamTokens.size48 -> typography.bodyEmphasis
-        this < StreamTokens.size80 -> typography.headingMedium
+        this < StreamTokens.size56 -> typography.headingMedium
```
gpunto, line 82: The KDoc above this function is now stale: it still says "Avatars from 80dp up have no matching text style in the design tokens, so the initials scale with the avatar instead", but the fall-through to the scaling `else` starts at 56dp on this line.

Someone reading the doc to pick an avatar size will expect a 64dp avatar to render `headingMedium` (18sp) when it actually renders scaled 25.6sp initials.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/pinning/ParticipantActions.kt
```diff
@@ -238,7 +237,7 @@ internal fun BoxScope.ParticipantActionsDialogContent(
                 shape = RoundedCornerShape(StreamTokens.radiusLg),
             )
             .align(Center)
-            .width(220.dp),
+            .width(StreamTokens.size208),
```
gpunto: Non-blocking, but worth knowing the cost of this particular snap.

`StreamListItem` chrome eats exactly 64dp here (4+4 outer padding, 12+12 inner, 20dp icon, 12dp gap), so the title column goes 156dp → 144dp. I measured `"Unpin for everyone"` against the Roboto that layoutlib ships: **133.1dp at 16sp**. It still fits at the default font scale, but it starts ellipsising at scale ≈1.08 — i.e. Android's "Large" setting (1.15) — where 220dp held to ≈1.17.

No golden covers it either; the snapshot only renders `"Pin"`. There is no 220dp token, so the options are to accept it, let the title wrap to two lines, or go to `size240`. Your call — just flagging that this spends the last of the headroom on an already-tight menu.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/livestream/LivestreamPlayerOverlay.kt
```diff
@@ -142,12 +149,18 @@ private fun BoxScope.LiveControls(
...
+        modifier = Modifier.align(Alignment.CenterEnd),
         isSpeakerphoneEnabled = isAudioEnabled,
         onCallAction = { callAction ->
             onAudioToggle(callAction.isEnabled)
         },
     )
 }
+
+/** The dark translucent background that keeps the overlay readable over any video or fallback screen. */
+private fun Modifier.overlayPill(): Modifier = composed {
```
gpunto: Non-blocking nit. This only reads `VideoTheme.colors`, so a plain `@Composable fun overlayPillModifier(): Modifier` (or just inlining the two-line `background(...)` at both call sites) would do the same with none of `composed`'s downsides — no `inspectorInfo`, never skipped, breaks modifier equality.

`call.state.duration` is a one-second ticker (`CallState.kt:553`), so `LiveDuration` re-runs this factory every second. The cost is trivial; the simpler form is just free. There is precedent for `composed` in `SpotlightVideorenderer.kt`, so take it or leave it.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/base/StreamButtonSize.kt
```diff
@@ -49,4 +49,11 @@ public enum class StreamButtonSize(
         iconSize = StreamTokens.iconSizeMd,
         labelPadding = StreamTokens.buttonPaddingXWithLabelLg,
     ),
+
+    /** 64dp tall, 20dp icons. The size of the ringing call controls. */
+    ExtraLarge(
+        minimumSize = StreamTokens.size64,
```
gpunto: Non-blocking. Every other entry sizes itself from a `buttonVisualHeight*` token; this one reaches for a generic `size64` because the button-height scale stops at 48dp.

That is the right pragmatic choice here. But if the design system should grow a `buttonVisualHeightXl`, doing it in this PR is cheaper than later — `StreamTokens` is public API, so it means another `.api` dump bump whenever it happens.

stream-video-android-ui-compose/src/debug/kotlin/io/getstream/video/android/compose/ui/components/livestream/LivestreamBackStagePreview.kt
```diff
@@ -16,58 +16,34 @@
...
-                color = VideoTheme.colors.textPrimary,
-            )
-            Text(
-                modifier = Modifier,
-                text = "2 participants have joined the call",
-                fontSize = 12.sp,
-                color = VideoTheme.colors.textSecondary,
-            )
-        }
+        LivestreamBackStage(call = previewCall)
```
gpunto: Non-blocking. Rendering the real component instead of the hand-built fake is the right move, but it does cost coverage: `previewCall` never sets `startsAt` and its session is null, so the new golden is just the heading — `CountDownTimerUi` and `ParticipantCountUi` are no longer exercised at all, where the old fake at least showed the layout.

`CountDownTimerUi` is `internal` and takes a plain `String`, so a small dedicated preview would cover it without having to touch the shared `previewCall`.

gpunto (approved):
One question on scope: the title says "remaining", but three live call surfaces still carry raw `dp`/`sp`: `call/moderation/ModerationWarningUiContainer.kt`, `participants/internal/CallParticipantsList.kt` and `video/VideoRenderer.kt`. Are they intentionally out of scope?

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/avatar/InitialsAvatar.kt
andremion: Fixed in b26a39698ea62d1729712295699ccbc5f1536ecc, the doc says 56dp now.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/livestream/LivestreamPlayerOverlay.kt
andremion: Went with the plain @Composable factory, b26a39698ea62d1729712295699ccbc5f1536ecc.

andremion:
Good catch, "remaining" was not accurate.

`CallParticipantsList` is intentional: it is part of the participants info menu that AND-1527 proposes to remove, so its token pass sits in AND-1538 until the team decides.

`ModerationWarningUiContainer` and `VideoRenderer` were missed by the scope survey, together with a few smaller ones (diagnostics spacers, the call activity delegate, the moderation theme defaults, the audio call spacer). All tokenized in b26a39698ea62d1729712295699ccbc5f1536ecc. The Implementation section now lists what stays literal on purpose: the 100dp tile height in the spotlight and lazy renderers and the 100dp audio grid cell, because the nearest token is 80dp and that changes the layout, plus offsets, elevations and the debug colors.

While comparing the goldens with 1.x I also noticed the participant label running under the connection indicator on narrow tiles. It was already like that on 1.x, fixed in ab5234443126f6febc28fba23f3123a205167eb2.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
```diff
@@ -187,12 +185,35 @@ public fun ParticipantVideo(
...
+            Row(
+                modifier = Modifier
+                    .align(BottomStart)
+                    .fillMaxWidth(),
+                horizontalArrangement = Arrangement.SpaceBetween,
+                verticalAlignment = Alignment.Bottom,
+            ) {
+                if (style.isShowingParticipantLabel) {
+                    Box(modifier = Modifier.weight(1f, fill = false)) {
+                        labelContent.invoke(this, participant)
```
gpunto: `labelContent` and `connectionIndicatorContent` are public slots typed `@Composable BoxScope.(...) -> Unit`, and the receiver has so far been the participant tile. In this branch they're invoked on wrap-content wrapper `Box`es instead (here and on line 206), so anything positional in integrator-supplied content silently changes meaning: a custom indicator written as `Modifier.align(Alignment.TopEnd)` now renders bottom-right beside the label, and a custom label using `matchParentSize()` to draw a full-tile scrim now covers only its own text.

Every built-in style defaults `labelPosition` to `BottomStart` (`VideoRendererStyle.kt:112`), so this is the path everyone is on, and there's no compile error to warn them.

The defaults are fine — I checked that `NetworkQualityIndicator` keeps its `height(size32)` and lands in the same place. So this is about the contract, not the rendering: worth either documenting the new receiver in the slot KDoc plus a changelog line, or keeping the slots on the tile's `BoxScope` and constraining the label width another way (e.g. `widthIn(max = ...)` off a `BoxWithConstraints`).

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
```diff
@@ -187,12 +185,35 @@ public fun ParticipantVideo(
...
+                    }
+                } else {
+                    Spacer(modifier = Modifier.weight(1f))
+                }
+                if (style.isShowingConnectionQualityIndicator) {
+                    Box { connectionIndicatorContent.invoke(this, connectionQuality) }
```
gpunto: Non-blocking, follows from the comment above. Wrapping the slot here makes two pieces of positioning code inert without removing them:

- `VideoComponentFactory.kt:322` — the default `ParticipantVideoConnectionIndicatorContent` still applies `.align(Alignment.BottomEnd)`, which is now a no-op inside this `Box`.
- `ParticipantLabel` still takes `labelPosition` and applies `.align(labelPosition)`, also a no-op on this path.

Both still read as live positioning. Someone customizing the factory and switching to `Alignment.TopEnd` will see nothing move and no obvious reason why. Worth a comment at each site saying the Row owns placement now, if the wrapper approach stays.

gpunto (commented):
Re-reviewed the two commits pushed since the approval, some inline comments

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
andremion: Agreed, the wrapper changed the receiver. In a5a82cbe3cbfed31f31b5a9f33280b574e51ee5f both slots are invoked on the tile scope again and the width cap sits inside ParticipantLabel: a BoxWithConstraints aligned at labelPosition caps the pill at the tile width minus the indicator's 32dp, passed through an internal composition local only while the indicator is shown. Same pixels, no golden moved.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/ui/components/call/renderer/ParticipantVideo.kt
andremion: Both alignments are live again with a5a82cbe3cbfed31f31b5a9f33280b574e51ee5f, nothing left to comment.

PR #1837, opened 2026-09-17
Deprecate the call participants info menu that no SDK screen opens

andremion:
### Goal

Refs [AND-1527](https://linear.app/stream/issue/AND-1527/remove-the-unreachable-call-participants-info-menu-and-invite-flow)

`CallParticipantsInfoMenu` lost its only call site in June 2023, when the participant invitation flow was removed from `CallContainer` and `CallViewModel`. Since then nothing in the SDK or the demo app has opened it, and its only reference is a debug preview. The demo app shows its own `ParticipantsDialog` instead.

The v2 branch removes the menu and the types around it in #1838. This PR gives 1.x integrators one release of notice first.

### Implementation

- `@Deprecated` with `WARNING` level on the three public declarations that go away in v2: `CallParticipantsInfoMenu`, the `CallParticipantInfoMode` interface, and `ParticipantInfoAction` with its `ChangeMuteState` and `InviteUsers` cases. Each message says the declaration is removed in v2; the menu message points at `call.state.participants` as the building block.
- Deprecation suppressions inside the same three files, where the declarations reference each other.
- Only the files that the v2 removal deletes are touched, so the later `develop` to `develop-v2` merge resolves each of them by keeping the deletion.
- The API dump does not change. Core's `InviteUsersToCall` action is left alone.

### Testing

No behaviour change, so there is nothing to try on a device. The deprecation warning shows up on any call to the three declarations, for example in `CallParticipantsInfoMenuPreview` in the compose module's debug source set. That preview is left as is on purpose, to keep the merge-down footprint to the three files above.

Ran on `stream-video-android-ui-compose`: `spotlessApply` (no edits), `compileDebugKotlin`, `apiCheck`.

## Summary by CodeRabbit

* **Deprecations**
  * Deprecated the participant information menu and related participant information types.
  * The SDK no longer opens the participant information menu; it is scheduled for removal in version 2.
  * Builders should use `call.state.participants` for participant data.


andremion:
@coderabbit review

aleksandar-apostolov (approved):

Issue #73, opened 2022-11-14
Setup Maven and Sonatype publications


skydoves:
Setup Maven and Sonatype publication.

Issue #68, opened 2022-11-08
Improve SDK performance by building baseline profile features


skydoves:
Improve SDK performance by building baseline profile features for UI components and our core SDK.

PR #1840, opened 2026-09-18
Re-sync the design tokens with design-system-tokens b62e2a5

andremion:
### Goal

Refs [AND-1435](https://linear.app/stream/issue/AND-1435/theming-wrap-up-retire-the-style-layer-token-re-sync-unused-resources)

Bring the vendored token classes up to date with the design-system-tokens repo at `b62e2a558a5934506daaea15642b305726af2603`. The previous snapshot was `4700c68`. The only semantic change between the two is design-system-tokens #73, plus the decline tokens from #68 that were never applied.

The style layer retirement and the unused resource cleanup from the same ticket follow in a separate PR stacked on this one.

### Implementation

Core tokens:
- `accent/warning` moves from yellow 400/300 to yellow 200/150 (light/dark). `border/utility/warning` follows because it references it. Two new primitives, `yellow150` and `yellow200`.
- New `backgroundCoreOnElevation` and `borderCoreOnElevation` on `StreamDesign.Colors`, for controls placed on elevated surfaces. Two new public properties, so the data class constructor and `copy` signatures change in the API dump.

Video tokens (`VideoDesignTokens.kt` now holds the full set from the source):
- `indicator/fair|great|poor|speaking` renamed under `connection-quality` and `sound-indicator`. The speaking colour moves from brand 300 to brand 400.
- New `call-control-error-badge` and `microphone-level` tokens. `AudioVolumeIndicator` defaults its bars to the microphone-level active colour instead of `accentPrimary`; the inactive bar token has no consumer yet.
- `DeclineCallAction` and `CancelCallAction` read the new decline tokens through a shared `declineCallStyle()` instead of `StreamButtonStyleDefaults.destructiveSolid`. Pixel-identical, since both resolve to `accentError` and `textOnAccent`.

No icons are added. The SDK and the demo show no blur-off or low-bandwidth icon today, so the like-for-like rule excludes the two icons the token repo added.

Six goldens are re-recorded, every differing pixel checked against the token values:
- moderation warning strip (two goldens) and the fair connection-quality bars: the new yellows
- sound indicator bars and the remote participant speaking border (light and dark): brand 400

### Testing

Ran on the ui-compose module: `compileDebugKotlin`, `apiDump`, `recordPaparazziDebug`, `verifyPaparazziDebug` (175 tests, 0 failures). Ran repo-wide: `spotlessCheck`, `apiCheck`. The demo app compiles.

To see the change on a device: open a call and speak, the tile border and the sound bars are brand 400 instead of the old blue and 300; trigger a moderation warning, the strip is the lighter yellow.

## Summary by CodeRabbit

* **New Features**
  * Added refined color tokens for elevated surfaces, borders, call controls, error badges, and microphone level indicators.
  * Added updated microphone and connection-quality indicator colors.
* **Improvements**
  * Updated decline and cancel call buttons with dedicated themed styling, including disabled states.
  * Improved warning colors for better visibility across light and dark themes.
  * Updated speaking indicators and audio volume bars to use the new visual design tokens.


andremion:
@coderabbit review

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/design/StreamDesign.kt
```diff
@@ -165,6 +168,7 @@ public object StreamDesign {
         public val borderCoreStrong: Color,
         public val borderCoreOnAccent: Color,
         public val borderCoreOnSurface: Color,
+        public val borderCoreOnElevation: Color,
```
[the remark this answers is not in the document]
andremion: This targets `develop-v2`, the next major, where public API breaks are in scope. The constructor is not meant to be called directly; `default()` and `defaultDark()` are the entry points and both got the new values. Leaving as is.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/design/StreamDesign.kt
```diff
@@ -397,7 +401,7 @@ public object StreamDesign {
                 chrome = chrome,
                 accentPrimary = brand.s500,
                 accentSuccess = StreamPrimitiveColors.green400,
-                accentWarning = StreamPrimitiveColors.yellow400,
+                accentWarning = StreamPrimitiveColors.yellow200,
```
gpunto: **Non-blocking, demo app only.** Moving light-mode `accentWarning` from `yellow400` (`#F26D10`) to `yellow200` (`#F6BF57`) regresses the one place this token is used as a *foreground* rather than a fill.

`StatIndicator` (`demo-app/.../CallStats.kt:544`) paints its label **in** `indicatorColor` on a 16% alpha tint of the same colour, and the chip's parent `StatItem` has `backgroundCoreSurfaceDefault` = `chrome.s100` = `slate100` = `#EBEEF1` in light. The "Ok" chip therefore goes from **2.21:1 to 1.35:1** — effectively unreadable.

The reachable runtime path is `CallStats.kt:235` (`LatencyOrJitter`); the other two `accentWarning` uses in that file are `@Preview`. Light mode is reachable: `CallStatsDialog` (`CallScreen.kt:731`) renders under the `VideoTheme` at `CallScreen.kt:309`, which takes the default `isInDarkMode = isSystemInDarkTheme()`.

Context that softens this: all three chips already fail AA on that pattern (`Good` 2.68:1, `Bad` 3.95:1), so it's a pre-existing demo weakness this exposes rather than one the PR introduces, and it's outside the re-sync scope. The SDK-side uses are all safe — the moderation strip and the network-quality bars are fills with no text on them. If you want it gone cheaply, give the label a text-grade token instead of `indicatorColor`.

gpunto, line 404: Non-blocking, demo app only. This is the one place `accentWarning` is used as a foreground rather than a fill: `StatIndicator` (`demo-app/.../CallStats.kt:544`) paints its label in `indicatorColor` over a 16% tint of the same colour, on a `backgroundCoreSurfaceDefault` (`#EBEEF1`) parent. The "Ok" chip drops from 2.21:1 to 1.35:1.

Reachable via `CallStats.kt:235`; the file's other two `accentWarning` uses are `@Preview`. The neighbouring chips already fail AA (`Good` 2.68:1, `Bad` 3.95:1), so it's pre-existing in kind and outside this PR's scope. SDK-side uses are unaffected — the moderation strip and network bars carry no text. Cheap fix if you want it: give the label a text-grade token.

gpunto (approved):
Nothing blocking.

Nit (non-blocking): `ModerationWarningUiContainer.kt:162` still reads `// Orange column on the left` — the strip is yellow now. Not in this diff, so purely optional.

stream-video-android-ui-compose/src/main/kotlin/io/getstream/video/android/compose/theme/design/StreamDesign.kt
andremion: 83cc4966ae68f0f679209a27027e5d8f68332f96 moves the label to `textPrimary`; the tint keeps the status colour. Same commit fixes the "Orange column" comment.

